PR #77, opened 2026-01-30
Create rlm_wrapper.py

darianSmalley:
rlm_wrapper converts any standard GenerativeModel (OpenAIAsGenerativeModel or LiteLLMGenerativeModel) into a Recursive Language Model (RLM). The RLM approach alleviates issues with hitting the max context length and context rot.

### DIFF
--- a/scilink/wrappers/rlm_wrapper.py
+++ b/scilink/wrappers/rlm_wrapper.py
@@ -1,0 +1,345 @@
+import re
+import io
+import sys
+import traceback
+import logging
+import textwrap
+from types import SimpleNamespace
+from typing import List, Any, Dict, Optional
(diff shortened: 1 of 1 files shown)


ziatdinovmax (commented):
This looks very interesting! Can you add a few examples of how it should be used with current agents?

darianSmalley:
I am currently using it by adding a "recusive_mode" flag to the planning agent, as shown below:

image

Ideally, we would have all agents inherit from a base class which has this functionality (so we don't have to duplicate the logic across all the individual agents). 

I am working on adding examples, but I first need to figure out how to start scilink (and my personal scilink-tests repo) inside a docker container per the security sandbox check.

PR #118, opened 2026-03-13
Fix hardcoded R² threshold in LLM verification prompt

tkavinchakravarthy-del:
The FIT_VERIFICATION_PROMPT contained a hardcoded 0.98 value in the closing reminder line, which caused the LLM to approve fits above R²=0.98 regardless of the parameterized accept_threshold. This overrode the user-specified threshold (e.g. 0.995) and prevented retries from triggering.

Replaced hardcoded 0.98 with {accept_threshold:.3f} to match the parameterization already applied to the accept/reject criteria above.

### DIFF
--- a/scilink/agents/exp_agents/controllers/curve_fitting_controllers.py
+++ b/scilink/agents/exp_agents/controllers/curve_fitting_controllers.py
@@ -1724,7 +1724,7 @@ Return JSON:
 }}
 
 
-Remember: Rejecting a good fit (R² > 0.98) to chase marginal improvements often makes things WORSE through overfitting or convergence failures.
+Remember: Rejecting a good fit (R² > {accept_threshold:.3f}) to chase marginal improvements often makes things WORSE through overfitting or convergence failures.
 '''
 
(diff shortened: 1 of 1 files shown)


ziatdinovmax (commented):
Good catch -- this is a real bug where the hardcoded `0.98` in the closing reminder contradicts the parameterized `{accept_threshold}` used in the accept/reject criteria above. The fix is correct and minimal.

One minor nit: the existing accept threshold on line 1692 uses `:.2f`, but this fix uses `:.3f`. For thresholds like 0.995 the extra decimal place makes sense, so it might be worth updating line 1692 to `:.3f` as well for consistency across the prompt.

ziatdinovmax:
Thanks for catching this! The hardcoded `0.98` was indeed a real bug.

It's resolved in main via commit 2ae9447, where the `FIT_VERIFICATION_PROMPT` was rewritten to add a comparative-physics evaluation block. As part of that rewrite, the closing reminder now reads `R² > {accept_threshold:.2f}` — exactly the parameterization you proposed, just with `.2f` for consistency with the surrounding template formatting.

Closing as superseded by main. Appreciate the careful read.

PR #154, opened 2026-05-09
Add VASP wizard symmetric to LAMMPS + simulate-mode fixes

sarah-allec:
## Summary
  
  This is the follow-up called out in the review of the LAMMPS-only wizard PR:

  > A symmetric VASP wizard that drives `SimulationOrchestratorAgent` (#144 + #151) is a natural follow-up — out of scope here.

  While building it, real-cluster testing on a cluster surfaced several pre-existing latent bugs in the simulate-mode flow that were blocking validation. They're bundled here because each was needed to actually exercise the wizard end-to-end - without them, simulate mode doesn't work.

  ## What's in scope

  **Main feature**
  - *Adds VASP wizard symmetric to LAMMPS in the Generate tab* — engine registry split (`engines.py`); VASP wizard (`vasp_workflow.py`) drives `SimulationOrchestratorAgent.run_task` to produce POSCAR/INCAR/KPOINTS, wraps a SLURM script that assembles POTCAR from `PSEUDO_DIR` and runs the user's `vasp_command`. Reuses the engine-agnostic monitor/results/upload helpers from `sim_workflow.py`. Also lands `scilink/hpc/probe.py`, which the already-merged LAMMPS wizard imports lazily but was missing from main (latent runtime bug there too).

  **Polish**
  - *Show Model/HPC caption + Detected environment in VASP wizard* — orientation parity with LAMMPS; extends `probe_remote` with VASP binary/module detection.
  - *Drop redundant "Launch Simulations" button on welcome screen* — workaround for the now-fixed simulate dispatch bug.
  - *Add wizard form persistence across app restarts* — pre-fills HPC/SLURM/path fields from `~/.scilink/wizard_state.json`. Scientific fields (description, research goal) and secrets are deliberately not persisted.

  **Latent bug fixes surfaced during real-cluster testing**
  - *Wire simulate-mode dispatch in `start_session`* — `start_session` only branched on `app_mode == "plan"`; simulate fell through to `_init_analysis_agent`, and `agent_config["api_key"]` never got populated, so the wizards couldn't find the LLM key. `_start_simulate_session` existed but was never called.
  - *Add timeouts to `conn.run` in submit path* — defense-in-depth so a stalled SSH channel surfaces as a real error rather than locking the spinner.
  - *`HPCConnection`: SSH keepalive + real exec timeout* — the actual root-cause fix. Idle transports were silently dropped (no keepalive), and the existing `timeout=` arg on `conn.run` did nothing because `recv_exit_status()` ignores it. Set `transport.set_keepalive(30)` on every successful connect, and use `channel.status_event.wait(timeout)` so the documented timeout is actually enforced.
  - *Refresh connection-bar job counts on any status change* — monitor fragments only triggered a parent rerun on terminal transitions, so the connection bar's "N running · M pending" stayed stale through PENDING → RUNNING.

### DIFF
--- a/scilink/hpc/connection.py
+++ b/scilink/hpc/connection.py
@@ -73,6 +73,7 @@ class HPCConnection:
             )
             client.connect(**kw)
             self._client = client
+            self._enable_keepalive()
             return
     
         # Resolve all IPs and try each with a short per-IP timeout
@@ -109,6 +110,7 @@ class HPCConnection:
                 kw["timeout"] = 30
                 client.connect(**kw)
                 self._client = client
+                self._enable_keepalive()
                 return
             except Exception as exc:
                 last_err = exc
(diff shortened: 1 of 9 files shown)


ziatdinovmax (approved):
Minor comment: is the current assumption that user has POTCAR file locally?

sarah-allec:
The assumption is that the user has the POTCARs wherever they are running VASP (locally or on a cluster).

PR #161, opened 2026-05-13
 Generalize SimulationOrchestrator: scale-aware routing across periodic-DFT, MD, MLIP

sarah-allec:
## Summary
  
Generalizes the simulation orchestrator from VASP-only to scale-aware routing across all currently available simulation scales (periodic DFT, classical MD, MLIP). Nothing in `scilink/agents/sim_agents/` is hardcoded to a specific software package any more — engine-specific behavior lives in skill bundles, and the orchestrator picks (scale, engine) for the user's goal by intersecting `agent_supports × user_available × LLM_judgment`.
  
The architectural punchline: **adding a new engine is a markdown-only skill bundle drop-in.** A user who creates `scilink/skills/periodic_dft/qe/qe.md` with the canonical six sections + a small `detect:` frontmatter block — no Python, no new agent class — gets QE auto-discovered, auto-detected, and considered by the router on the next call.

  ## What the user gets
  
  ### Scale-aware base agent 
  
  - `VaspInputAgent` → `PeriodicDFTAgent` 
  - VASP-specific prose / parameter vocabulary moved out of the prompt template into the skill bundle (`skills/periodic_dft/vasp/vasp.md`)
  - Generic API works for VASP today, QE / ABINIT / CP2K tomorrow with no agent changes
  - `VaspInputAgent` retained as a backward-compat subclass
  
  ### Skill discovery
  
  - `SCILINK_SKILLS_PATH` env var for user-provided skill roots
  - `Agent.supported_software()` classmethod that auto-discovers from the bundle directory (built-in + user) — no static `SUPPORTED_SOFTWARE` tuples to maintain
  
  ### Available-software discovery
  
  ```yaml
  ---
  description: VASP DFT input generation
  detect:
    binaries:        [vasp_std, vasp_gam, vasp_ncl, vasp]
    env_vars:        [VASP_HOME, VASP_DIR]
    python_modules:  []
    guidance: |
      Free-form prose for the future LLM-driven probe fallback.
  ---
  ```

### Router + orchestrator integration
  
  - `SimulationRouter` in its own module; importable standalone
  - `route_simulation` exposed as an orchestrator tool the chat LLM calls early in the conversation
  - For VASP routings the existing dispatch flow runs end-to-end; for LAMMPS / MLIP routings the user is pointed at MDSimulationAgent / MLIPAgent directly (concrete dispatch tools deferred to follow-up — they don't gate the routing architecture)

## Cross-cutting auth fix (commits 8 + 9)
  
While testing the demo I hit a bug: most sim agents had only the proxy fallback (`SCILINK_API_KEY` inside `if base_url:`) and passed `api_key=None` straight to LiteLLM on the public path. That fails for users who store their provider key under `SCILINK_API_KEY` without also setting `ANTHROPIC_API_KEY` /  `GOOGLE_API_KEY` / etc.
  
Three files already had the right pattern (`dft_orchestrator`, `periodic_dft_agent`, `vasp_quality`). Inlined the same 5-line fallback chain in 10 more `sim_agents`, plus the `route_demo` script:
  
```python
  if api_key is None:
      provider = infer_provider(model_name) or "google"
      api_key = get_api_key(provider) or get_internal_proxy_key()
  if not api_key:
      raise APIKeyNotFoundError(infer_provider(model_name) or "google")
  
```
 
 ## Out of scope (follow-ups)
  
  - LLM-driven probe fallback for skills without structured `detect:` hints — would use the guidance field as context
  - `HPCConnection` remote probe (run probe commands over SSH on the cluster login node)
  - Interactive user-confirmation flow on top of `AvailableSoftware.auto()`
  - Concrete dispatch tools (`generate_md_inputs`, `deploy_mlip`) so non-VASP routings can complete in-orchestrator
  - Migrating in-tree callers off the `VaspInputAgent` shim and folding `VASPProjectConfig.apply_to_incar` into a per-bundle tools module, so the shim can be deleted entirely


ziatdinovmax (changes requested):
This is really cool:
> adding a new engine is a markdown-only skill bundle drop-in

One issue worth fixing before merge:
The skill rename (```vasp/vasp_input_generation/ ```→ ```periodic_dft/vasp/```) is handled by ```PeriodicDFTAgent._LEGACY_SKILL_ALIASES```, but ```VaspQualityAgent._load_skill_text``` bypasses the agent and calls the bare loader with the old domain:

```python
parsed = load_skill(skill, domain="vasp")   # skill defaults to "vasp_input_generation"
```

As a result,

```
python3 -c "from scilink.skills.loader import load_skill; load_skill('vasp_input_generation', domain='vasp')"
# FileNotFoundError: Skill 'vasp_input_generation' not found for domain 'vasp'.  
```

The try/except logs a WARNING and returns "", so the failure is silent. 

Otherwise looks good!

sarah-allec:
Good catch! Fixed in 4193dd3: same `_LEGACY_SKILL_ALIASES = {"vasp_input_generation": "vasp"}` mapping applied locally, default flipped from `"vasp_input_generation"` to `"vasp"`, and the loader now points at `domain="periodic_dft"`.

ziatdinovmax (approved):

PR #162, opened 2026-05-13
Adding LangGraph backbone

laytonwashburn-pnnl:
Skeleton for the LangGraph backbone. Implements the graph, graph state, and associaed methods.

### DIFF
--- a/scilink/graphs/__init__.py
+++ b/scilink/graphs/__init__.py
@@ -1,0 +1,13 @@
+"""
+scilink.graphs
+==============
+
+LangGraph graph definitions for SciLink orchestrators.
+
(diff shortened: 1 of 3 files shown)


ziatdinovmax:
Thanks for the skeleton! Since this is framed as an experiment, I'd like to keep it off `main` while you iterate. Going to redirect the PR base to a new long-lived branch `experimental/langgraph` on this repo so the work has a home without committing `main` to LangGraph yet.

Nothing on your end needs to change — same branch, same commits, same discussion thread, just a different base. Future PRs in this line can target `experimental/langgraph` directly. Happy to flip it toward `main` once the experiment matures.

PR #159, opened 2026-05-10
Generic MD simulation agent + LAMMPS skill (Tier C layout)

sarah-allec:
## Summary

First half of work originally on `mlip_agents`, split into a focused PR.
  
Introduces a base simulation-agent abstraction (`SimulationAgent`) and a generic MD agent (`MDSimulationAgent`) that handles classical MD across engines (LAMMPS, GROMACS, OpenMM, …) via skills + tools — no engine-specific methods on the agent itself. `LAMMPSSimulationAgent` becomes a thin backward-compat wrapper that constructs `MDSimulationAgent(skill="lammps")`.
  
Adds a comprehensive LAMMPS skill bundle (planning → implementation → validation → analysis → interpretation), placed in the upstream Tier C layout that landed in `docs/claude-md-skill-subsystem`:
  
      scilink/skills/molecular_dynamics/lammps/
          ├── __init__.py
          ├── lammps.md        # skill markdown
          └── lammps.py        # per-skill helpers (data parsing, validation)
  
Adding a new MD engine is now "drop in a skill bundle + tools module + register it on `MDSimulationAgent.TOOL_REGISTRY`", not "subclass LAMMPSSimulationAgent."
  
  ## What's in scope
  
  | File | Purpose |
  |---|---|
  | `scilink/agents/sim_agents/base_agent.py` | `SimulationAgent` abstract base |
  | `scilink/agents/sim_agents/md_simulation_agent.py` | Generic MD pipeline |
  | `scilink/agents/sim_agents/lammps_agent.py` | Backward-compat wrapper |
  | `scilink/skills/molecular_dynamics/lammps/lammps.md` | LAMMPS skill |
  | `scilink/skills/molecular_dynamics/lammps/lammps.py` | LAMMPS tools |
  | `tests/test_lammps_skill/` | 109 tests — tools + skill content |
  | `tests/test_md_agent_smoke.py` | 2 assembly / backward-compat smoke tests |
  
  ## Out of scope — follow-up PR
  
  `MLIPAgent` + machine-learning-potentials skills (general + MACE) + the
  `_shared/mlip_tools` dispatcher are held on `mlip_agents_v2` and will
  open as a separate PR after this one merges.
  
  ## Notes for review
  
  - The base abstraction is complementary to `SimulationOrchestratorAgent`, not a replacement: `SimulationAgent` is for individual computational agents; the orchestrator is the chat-driven coordinator that dispatches to them.
[...]

PR #166, opened 2026-05-14
 MLIP agent with two universal backends: MACE + CHGNet via skill bundles

sarah-allec:
## Summary
  
Adds `MLIPAgent` — a machine-learning-interatomic-potential agent — with **two universal pretrained backends** (MACE and CHGNet) wired through the same skill-bundle pattern the other scale agents use. Once merged, the simulation router auto-discovers `machine_learning_potentials` as a routable scale and both backends become candidates whenever the user has the packages installed.
  
The architectural hook: **CHGNet was added as a drop-in skill bundle + a handful of dispatcher branches** — no new agent class, no changes to the router. That's the multi-backend extensibility the skill-bundle design was built for.

### Two runners, one agent 
  
`MLIPAgent.deploy_pretrained(runner=...)`:
  - `runner="lammps"` — emits a LAMMPS input file (MACE via `pair_style mliap unified`)
  - `runner="ase"` — emits a runnable Python MD script using the backend's ASE calculator. The escape hatch when LAMMPS+MACE isn't built, and the *only* path for ASE-only backends like CHGNet.

### CHGNet is ASE-only — and the code says so honestly

CHGNet has no LAMMPS `pair_style`. Rather than paper over it:
  - `generate_lammps_input(backend="chgnet")` raises `NotImplementedError` with a message pointing at the ASE runner
  - the `chgnet.md` skill bundle's `implementation` section states the constraint up front
  - `deploy_example.py --backend chgnet --runner lammps` exits early with a clear error
  
### Backend selection
`deploy_pretrained` normally asks the LLM to pick the backend from the skill context. The new `backend=` kwarg bypasses that for benchmarks / regression tests / demos that need a specific engine — `deploy_example.py --backend {mace,chgnet}` threads it through.
  
## Cluster validation

Validated end-to-end on a cluster (A100 GPU node). The same `cu_co_slab` (14 atoms, 50 NVT steps) through both backends:
  
  | Backend | ms/step | Notes |
  |---|---|---|
  | MACE-mp-0 medium | 221.8 | `pair_style mliap` capable; more accurate forces |
  | CHGNet | 43.6 | ASE-only; ~5× faster — matches the skill's documented tradeoff |
  
  ## Testing
  
  - [x] 12 MLIP smoke tests pass (`tests/test_mlip_agent_smoke.py`): agent assembly, skill load, MACE + CHGNet ASE-script generation, LAMMPS-input generation, CHGNet skill discoverable, CHGNet LAMMPS-input raises cleanly, unknown-backend errors
  - [x] Rebased onto current `main` (includes #157, #159, #160, #161); conflict resolution was the two trivial `__init__.py` / `md_simulation_agent.py` merge markers
  - [x] `discover_scale_agents()` now lists `machine_learning_potentials → ['mace', ...]`; CHGNet joins the candidate set once `chgnet` is pip-installed
  - [x] End-to-end cluster run, both backends, real MD output
  
  ## Out of scope (follow-ups)
  
  - **Protocol-based backend refactor** — each backend currently touches ~6 dispatcher branches in `mlip_tools.py`. A
    `MLIPBackend` Protocol (one class per backend, dispatchers become one-liners) would make the 3rd/4th backend a ~30-min job. 
  - **NequIP / DeePMD** — dispatch points exist but raise `NotImplementedError` (no foundation models — they train from
    scratch)
  - **Active-learning loop backends** — `evaluate_uncertainty` / `train` / `validate_model` are MACE-only today


ziatdinovmax (changes requested):
Nice work! One regression to fix before merge: `MLIPAgent.__init__` (`mlip_agent.py:99-108`) uses the pre-#161 auth pattern and didn't pick up the cross-cutting fallback chain. Same shape as the bug #161 patched in the other 10 sim agents — users with `SCILINK_API_KEY` set but no `ANTHROPIC_API_KEY`/`GOOGLE_API_KEY` get `api_key=None` straight to LiteLLM.

Inline the same 5 lines used elsewhere, right before the `if base_url:` branch:

```python
if api_key is None:
    provider = infer_provider(model_name) or "google"
    api_key = get_api_key(provider) or get_internal_proxy_key()
if not api_key:
    raise APIKeyNotFoundError(infer_provider(model_name) or "google")
```

Plus the corresponding `APIKeyNotFoundError, get_api_key, infer_provider` imports from `...auth`.

Impact is small today (MLIPAgent isn't router-dispatched yet — only constructed by examples/tests/manual users) but grows the moment LAMMPS/MLIP dispatch tools land. Otherwise looks good.

PR #171, opened 2026-05-18
One Agent to Rule Them All

ziatdinovmax:
## Summary

Adds the **meta-agent** — an orchestrator-of-orchestrators that presents a
single conversational surface and routes each piece of research work to the
right specialist mode automatically, so the user no longer has to pick
`analyze` / `plan` / `simulate` up front.

Bare `scilink` (no subcommand) and the new `scilink meta` launch it; in the UI
it appears as the **Explore** mode. v1 covers analysis + planning; simulation
delegation is deferred behind a guarded lazy seam (the simulation package
hard-imports an optional dependency).

`scilink analyze` / `plan` / `simulate` and their orchestrators, playgrounds,
and UI modes are unchanged — the meta-agent reuses the existing orchestrators
only through their public methods.

## What the meta-agent does

- **Routes automatically.** Given a goal or uploaded files, it probes each
  file's content (array shape, table columns, document text, JSON keys) and
  delegates analysis vs planning work to the matching specialist — routing
  from evidence, not filenames.
- **Persistent specialists.** One analysis child and one planning child live
  for the whole session and are reused across delegations, so context
  accumulates — a follow-up or refinement delegation builds on the prior one
  rather than starting cold.
- **Threads context.** A delegation ledger records every delegation; the
  relevant findings of one specialist are passed as structured `context` into
  the next, with provenance.
- **Self-describing routing.** On its first turn the meta builds a capability
  inventory by reading each specialist's *live* tool registry and sub-agent
  list, so a new tool in any mode appears in the meta's routing knowledge with
  no meta-agent edit.

## Architecture

- New package `scilink/agents/meta_agent/` — `MetaOrchestratorAgent`,
  `MetaMode`, `MetaOrchestratorTools`, plus a telemetry collector.
- Specialists are consumed only through a uniform
  **`run_task(task, context) -> dict`** contract — added to the analysis and
  planning orchestrators alongside the existing `chat`. No base class;
  duck-typed.
- Children are created lazily, persist in fixed sub-directories under the
  meta-session directory, and restore from their own checkpoints, so context
  survives a meta restore.
- Each delegation runs the child under the meta's autonomy mode; in autopilot
  the specialist's human-feedback prompts (plan / code review) reach the user
  driving the session.
- Tool surface: `delegate_to_analysis`, `delegate_to_planning`,
  `inspect_uploads`, `get_delegation_history`, `summarize_session_state`.
  `delegate_to_simulation` is left as a documented, unbuilt lazy seam.

Meta session layout:

```
meta_session_YYYYMMDD_HHMMSS/
├── analysis/   results/<id>/              # persistent analysis child
├── planning/   delegations/<NN>_<slug>/   # persistent planning child
├── uploads/
├── chat_history.json
└── checkpoint.json
```

## Routing behavior

The meta gathers what a specialist genuinely needs *before* delegating:

- **Experimental metadata** — technique, instrument, conditions — is requested
  from the user up front if missing, rather than letting a specialist stall
  mid-task.
- **Equipment & setup** — for any task that produces an experimental plan,
  protocol, optimization campaign, or a recommendation of what to measure
  next, the meta asks for the user's available instruments and equipment
  before the *first* such delegation — a plan for equipment the user does not
  have is not actionable.
- A multi-file experimental series is delegated as one batched task; documents
  route by intent (analysis reference vs experiment-design input);
  charts / figures lifted from documents are treated as reference context,
  not scientofic imaging data.

## Shared extensions: skills, custom tools, MCP

One mechanism shares extensions from the meta to every specialist child:
`register_skill`, `register_tools`, and `connect_mcp_server` each record the
extension and apply it to every child — including children created lazily
afterwards. Connect an MCP server once on the meta and the analysis and
planning specialists can call its tools during a delegation.

## Telemetry

The Explore UI gains a **Telemetry** tab: a delegation dependency graph, a
per-agent tool-call sequence that updates in real time, input/output types
with tool worked/failed status, and download links to the underlying session
JSON.

## Analysis ↔ planning bridge

- Analysis runs emit a flat **feature table** (one row per spectrum / image,
  columns = conditions + extracted scalar features); a follow-up planning or
  Bayesian-optimization task ingests that file directly instead of re-typing
  numbers as prose.
- The analysis agent gains a `read_document` tool for direct PDF / DOCX
  ingestion, so a methods paper can ground an analysis plan.

## Supporting changes to existing modes

- `run_task` non-interactive entry point on the analysis and planning
  orchestrators (the delegation contract; standalone `chat` use unchanged).
- Planning per-delegation output isolation — a reused planning child writes
  each delegation's `plan.json` etc. into its own `delegations/<NN>/`
  sub-directory instead of overwriting the previous one.
- Middle autonomy level renamed `supervised` → `autopilot` (back-compatible).
- Image-series feature-table adapter; series analysis supports multiple
  control variables (primary + secondary).

## Fixes folded into this branch

- `query_knowledge_data` accepts an absolute file path, so a meta-delegated
  task whose data lives outside the knowledge directory can be queried.
- Planning no longer reports a false failure when no knowledge base is
  loaded — the LLM-knowledge fallback is used and real failures still surface.
- Planning AUTOPILOT now surfaces generated plans/code for human review, as
  its directive promises (the gate had been CO-PILOT-only).
- For a Bayesian-optimization campaign, planning runs BO directly instead of
  hand-designing a contradictory batch.
- Fixed a `TypeError` launching the hyperspectral / FFT / SAM / atomistic
  analysis agents.
- `preview_image` accepts `.npy` / `.h5` image arrays.

## Surfaces

- **CLI** — bare `scilink` and `scilink meta` launch the meta-agent;
  `scilink --help` / `-h` / `help` still print usage; `scilink meta` has its
  own argparse help. New `scilink/cli/meta.py` playground (`/children`,
  `/status`, `/mode`, …).
- **UI** — new **Explore** mode reusing the chat surface, a free-text "what do
  you want to do" upload zone with content-based file routing, a sidebar
  delegation tree with live "running" rows and status coloring, and the
  Telemetry tab. The chat input is now scoped to the Chat tab.

## Backward compatibility

- `scilink analyze` / `plan` / `simulate` behave exactly as before — dispatch
  branches, orchestrators, and playgrounds untouched.
- Bare `scilink` previously printed usage and now launches the meta-agent
  (the intended change); every `scilink <subcommand>` invocation is unchanged.
- The planning per-delegation output path is gated on a field set only inside
  `run_task`; a standalone `scilink plan` session writes to exactly today's
  locations. New checkpoint fields are read with defaults, so old checkpoints
  still load.
- No top-level import of the simulation package — installs without the
  optional simulation dependency are unaffected.

## Verification

Verified locally with ad-hoc live sessions (not committed, per project
convention):

- Importable without the optional simulation dependency; the meta exposes
  `delegate_to_analysis` / `delegate_to_planning` and no
  `delegate_to_simulation`.
- Two delegations to each specialist confirm the persistent child is reused
  and the second delegation sees the first.
- A follow-up analysis delegation answers from a prior analysis's fit results
  without re-running; a follow-up planning delegation drives
  `generate_initial_plan` → `refine_plan_with_results` to refine a plan from
  new results, with each delegation's plan isolated in its own sub-directory.
- CLI: bare `scilink`, `scilink meta --help`, `scilink --help`, and every
  `scilink <subcommand>` behave as expected.

## Next up: simulation

**Simulation will be added as the third delegation target.** The design
already accommodates it: `delegate_to_simulation` is in place as a guarded
lazy seam (the import lives inside the function body, so installs without the
optional simulation dependency stay unaffected), the capability inventory
picks up any orchestrator added to the child set automatically, and the meta
routes against live tool registries — so wiring in the simulation
orchestrator is additive and needs no change to the routing core. It is held
back here only until the simulation orchestrator stabilizes.

## Also deferred (intentionally not in this PR)

- Loading prior-analysis artifacts (`run_analysis(prior_analysis_paths=…)`)
  for the curve-fitting and hyperspectral agents — currently image-agent only.
- UI child-session navigator; possible `BasePlayground` extraction across the
  now-four mode CLIs.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -147,7 +147,7 @@ meta agent delegates through.
 ## The meta agent
 
 The meta agent sits on top of the mode orchestrators so users don't switch
-manually — bare `scilink` (or `scilink meta`) launches it. It is **not a
+manually — bare `scilink` (or `scilink explore`) launches it. It is **not a
 fourth mode**; it's an orchestrator-of-orchestrators with a different role
 (router + context bridge). It lives in `scilink/agents/meta_agent/`
 (`MetaOrchestratorAgent` + `MetaOrchestratorTools`), copying the
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -13,7 +13,7 @@ this is a settled architectural commitment, not a refactoring waypoint:
 |---|---|---|
 | `analyze` | `AnalysisOrchestratorAgent` | Experimental data analysis (microscopy, spectroscopy, …) |
 | `plan` | `PlanningOrchestratorAgent` | Experimental campaign design |
-| `simulate` | `SimulationOrchestratorAgent` *(not yet built — stub at `cli/simulate.py`)* | Computational simulations (DFT today, LAMMPS later) |
+| `simulate` | `SimulationOrchestratorAgent` | Computational simulations (DFT today, LAMMPS later) |
 
 Anything in scientific workflow falls under one of these three. There will
 **not** be a fourth mode. Future capability growth happens *inside* one of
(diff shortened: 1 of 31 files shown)


sarah-allec (approved):
Looks good! However, a short contract-test file I drafted locally (mocked LLM, 13 assertions over the three orchestrators' `run_task` shape / error path / autonomy restore / planning per-delegation isolation) caught a pre-existing `KeyError` in `simulation_orchestrator.py:89` that's silently broken - I'll send a separate PR for it.

PR #195, opened 2026-05-22
Add opt-in LLM call tracing (prompts, responses, token usage, latency)

sarah-allec:
## Summary

SciLink currently records nothing about individual LLM calls (no prompts, responses, or token usage). This adds **opt-in** tracing: when enabled, every LLM call appends one JSON record to a JSONL file — model, prompt messages, response text, finish reason, token usage, and latency. Useful for cost accounting, model comparison, and variability / reproducibility studies.

**No behavior change unless tracing is enabled.**

## Usage

```python
import scilink
scilink.enable_tracing("run/llm_trace.jsonl")   # or: export SCILINK_TRACE_FILE=run/llm_trace.jsonl
# ... run agents / workflows ...
scilink.disable_tracing()
```

Each line of the trace:

```json
{"timestamp", "model", "messages", "response_text", "finish_reason",
 "usage": {"prompt_tokens", "completion_tokens", "total_tokens"}, "latency_s"}
```

## What's in the PR

- **`scilink/tracing.py`** (new): thread-safe, global, opt-in JSONL sink — `enable_tracing` / `disable_tracing` / `is_enabled`, plus `$SCILINK_TRACE_FILE`. Inline image payloads collapse to `<image>` so traces stay small.
- **`scilink/wrappers/litellm_wrapper.py`**: a shared `_record_trace` helper hooks both `litellm.completion` sites — `LiteLLMGenerativeModel.generate_content` and `LiteLLMChatSession.send_message` — pulling text / finish_reason / usage from the raw response plus latency.
- **`scilink/__init__.py`**: exports `enable_tracing` / `disable_tracing` / `is_tracing_enabled`.
- **README**: a short "Tracing" section.

## Design

Tracing is process-global because the LLM wrapper has no per-agent/session output directory; a single sink keeps the change minimal and free of behavior change unless explicitly enabled. Streaming responses are not traced in this first cut (token usage isn't reliably available on chunks).

## Testing

- Standalone unit check of the tracer (enable → record → JSONL → disable; image sanitization; usage/latency capture).
- Live smoke test against a real Anthropic response via litellm — both hooked paths (`generate_content` and `send_message`) recorded real token counts, finish reason, and latency.
- `py_compile` clean on all changed files.

## Future work (not in this PR)

- **Auto-enable per session:** have each orchestrator (analyze / plan / simulate) enable a trace into its own session/output directory, so every run is traced without an explicit `enable_tracing()` call. Requires threading the session dir into the wrapper and touching all three orchestrators.
- **Explicit agent/role tag:** today the calling agent is implicitly identifiable from the recorded `system` message; an explicit compact `agent` field would be easier to filter on, but needs a label passed through the agents.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,33 @@ When using `SCILINK_API_KEY`, also provide a `--base-url` pointing to your OpenA
 
 ---
 
+## Tracing
+
+By default SciLink records nothing about individual LLM calls. Enable tracing to append one JSON
+record per call — model, prompt messages, response text, token usage, and latency — to a JSONL file.
+Useful for cost accounting, model comparison, and variability / reproducibility studies.
+
+```python
+import scilink
+scilink.enable_tracing("run/llm_trace.jsonl")   # or: export SCILINK_TRACE_FILE=run/llm_trace.jsonl
+# ... run agents / workflows ...
+scilink.disable_tracing()
+```
+
+Each line of the trace is a JSON object:
+
+```json
+{"timestamp": "...", "model": "...", "messages": [...], "response_text": "...",
+ "finish_reason": "stop",
+ "usage": {"prompt_tokens": 0, "completion_tokens": 0, "total_tokens": 0},
+ "latency_s": 0.0}
+```
+
+Tracing is global and opt-in — nothing is recorded unless you enable it. Inline image payloads are
+collapsed to `<image>` so traces stay small.
+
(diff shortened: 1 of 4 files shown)


ziatdinovmax (approved):
This can be included into the agent telemetry in the future (e.g., how much token did each tool call cost?)

PR #198, opened 2026-05-25
Add Quantum ESPRESSO (pw.x) skill bundle to periodic_dft

sarah-allec:
## Summary

Adds a Quantum ESPRESSO (`pw.x`) skill bundle under `periodic_dft`, so the periodic-DFT
stack can generate QE inputs (`pw.in`) — not just VASP. This is the engine-skill pattern:
dropping in `scilink/skills/periodic_dft/qe/qe.md` makes `qe` an available engine via
`PeriodicDFTAgent.supported_software()` discovery (now `['qe', 'vasp']`), with no agent
code changes.

## What's here

- `scilink/skills/periodic_dft/qe/qe.md` (+ `__init__.py`) — QE input-generation guidance
  mirroring `vasp.md`:
  - **Namelists/cards:** `&CONTROL`/`&SYSTEM`/`&ELECTRONS`/`&IONS`/`&CELL`;
    `ATOMIC_SPECIES`, `ATOMIC_POSITIONS`, `K_POINTS`, `CELL_PARAMETERS`.
  - **QE-specific conventions:** SSSP (PBE) as the recommended pseudopotential library
    (with per-element `ecutwfc`/`ecutrho`); functional carried by the pseudopotentials
    (no VASP-style functional tag); everything in **Rydberg**; smearing/`degauss` for
    metals; spin polarization; slabs vs molecules vs bulk; worked templates.
  - **`detect:` frontmatter** for `AvailableSoftware` — detects QE via `pw.x` on `$PATH`
    (which `module load quantum-espresso` provides). `env_vars` probing is intentionally
    empty: QE env vars (`$ESPRESSO_PSEUDO` etc.) point at a pseudopotential directory, not
    the `pw.x` binary, so they aren't valid detection hits.

## Why

- Engine choice shouldn't be VASP-only. Quantum ESPRESSO is free/open and widely available,
  so this lets users (e.g. workshop participants) without a VASP license run periodic DFT.

## Testing

- Smoke-tested: generated a valid `pw.in` from an In:ZnO POSCAR via the periodic-DFT path.
- `supported_software()` discovers `qe` from the bundle with no code changes.

### DIFF
--- a/scilink/skills/periodic_dft/qe/qe.md
+++ b/scilink/skills/periodic_dft/qe/qe.md
@@ -2,7 +2,7 @@
 description: Quantum ESPRESSO (pw.x) DFT input generation — namelist/card construction (&CONTROL/&SYSTEM/&ELECTRONS/&IONS/&CELL; ATOMIC_SPECIES, ATOMIC_POSITIONS, K_POINTS, CELL_PARAMETERS), plane-wave cutoffs, smearing, spin polarization, and pseudopotential conventions for metals, semiconductors, slabs, and molecules.
 detect:
   binaries: [pw.x]
-  env_vars: [ESPRESSO_PSEUDO, ESPRESSO_TMPDIR, ESPRESSO_ROOT]
+  env_vars: []
   python_modules: []
   guidance: |
     Quantum ESPRESSO's plane-wave SCF / relaxation engine is pw.x. On
@@ -10,8 +10,10 @@ detect:
     quantum-espresso/<version>` (or `qe/<version>`), so a bare $PATH
     check can miss it. Pseudopotentials are located via the `pseudo_dir`
     input variable or the $ESPRESSO_PSEUDO environment variable.
-    Detection should treat any pw.x on $PATH, or $ESPRESSO_PSEUDO /
-    $ESPRESSO_TMPDIR being set, as a positive hit.
+    Detection relies on pw.x being on $PATH (which `module load
+    quantum-espresso` provides). QE env vars like $ESPRESSO_PSEUDO point at a
+    pseudopotential directory, not the pw.x binary, so they are not used as a
+    detection hit.
 ---
 # Quantum ESPRESSO Input Generation Skill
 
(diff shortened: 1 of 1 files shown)


ziatdinovmax (changes requested):
Awesome to see SciLink's first non-VASP periodic-DFT capability, especially since QE was the first (and only) DFT code I used to run myself back in the day. And it lands with zero agent-code changes. Nice!

Two things Claude recommended to address before merging:

1) `detect:` env_vars list doesn't do what the PR description claims (qe.md frontmatter). `_probe_env_vars` (utils/available_software.py:87-104) treats an env var as a hit only when it points to a directory containing one of the binaries (it checks `Path($VAR)/<binary>`). For QE that fails: `ESPRESSO_PSEUDO` conventionally points to a directory of `.UPF` pseudopotentials (no `pw.x`), `ESPRESSO_TMPDIR` is a scratch directory, and `ESPRESSO_ROOT` — when set at all — usually points at the install root, not the `bin/` directory the probe expects. So the env-var fast-path is effectively dead for typical QE installs, and the PR description's "treats ... `$ESPRESSO_PSEUDO`/`$ESPRESSO_TMPDIR` being set as a positive hit" overstates the behavior. Fix: drop the env_vars list to `[]` and rely on PATH (which is what `module load quantum-espresso` actually gives you anyway), or — separate PR — extend `_probe_env_vars` to also accept "env var is set" as a soft hint. Either way the PR description's env-var claim should come out.

2) "Template — bulk vc-relax (metal)" is physically inconsistent (qe.md, implementation section). The example shows Cu (an FCC metal) as 2 atoms at crystal (0,0,0) and (0.5,0.5,0.5) in a 3.61 Å cubic cell — that layout is BCC, not FCC. Conventional FCC needs 4 basis atoms at (0,0,0), (½,½,0), (½,0,½), (0,½,½). Worked templates are the part of a skill an LLM leans on most heavily, so a physically wrong example teaches the model wrong physics — and once this skill is the template that ABINIT / CP2K skills copy from, the same mistake propagates. Either fix the basis to conventional 4-atom FCC, or swap Cu for a BCC metal (Fe, W, Cr) and keep the 2-atom layout.

Non-blocker, mentioning as a comment: the planning section says "~0.03 Å⁻¹ spacing" for K_POINTS but the same template uses 12³ on a 3.61 Å cell (~0.145 Å⁻¹). 12³ is the right production value for Cu; the "0.03 Å⁻¹" prose rule is overstated. Easiest fix is to soften the rule wording (e.g. "~0.15-0.25 Å⁻¹ for metals; 8-16³ on cubic 3-4 Å cells").

sarah-allec:
Thanks for the careful review! Addressed all three in 99443311:

1. **`detect.env_vars` → `[]`** — you're right that `_probe_env_vars` needs `$VAR/<binary>`, which QE's pseudo-dir/scratch/install-root vars never satisfy. Dropped to `[]` (rely on `$PATH` from `module load`), updated the frontmatter guidance, and removed the overstated env-var claim from the PR description. (A soft "env var is set" hint in `_probe_env_vars` could be a nice separate PR.)
2. **FCC Cu template** — fixed the basis to the conventional 4-atom FCC (`nat` 2→4) at a=3.61 Å; kept Cu since it's the canonical FCC example and a=3.61 Å is its correct conventional constant.
3. **K_POINTS rule** — softened to ~0.15–0.25 Å⁻¹ (8–16³ on cubic 3–4 Å cells), consistent with the 12³-on-3.61 Å template.

ziatdinovmax (approved):

PR #205, opened 2026-05-26
Uploaded the afm skill

lanceyao-research:


ziatdinovmax (changes requested):
Thanks for the contribution! The content is strong - the row-alignment, pixel-squareness, and physical-units-discipline guidance covers AFM gotchas that most analysis pipelines get wrong.

Two small fixes to address before merging - both are zero-effort and bring this bundle in line with the existing image_analysis skill convention:

1) **Add a YAML frontmatter `description:` field** at the top of `afm.md`. Compare with the sibling `stm_cafm.md`:

```yaml
---
description: STM and low-bias conductive-AFM image analysis — interprets bias-dependent features as local density of states or conductance rather than directly as atomic positions.
---
```

Without this, the orchestrator's `run_analysis` tool falls back to the first line of `## overview` for the skill-routing blurb (`scilink/agents/exp_agents/analysis_orchestrator_tools.py:124-126`). In this PR that gives the half-sentence `"Atomic force microscopy (AFM) produces images where the pixel"`, which is what the LLM sees when deciding whether to load the AFM skill - not useful for routing. Suggested wording (feel free to rephrase):

```yaml
---
description: General AFM image analysis (topography, KPFM, PFM, conductance) where intensity is a physical quantity with units - emphasizes row alignment, pixel-square resampling, and physical-unit preservation through the pipeline.
---
```

2) **Add an empty `__init__.py`** at `scilink/skills/image_analysis/afm/__init__.py`. Every sibling bundle (`atomic_stem/`, `overlapping_objects/`, `stm_cafm/`) ships one. Not strictly required for a markdown-only skill today (the loader globs `.md` files and Python 3 namespace packages cover the rest), but the convention is consistent across the codebase and adding it is a zero-cost defense against future tooling that does require package status.

Optional, not blocking: a couple of sentences in the PR description on what AFM analysis traps motivated the skill and whether it was tested in a `scilink analyze` session against a real AFM file - helps future reviewers and downstream users.

ziatdinovmax (approved):

PR #206, opened 2026-05-26
Auto-discover LLM credentials via litellm.validate_environment + SCILINK_API_KEY fallback

sarah-allec:
## Summary

Same gap as `DFTOrchestrator`'s commit `73059f74`, but for the experiment
agents (`ImageAnalysisAgent`, `CurveFittingAgent`, `HyperspectralAnalysisAgent`)
and `BOAgent`'s public-LiteLLM path. Also folds in the same simplification for
`StructureOrchestrator` (which now owns DFT's auto-discovery after the
structure-step refactor), so all of scilink's auth auto-discovery uses one
pattern.

This iteration takes your PR-review point about model-name heuristics
seriously: instead of `infer_provider`'s hardcoded `'claude'/'gpt-'/'gemini'`
substrings (which would have to be chased every time a provider renames a
model), **we delegate model→provider→env-var resolution to LiteLLM itself**
via `litellm.validate_environment(model_name)`. If LiteLLM reports the
expected env vars are missing, we fall back to `SCILINK_API_KEY`. As LiteLLM
grows support for new providers or Anthropic renames their models, scilink
automatically benefits — no change here required.

Tested in both directions:
- `ANTHROPIC_API_KEY` unset + `SCILINK_API_KEY` set → all 4 agents +
  `StructureOrchestrator` resolve to `SCILINK_API_KEY`.
- `ANTHROPIC_API_KEY` set → agents leave `api_key=None`, LiteLLM does its
  own lookup.
- Bedrock works correctly when `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY`
  are in env.

---

### The gap (recap)

`DFTOrchestrator` / `StructureOrchestrator` got model-aware auto-discovery in
commit `73059f74`: infer provider from the model name → look up the matching
env var → fall back to `SCILINK_API_KEY`. The four other agents (Image,
Curve, Hyperspectral, `BOAgent`'s public-LiteLLM path) had **no auto-discovery
at all** on their public path, so any caller relying on `SCILINK_API_KEY`
alone failed.

### The fix

```python
import litellm

# in each agent's __init__, right after normalize_params and before super().__init__:
if api_key is None and base_url is None:
    env = litellm.validate_environment(model_name)
    if not env["keys_in_environment"]:
        api_key = get_internal_proxy_key()
        if not api_key:
            raise APIKeyNotFoundError(", ".join(env["missing_keys"]) or "unknown")
```

Applied to:

- `ImageAnalysisAgent.__init__`
- `CurveFittingAgent.__init__`
- `HyperspectralAnalysisAgent.__init__`
- `BOAgent.__init__` (inside the `else: # PUBLIC LITELLM` branch, using local `api_key` instead of `self.api_key`)
- `StructureOrchestrator.__init__` (was using `infer_provider(generator_model) or 'google'`; same `validate_environment` pattern, so all of scilink's auth auto-discovery is consistent)

### Why `litellm.validate_environment`

`litellm.validate_environment(model)` returns the env vars LiteLLM itself
expects for a given model, so we always match LiteLLM's own resolution.
Examples:

```
claude-opus-4-6                        -> {"keys_in_environment": false, "missing_keys": ["ANTHROPIC_API_KEY"]}
gpt-4o                                 -> {"keys_in_environment": false, "missing_keys": ["OPENAI_API_KEY"]}
gemini-2.0-flash                       -> {"keys_in_environment": false, "missing_keys": ["VERTEXAI_PROJECT","VERTEXAI_LOCATION"]}
mistral/mistral-large                  -> {"keys_in_environment": false, "missing_keys": ["MISTRAL_API_KEY"]}
bedrock/anthropic.claude-3-5-sonnet    -> {"keys_in_environment": false, "missing_keys": ["AWS_ACCESS_KEY_ID","AWS_SECRET_ACCESS_KEY"]}
```

For **Bedrock**: when AWS creds are set, `validate_environment` reports
`keys_in_environment: true` and we leave `api_key=None` so LiteLLM does its
own AWS SigV4 signing (works correctly). When AWS creds are missing, the
error names `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` instead of an
unrelated `'anthropic'`.

### Verification

```
SCILINK_API_KEY only (no ANTHROPIC_API_KEY):
  ImageAnalysisAgent          api_key = 'test-key-XYZ'   ✓
  CurveFittingAgent           api_key = 'test-key-XYZ'   ✓
  HyperspectralAnalysisAgent  api_key = 'test-key-XYZ'   ✓
  BOAgent (LiteLLM)           model.api_key = 'test-key-XYZ'   ✓
  StructureOrchestrator       api_key = 'test-key-XYZ'   ✓

ANTHROPIC_API_KEY set:
  agents leave api_key=None  -> LiteLLM handles env-var lookup itself
```

### DIFF
--- a/scilink/agents/exp_agents/curve_fitting_agent.py
+++ b/scilink/agents/exp_agents/curve_fitting_agent.py
@@ -40,6 +40,11 @@ from .preprocess import CurvePreprocessingAgent
 from .pipelines.curve_fitting_pipelines import create_unified_curve_fitting_pipeline
 from ...skills._shared.curve_fitting_tools import load_curve_data, plot_curve_to_bytes
 from ._deprecation import normalize_params
+import litellm
+from ...auth import (
+    get_internal_proxy_key,
+    APIKeyNotFoundError,
+)
 from ...skills.loader import load_skill
 
 from .instruct import (
@@ -200,6 +205,16 @@ class CurveFittingAgent(SimpleFeedbackMixin, BaseAnalysisAgent):
         self.api_key, self.base_url = normalize_params(
             api_key, google_api_key, base_url, local_model, source="CurveFittingAgent"
         )
+
+        # Auto-discover API keys: delegate model→provider→env-var resolution to LiteLLM
+        # (validate_environment tells us which env vars LiteLLM expects for this model).
+        # If those aren't set, fall back to the internal-proxy key (SCILINK_API_KEY).
+        if self.api_key is None and self.base_url is None:
+            env = litellm.validate_environment(model_name)
+            if not env["keys_in_environment"]:
+                self.api_key = get_internal_proxy_key()
+                if not self.api_key:
+                    raise APIKeyNotFoundError(", ".join(env["missing_keys"]) or "unknown")
 
         super().__init__(
             api_key=self.api_key,
--- a/scilink/agents/exp_agents/hyperspectral_analysis_agent.py
+++ b/scilink/agents/exp_agents/hyperspectral_analysis_agent.py
@@ -26,6 +26,11 @@ from ...executors import require_sandbox_approval
 from ...skills.loader import load_skill
 
 from ._deprecation import normalize_params
+import litellm
+from ...auth import (
+    get_internal_proxy_key,
+    APIKeyNotFoundError,
+)
 
 
 class HyperspectralAnalysisAgent(SimpleFeedbackMixin, BaseAnalysisAgent):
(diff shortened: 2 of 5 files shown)


sarah-allec:
Closing this — your clarification on Teams made it clear I had the wrong mental model for `SCILINK_API_KEY`.

`SCILINK_API_KEY` is the **internal-proxy key**, intended to be paired with `base_url` (see `auth.py` — `get_internal_proxy_key()` is documented as "Get API key for internal proxy deployments", and `APIKeyManager.get_key_for_model` says "For public deployments only. Internal proxy should use `get_internal_proxy_key()`"). It's not meant to be a universal fallback for the public-LiteLLM path.

This PR (and arguably the public-path `get_internal_proxy_key()` fallback in `73059f74`) crossed that line — they worked in practice for our setup only because the value we'd been storing under `SCILINK_API_KEY` happens to be a real Anthropic key. The right fix on our side is to set `ANTHROPIC_API_KEY` (same value) instead, which is what the workshop README now says.

No upstream change needed. Sorry for the noise. Happy to also send a small PR to revert the public-path fallback in `73059f74` if you want the design enforced cleanly across the board — your call.

PR #208, opened 2026-05-27
Make tool-iteration cap configurable; surface exhaustion as run_task error

ziatdinovmax:
## Summary

Every SciLink orchestrator (analyze, plan, simulate, meta) caps its
chat-tool loop at 20 iterations. Two real consequences for programmatic
use:

1. **The cap was hardcoded** — no way to raise it for long-running
   campaigns when calling `run_task` from Python or the meta agent.
2. **Exhaustion silently looked like success.** When the cap was hit,
   the chat loop returned the warning string `⚠️ Maximum tool
   iterations reached. Please simplify your request.`, but `run_task`
   only flipped `status` to `"error"` if `chat()` *raised* — so the
   warning rode along inside `result["summary"]` while `result["status"]`
   stayed `"success"`. The meta agent's delegation ledger then recorded
   the run as successful and could chain further work off output that
   was never produced.

CLI/UI sessions don't see this because each interactive turn gets its
own fresh 20-iteration budget and the user naturally breaks a campaign
across many turns. `run_task` packs the whole goal into one turn under
AUTONOMOUS autonomy with no approval pauses, so it actually hits the
cap on non-trivial work (e.g. the produced-water tutorial: four KB
query rounds + plan generation in one shot).

This PR makes the cap configurable per-instance and per-call, and makes
exhaustion an explicit error in the programmatic surface — without
changing what CLI/UI users see.

### What changed for users

- New `max_iterations` kwarg on all four orchestrator constructors
  (default = previous 20, so existing code is untouched).
- New `max_iterations` kwarg on `run_task` (`PlanningOrchestratorAgent`,
  `AnalysisOrchestratorAgent`, `SimulationOrchestratorAgent`) — overrides
  the instance cap for a single call, then restores it.
- `run_task` now returns `status="error"` with the warning in `error`
  when the cap is hit (instead of `status="success"`).
- `chat()` returns the same warning string as before; no exception is
  raised. CLI / UI behavior is byte-identical.

### What this means for the meta agent

`MetaOrchestrator._close_delegation` already has an `error` field and
already routes the child's `status` straight through to the meta LLM.
Today an exhausted delegation gets reported as `success`; after this
change it correctly surfaces as `error`. No code change in meta beyond
its own constructor + handlers gaining the same `max_iterations`
kwarg.

### Verified live

`claude-opus-4-6` against `SimulationOrchestratorAgent`:

- Default `self.max_iterations == 20` (no API call).
- `max_iterations=1` + a tool-requiring task → `run_task` returns
  `status="error"`, warning string in `error`, `_last_chat_hit_iter_cap`
  flag set.
- Per-call `run_task(max_iterations=15)` on the same low-cap instance
  → `status="success"`; instance cap (1) restored after the call.
- `chat()` on an exhausted cap returns the warning string as a normal
  `str` (no exception), preserving CLI/UI contract.

Existing `tests/test_simulation_orchestrator.py::test_5_run_task_without_llm`
still passes. The one preexisting failure in `test_1_orchestrator_constructs`
reproduces on `main` unchanged (EXPECTED_TOOLS set drift — unrelated).

### Files changed

- `scilink/agents/planning_agents/planning_orchestrator.py`
  - Constructor: added `max_iterations: int = 20` kwarg.
  - Init body (after `_active_output_subdir`): `self.max_iterations = max_iterations`; `self._last_chat_hit_iter_cap = False`.
  - `chat()`: resets the flag at top of each turn.
  - `_handle_openai_chat` (line ~1457) and `_handle_litellm_chat` (line ~1530): `max_iterations = self.max_iterations`; on exhaustion set `self._last_chat_hit_iter_cap = True` *before* returning the warning string.
  - `run_task`: added `max_iterations: Optional[int] = None` kwarg, snapshots `original_max_iter` alongside `original_level`, applies override under the same try/finally, reads `_last_chat_hit_iter_cap` after `chat()` to flip status, restores cap in `finally`.

- `scilink/agents/exp_agents/analysis_orchestrator.py`
  - Same pattern as planning, but constructor default is `Optional[int] = None` falling back to the existing class constant `MAX_TOOL_ITERATIONS = 20` (kept for documentation).
  - Both handlers used `self.MAX_TOOL_ITERATIONS` → now `self.max_iterations`.
  - `run_task` mirrors planning's edits.

- `scilink/agents/sim_agents/simulation_orchestrator.py`
  - Same as analysis (class-constant fallback).

- `scilink/agents/meta_agent/meta_orchestrator.py`
  - Same as analysis (class-constant fallback).
  - No `run_task` method on meta — only constructor + handlers + `chat()` flag reset.

### Design choice: flag vs. exception

A `MaxIterationsReached` exception raised by the handlers (caught by
`chat()` and re-raised through `run_task`) would also work and is
arguably more idiomatic. Rejected because (a) `chat()`'s public contract
returns a string for every termination including the cap, so adding
*any* new raise from inside the loop is a CLI/UI surface change we'd
have to defend; (b) a per-instance flag is local to the orchestrator,
doesn't require defining/duplicating an exception class across four
files (CLAUDE.md "no base class" rule), and is read once at exactly
one site (`run_task` after `chat()` returns).

### Backward compatibility

- All four constructor `max_iterations` defaults preserve the previous
  cap of 20 — existing instantiations get identical behavior.
- `chat()` signature and return type are unchanged. CLI files
  (`cli/plan.py`, `cli/analyze.py`, `cli/simulate.py`, `cli/meta.py`)
  and `ui/app.py` call `agent.chat(...)` only.
- `ui/components/vasp_workflow.py:487` is the one UI surface that calls
  `run_task`; it ignores `result["status"]` and checks
  `agent.generated_structures` side-effects, so the status flip is
  invisible to it.
- Class constants `MAX_TOOL_ITERATIONS` kept in analyze/sim/meta as the
  documented default the constructor falls back to.

### Meta integration

`meta_orchestrator.py:903` already initializes `"error": None` in the
delegation ledger entry; `:911, :918` already write `result["status"]`
and `result["error"]` verbatim; `:935-936` already conditionally
surfaces `error` to the driving LLM. The status-flip lands in code
that's been waiting for it — no additional meta-side change required.

### DIFF
--- a/scilink/agents/exp_agents/analysis_orchestrator.py
+++ b/scilink/agents/exp_agents/analysis_orchestrator.py
@@ -499,6 +499,7 @@ class AnalysisOrchestratorAgent:
         analysis_mode: AnalysisMode = AnalysisMode.CO_PILOT,
         futurehouse_api_key: Optional[str] = None,
         image_analysis_depth: str = "basic",
+        max_iterations: Optional[int] = None,
         # Deprecated
         google_api_key: Optional[str] = None,
         local_model: Optional[str] = None,
@@ -608,7 +609,17 @@ class AnalysisOrchestratorAgent:
         
         self.message_count = 0
         self.last_checkpoint_message_count = 0
-        
+
+        # Per-call tool-iteration cap (handlers read self.max_iterations) and
+        # a flag they set when they hit the cap. chat() returns the warning
+        # string as before; run_task reads the flag to flip status from
+        # success → error so programmatic callers can detect exhaustion.
+        self.max_iterations = (
+            max_iterations if max_iterations is not None
+            else self.MAX_TOOL_ITERATIONS
+        )
+        self._last_chat_hit_iter_cap = False
+
         # Restore from checkpoint if requested
         if restore_checkpoint and self.checkpoint_path.exists():
             self._restore_checkpoint()
@@ -1260,7 +1271,8 @@ class AnalysisOrchestratorAgent:
     def chat(self, user_input: str) -> str:
         """Main chat interface with robust function calling support."""
         self.message_count += 1
-        
+        self._last_chat_hit_iter_cap = False
+
         # AUTO-CHECKPOINT: Every N messages
         if self.message_count - self.last_checkpoint_message_count >= self.CHECKPOINT_INTERVAL:
             print(f"  💾 Auto-checkpoint triggered (every {self.CHECKPOINT_INTERVAL} messages)...")
@@ -1291,7 +1303,8 @@ class AnalysisOrchestratorAgent:
             return f"❌ Error: {e}\n\n(Emergency checkpoint saved to {self.checkpoint_path})"
 
     def run_task(self, task: str, context: Optional[Dict[str, Any]] = None,
-                 autonomy: Optional[AnalysisMode] = None) -> Dict[str, Any]:
+                 autonomy: Optional[AnalysisMode] = None,
+                 max_iterations: Optional[int] = None) -> Dict[str, Any]:
         """Non-interactive entry point — used by the meta agent.
 
         Runs the task and returns a structured summary that's easy to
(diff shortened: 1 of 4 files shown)


sarah-allec (commented):
Two non-blocking follow-ups captured in #211 — neither blocks merge:

- regression test for the `run_task` status-flip on cap exhaustion (locks the contract this PR establishes)
- meta → child `max_iterations` plumbing (or at least a docstring note that meta's cap doesn't propagate to children's `run_task` calls)

sarah-allec (approved):

PR #222, opened 2026-05-29
Prefill API keys in the UI based on environment variables

pzeiger:
Hi!

Thanks again for the nice workshop yesterday. While interacting with and testing the SciLink UI, I was bothered by the fact that I had to copy the API key every time by hand. So I had Claude implement some changes and the API key is now dynamically prefilled based on the set environment variables. If the variable is not set, the field is left empty. Works also for the Embedding key.

Not sure if this is of interest for this repo, so feel free to decline if this is not according to your own ideas :). And please let me know if I can improve something more.

Best,
Paul

### DIFF
--- a/scilink/auth.py
+++ b/scilink/auth.py
@@ -117,28 +117,6 @@ def find_env_var_for_model(model_name: str) -> Optional[tuple]:
     return find_env_var(provider) if provider else None
 
 
-# LLM vendor providers, in the order to fall back through when no provider was
-# matched to the selected model (used for credential prefill).
-VENDOR_PROVIDERS = ('openai', 'anthropic', 'google')
-
-
-def find_first_vendor_env() -> Optional[tuple]:
-    """
-    Find the first set vendor LLM env var, regardless of model.
-
-    Lets the UI surface a single exported key even when it does not match the
-    currently-selected model's provider (the user then picks a matching model).
-
-    Returns:
-        ``(env_var_name, value)`` or None if no vendor key is set.
-    """
-    for provider in VENDOR_PROVIDERS:
-        found = find_env_var(provider)
-        if found:
-            return found
-    return None
-
-
 class APIKeyManager:
     """Simple API key management with environment variable auto-discovery."""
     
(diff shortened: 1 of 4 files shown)


sarah-allec (changes requested):
Hi Paul! This is great and very helpful. Everything works as expected, but `main` has moved since you opened this (PR #236 added an Amazon Bedrock provider system, and it touches the same lines as your prefill render). The fix is straightforward: merge `main` into your branch and resolve `sidebar.py:298-321` by keeping both blocks. Here's the merged version of that region:

```python
# Prefill credential fields from environment variables (once, before
# the widgets render).
_env_src = _seed_credentials_from_env(model) if not _locked else {}

spec = provider_for(model)
api_key = st.text_input(spec.key_label, type="password", key="cfg_api_key", disabled=_locked)
if _env_src.get("api_key"):
    st.caption(f"✓ loaded from `{_env_src['api_key']}`")
# Provider-specific inputs (e.g. AWS region for Bedrock) — rendered only
# for the matching provider; nothing extra shows for direct API providers.
for _pf in spec.fields:
    if _pf.kind == "select":
        _idx = _pf.options.index(_pf.default) if _pf.default in _pf.options else 0
        st.selectbox(_pf.label, _pf.options, index=_idx,
                     key=f"cfg_prov_{_pf.name}", help=_pf.help, disabled=_locked)
    else:
        st.text_input(_pf.label, value=_pf.default,
                      key=f"cfg_prov_{_pf.name}", help=_pf.help, disabled=_locked)
```

One small note while you're in there: `infer_provider` in auth.py doesn't recognize the Bedrock model-name pattern (`bedrock/...`), so the prefill silently no-ops for Bedrock and the field stays empty. Not a blocker — Bedrock users get the same UX they had before your PR — but if you want a follow-up, adding a `bedrock` branch to `infer_provider` would close the loop.

pzeiger:
Hi Sarah!
Thank you for your feedback and sorry for my late reply. I was away from work last week and have just returned to this. I believe that Claude and I have updated the PR according to your feedback. Please let me know if there would be anything else I should or could improve.
Best,
Paul

sarah-allec (approved):
Thanks, Paul! Everything looks good ✅ I actually wired up Bedrock on my end and tested it both ways, and the bearer-token path the prefill targets works end to end, so I'm confident the AWS handling is right. The extra docs on registering a provider are a nice addition.

One small, optional follow-up: the region dropdown still defaults to `us-east-1` rather than reading `AWS_REGION_NAME`, so a user with a non-default region set still has to pick it manually. Not a blocker though. 

Approving as-is. Thanks again!

PR #232, opened 2026-05-30
Meta agent: split combined data+metadata files before delegating

ziatdinovmax:
A new, tightly-scoped capability for the **meta agent**: split one file that holds *both* data and metadata into a clean (data, metadata) pair before delegating, so specialists get what their `analyze(data, system_info)` contract expects.

## Summary (for scientists)

Sometimes a single uploaded file carries the data *and* its metadata together — an HDF5/NeXus with attributes, a MATLAB `.mat` or `.npz` with data+meta keys, a `.dat`/`.txt`/`.csv` with a comment/header metadata block, or a TIFF carrying metadata in its tags/ImageDescription. The meta agent can now split that into a data file + a metadata JSON (`prepare_inputs`) and hand both to the specialist. Because real files mix data and metadata in unpredictable ways, the split is **generated per-file by the LLM**, not a fixed parser — but it is held to a hard safety net: the split must **reconstruct the original** (round-trip verified), so a transformed or lossy split is rejected. This is the meta's *only* code-generation surface and it is strictly limited to lossless repackaging — it never analyzes, transforms, or computes; all analysis is delegated.

This PR also closes a related gap: scientific TIFFs carry calibration/acquisition metadata in their tags/ImageDescription, but image loading (OpenCV) was silently dropping it — that metadata is now recovered and handed to the analysis agent as `system_info`.

Multiple combined files of the same kind can be split in one call: SciLink writes one split recipe, verifies it, and reuses it across the rest (re-checking each), so a 50-file batch costs one code generation and yields a uniform, consistent split.

Validated live (`claude-opus-4-6`/`4-8`): combined CSV, `.npz`, HDF5, MATLAB `.mat`, and TIFF files all split losslessly and round-trip verified, with the scientific metadata surfaced as clean top-level fields and reconstruction bookkeeping tucked away.

## Technical details (for AI reviewers)

**Shared helper `scilink/utils/file_prep.py::prepare_inputs(path, model, executor, output_dir, probe, logger, max_retries)`** — `probe → LLM generates a split+reconstruct script → static guard → sandbox exec → VERIFY`, with retries on failure (meta calls it with `max_retries=2` → 3 attempts total).

Enforcement is layered:

1. **Narrow prompt** — read input → write data (`.npy`/`.csv`) + metadata (JSON: scientific fields top-level, reconstruction bookkeeping nested under `_reconstruction`); values unchanged; explicit forbidden ops; allowed-library allowlist. Includes common-layout guidance (header-block text, container formats) and an instruction to surface metadata that lives in container attrs/keys at the top level even when it's also needed for reconstruction.
2. **Round-trip verification (`_verify` / `_roundtrip_ok`)** — the script must also reconstruct the original from its two outputs; we check `reconstruction ≈ original`. **Honest scope:** this confirms a faithful reconstruction *path* exists; it does **not**, by itself, prove the data leg (`data_path`, the file actually delegated) is untransformed, because the reconstruction is produced by the same generated script and is not forced to derive solely from the two split outputs. The data leg's fidelity rests on the prompt + import guard; the round-trip is a strong corroborating signal, not a hermetic proof. If the round-trip can't be verified, the split is **rejected**. `_verify` trusts the caller-owned metadata/recon paths (only `data_out` is model-chosen). Comparison is **content-aware per format**:
   - byte-identical fast path;
   - `.npy`/`.npz`: member-wise via `_members_equal` (numeric → `allclose`, non-numeric → `array_equal`), so string/scalar metadata keys don't crash the check;
   - **HDF5** (`.h5/.hdf5/.nxs/.nx`): recursive group/dataset + attribute comparison via `_h5_equal` (a faithful rewrite is rarely byte-identical — chunking, compression, attr/dataset order, timestamps);
   - **MATLAB `.mat`**: `_mat_equal` — `scipy.io.loadmat` content compare for v5/6/7 (ignoring `__header__`/`__version__`/`__globals__`), HDF5 fallback for v7.3;
   - **TIFF** (`.tif/.tiff`): `_tif_equal` — decoded pixels per page numeric-tolerant, TIFF tags exact (PIL);
   - text (`.csv/.tsv/.txt/.dat`): token-wise numeric-tolerant compare.
3. **Static import guard (`_static_guard`)** — AST-based, denylist matched by **dotted prefix** (`_denied_module`). Catches `import scipy.optimize`, `from scipy.optimize import …`, and `from scipy import optimize`, while allowing IO-only `scipy.io` / `scipy.sparse` (classic `.mat` needs `scipy.io`). Conservative root-level textual fallback for unparseable scripts.

**Batch path — `prepare_inputs_batch` / tool `str | list`.** The meta tool accepts a single path or a list. For a list it clusters by a structural `_signature` (from `_probe_file`: ndim/dtype, table columns, container keys — NOT exact dimensions, so same-kind/different-size files cluster), generates ONE verified split per cluster, and reuses it across siblings via a runtime `_PREP` path-dict header (the body is written generically and re-run verbatim; the static guard rejects a body that hardcodes paths instead of reading `_PREP`). Every reused file is independently round-trip verified; a sibling that can't reuse the shared body falls back to its own codegen. Returns `{status: success|partial|error, results:[{file,status,reused,...}], summary:{n_clusters,n_codegens,n_reused,n_fallback,n_failed}}`. This is what makes a multi-file upload cost one codegen per structural cluster (not per file) and produce a uniform schema — verification, not the grouping guess, is the correctness guarantee, so a wrong grouping only costs an extra codegen. `_probe_file` gained `.npz`/`.h5`/`.hdf5`/`.nxs`/`.mat` branches, a structured `.dat`/`.txt` delimited-table branch, and a `comment="#"` CSV read so combined files cluster correctly. Output collisions are prevented by a per-file subdir keyed on the absolute path hash.

**Analysis-loader TIFF metadata recovery (related fix, analyze side).** `load_image_data` reads pixels with `cv2.imread`, which drops TIFF tags. New `extract_image_metadata()` (`skills/_shared/image_analysis_tools.py`, PIL — already a dependency) parses TIFF tags, `ImageDescription` (ImageJ `key=value` / OME-XML), and resolution-tag pixel size; best-effort, never raises, `{}` for non-TIFF. `ImageAnalysisAgent.analyze()` folds it into `system_info` under `embedded_file_metadata` without overriding user-provided keys. This fixes metadata loss for *all* TIFF inputs (direct uploads), not just meta delegations.

**Thin meta tool `prepare_inputs`** (`meta_orchestrator_tools.py`): reuses `_probe_file`, lazily creates a `ScriptExecutor` behind `require_sandbox_approval`, calls the helper with `max_retries=2`, returns `data_path` + `metadata_path` (recon file is deleted after the check). Tool description + a system-prompt line make the restriction explicit (meta's only codegen; lossless prep before delegation; all analysis delegated).

**Error policy:** on a non-verifiable split the tool returns `status:error`; the meta is instructed to **surface it to the user and ask how to proceed** (analyze as-is, or supply data/metadata separately) — *not* silently delegate an unsplit file (falls back to as-is only if no user is present).

### Validation
- **Live** (`claude-opus-4-8`): combined CSV (comment-header metadata), `.npz` (array + string/scalar meta keys), HDF5 (dataset + attribute metadata), and MATLAB v5 `.mat` (array + string/scalar metadata) — all split losslessly, round-trip verified, 1 attempt each; data leg confirmed faithful against ground truth; scientific metadata top-level, bookkeeping under `_reconstruction`. Live testing surfaced (and this PR fixes) two false-rejection bugs: `.npz` string metadata keys crashing `np.allclose`, and faithful HDF5 rewrites failing byte-identity with no content fallback.
- **Unit/deterministic**: `_static_guard` (dotted-prefix allow/deny incl. `scipy.io` allowed, `from scipy import optimize` denied), `_members_equal`, `_roundtrip_ok` (npy/npz/HDF5/`.mat` accept-lossless / reject-transformed / reject-altered-metadata), `_text_equal_tolerant`, `_verify`, `_extract_script`.

### Scope / boundary
- Implements design Option A (shared helper + thin meta tool); the helper is reusable by the analysis orchestrator if wanted later.
- Heavier than a deterministic util by design — combined data+metadata files are too varied for fixed parsers; codegen+verify is the SciLink-native answer (foundation-agent extensibility loop).
- **Format coverage is bounded by `_roundtrip_ok`.** Content-aware verification exists for `.npy`/`.npz`/HDF5/`.mat`/TIFF/text; **any other extension must reconstruct byte-identically to be accepted.** There are no specialized vendor readers on the allowlist (`numpy, pandas, json, csv, struct, io, re, h5py, scipy.io, PIL, os, pathlib`), so a proprietary "vendor container" is **not** a verified capability — only a simple container the model can parse with `struct`/`numpy` and rebuild byte-for-byte would pass, and that path is untested. Adding a format = adding a content-aware branch to `_roundtrip_ok` (+ a reader to the allowlist).
- Known soft limit: the model reliably preserves all metadata losslessly, but does not always promote *every* container attribute to the top level (it may leave some only under `_reconstruction`). Losslessness is unaffected; this is a presentation preference, nudged by a prompt line rather than enforced.

### DIFF
--- a/scilink/agents/exp_agents/image_analysis_agent.py
+++ b/scilink/agents/exp_agents/image_analysis_agent.py
@@ -42,6 +42,7 @@ from ...skills._shared.image_analysis_tools import (
     load_image_data,
     image_to_thumbnail_bytes,
     create_image_montage,
+    extract_image_metadata,
 )
 from ._deprecation import normalize_params
 from ...skills.loader import load_skill
@@ -394,6 +395,17 @@ class ImageAnalysisAgent(SimpleFeedbackMixin, BaseAnalysisAgent):
         handled_system_info, series_metadata = self._extract_series_metadata(
             handled_system_info, series_metadata
         )
+
+        # Recover embedded file metadata (e.g. TIFF tags / ImageDescription) that
+        # OpenCV-based pixel loading drops, and fold it into system_info without
+        # overriding user-provided keys. Best-effort; first image is representative.
+        if input_type == "file_paths" and image_paths:
+            embedded = extract_image_metadata(image_paths[0])
+            if embedded:
+                handled_system_info.setdefault("embedded_file_metadata", embedded)
+                self.logger.info(
+                    f"   🏷️  Recovered embedded metadata from {Path(image_paths[0]).name}"
+                )
 
         # Build initial state
         state = {
--- a/scilink/agents/meta_agent/meta_orchestrator.py
+++ b/scilink/agents/meta_agent/meta_orchestrator.py
@@ -118,18 +118,23 @@ attempt to delegate simulation work.
   specialist's batch tools engage — never one delegation per file.
 - `inspect_uploads` is for routing only — do not use its output to interpret
   or analyze the data yourself; hand that to the specialist.
-- When a SINGLE uploaded file holds BOTH the data and its metadata mixed
-  together (an HDF5/NeXus with attributes, a `.npz` with data+meta keys, a
-  CSV/text with a header/comment metadata block, a vendor container), call
-  `prepare_inputs` on it to split it losslessly into a data file + a metadata
-  JSON, then delegate with those two paths. `prepare_inputs` is your ONLY
-  code-generation tool and it is STRICTLY limited to lossless file
-  repackaging (round-trip verified) — NEVER analysis, transformation, or
-  computation. You do not write or run any other code; all data processing is
-  delegated. If it errors (could not produce a verified lossless split), do NOT
-  silently delegate — tell the user you could not safely split the file and ask
-  how to proceed (e.g. analyze it as-is, or supply the data and metadata as
-  separate files). Only fall back to delegating as-is if no user is available.
+- When an uploaded file holds BOTH the data and its metadata mixed together
+  (an HDF5/NeXus with attributes, a `.npz`/`.mat` with data+meta keys, a
+  CSV/text with a header/comment metadata block, a TIFF with tags/
+  ImageDescription), call `prepare_inputs` on it to split it losslessly into a
+  data file + a metadata JSON, then delegate with those two paths. When SEVERAL
+  such combined files are the same kind (or the user says they're a batch/
+  series), pass them all as a LIST to `prepare_inputs` in ONE call — it splits
+  them with a shared, reused recipe and returns one (data, metadata) pair per
+  file — then make ONE batched delegation (never one prep or delegation per
+  file). `prepare_inputs` is your ONLY code-generation tool and it is STRICTLY
+  limited to lossless file repackaging (round-trip verified) — NEVER analysis,
+  transformation, or computation. You do not write or run any other code; all
+  data processing is delegated. If it errors, do NOT silently delegate — tell
+  the user you could not safely split the file(s) and ask how to proceed (e.g.
+  analyze as-is, or supply the data and metadata separately); for a `partial`
+  batch result, surface which files failed and ask. Only fall back to
+  delegating as-is if no user is available.
 
 **EXPERIMENTAL METADATA:**
 - Experimental data needs metadata — measurement technique, instrument,
--- a/scilink/agents/meta_agent/meta_orchestrator.py
+++ b/scilink/agents/meta_agent/meta_orchestrator.py
@@ -118,6 +118,18 @@ attempt to delegate simulation work.
   specialist's batch tools engage — never one delegation per file.
 - `inspect_uploads` is for routing only — do not use its output to interpret
   or analyze the data yourself; hand that to the specialist.
+- When a SINGLE uploaded file holds BOTH the data and its metadata mixed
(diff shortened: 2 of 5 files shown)


sarah-allec (approved):
The design is well-considered for the stated threat model (LLM accidentally analyzes/transforms while splitting) and the "honest scope" caveat on the round-trip is documented honestly. The static guard, verifier layering, and batch-cluster pattern all hold up under inspection.

**Possible follow-up:** for cheap formats (.npz, HDF5) byte-compare the chosen data members directly, not just via round-trip; or an AST-level call-target check for known transform names. Defer until you see one slip through.

**Couple smaller things:** `_signature` clusters .npz/.mat/.hdf5 by key set, so two pipelines producing same-name keys with different physical units (raw vs normalized) share a split script. Fine for repackaging, worth noting if the meta agent later assumes "same signature ⇒ same downstream analyzer." Worth an adjacency check that `require_sandbox_approval` is actually enforced in AUTONOMOUS mode - that's where this code runs LLM-generated scripts unattended.

Issue #271, opened 2026-06-01
Expose max_verification_iterations (fast/bypass mode) for ImageAnalysisAgent & HyperspectralAnalysisAgent


ziatdinovmax:
## Context

Curve fitting now supports a per-call **thoroughness knob** so analysis can run fast (real-time / in-situ) or thorough (post-experiment) on the same agent, selectable from the orchestrator/UI (commit 4f949e1):

- `CurveFittingAgent.analyze(max_verification_iterations=...)` — `0` = bypass LLM verification (accept the initial fit), `1` = single check / no refit loop, default `7` = thorough.
- `run_analysis` forwards it per-call and its tool-schema description maps natural language (e.g. "use R²=0.98 and a single verification step", "bypass verification") so the orchestrator/meta LLM sets it.

This should be mirrored to the other two analysis foundation agents so the fast/in-situ path is uniform across modalities.

## Scope

**ImageAnalysisAgent**
- Already has `max_verification_iterations` as a constructor param.
- Add it to `analyze()` using the effective-override pattern (`effective = param if param is not None else self.max_verification_iterations`).
- Add a `<= 0` bypass branch in `image_analysis_controllers.py` (verification loop is at ~`for ... in range(self.max_verification_iterations)`), mirroring the curve-fitting guard: accept the **successful** initial result with no verification; only trigger at `<= 0`; a failed/degenerate initial result still falls through to the recovery loop.

**HyperspectralAnalysisAgent**
- Add the constructor param + `analyze()` param + the `<= 0` bypass branch in its verification controller.

**Orchestrator**
- No change needed: `run_analysis` forwarding already introspects each agent's `analyze()` signature and forwards `max_verification_iterations` only to agents that accept it (others log-and-ignore). It works automatically once their `analyze()` accepts the param.

## Reference

Curve-fitting implementation:
- `curve_fitting_agent.py` — `analyze()` signature + `effective_max_verification`.
- `controllers/curve_fitting_controllers.py` — `<= 0` bypass branch (`_fit_with_quality_control`).
- `analysis_orchestrator_tools.py` — `run_analysis` param + forwarding + tool-schema NL mapping.

## Backward compatibility

The curve-fitting change is fully backward compatible (additive optional params defaulting to existing behavior; default `7` path is the verbatim original block; `0` was a degenerate/unused input before). Mirror the same care for image/hyperspectral.

PR #272, opened 2026-06-01
fit_pattern: fast, robust full-pattern XRD fitting for one-off and in-situ series

ziatdinovmax:
## What this adds (for an XRD person)

A new **full-pattern profile-fitting** capability for SciLink's curve-fitting agent (`xrd_profile` skill), built for both one-off analysis and **in-situ temperature/time series**. Instead of fitting peaks one at a time, it fits the whole diffraction pattern in a single shot, captures real peak asymmetry, and is designed to run frame-after-frame across a series without re-coding.

### Key methods inside `fit_pattern` (plain XRD language)

- **SNIP background stripping, auto-tuned** — removes the smooth background / fluorescence floor without assuming a polynomial shape, and automatically picks how hard to strip so it doesn't carve into the base of sharp peaks.
- **Automatic peak finding** — locates every significant reflection across the scan by prominence, so weak peaks aren't silently dropped.
- **Whole-pattern (global) fit** — all reflections are fit *simultaneously* on one shared background, not peak-by-peak. Overlapping reflections and the background are handled consistently, and the reported R²/residual describe the *entire* pattern (the same view the quality check uses).
- **Split (asymmetric) pseudo-Voigt line shape** — each reflection is a Gaussian–Lorentzian mix with **independent left/right widths**, capturing the real low-angle peak asymmetry from axial divergence. A symmetric profile leaves a systematic residual at the top of strong sharp peaks; on this data that asymmetry accounted for ~half the residual on the two tallest peaks. (Kα2 splitting was tested and does *not* help — the low-angle Kα1/Kα2 separation is ~0.03°, negligible.)
- **Apex-anchored intensity seeding** — each peak's starting height is read from the measured peak top, so tall sharp reflections aren't clipped during the fit.
- **Crystallite-size / microstrain ready** — returns peak position (2θ → d-spacing), FWHM (plus left/right widths and an asymmetry value), height, integrated intensity, and the Gaussian/Lorentzian mixing — feeding straight into Scherrer size and Williamson–Hall.
- **Series mode = "lock the method, not the values"** — the *same* auto-detecting call runs on every frame of an in-situ ramp, so it follows peak shifts (thermal expansion), intensity changes (phase fraction), and even new/disappearing phases (e.g. precursor → metallic Au) **without re-seeding or re-coding**. Validated across a 53-frame 31→120 °C ramp through a phase transition: R² 0.987–0.994, **0/53 frames below 0.90**.
- **Adjustable thoroughness** — a per-analysis verification budget, plus a "bypass" for fast / real-time in-situ turnaround, lets you trade between a quick good-enough fit and an exhaustive post-experiment one (same agent, same skill).
- **Built to converge fast** — internal parameter scaling spans the huge range between intensities (~10⁵ counts) and angles (~tens of degrees), so a full ~20-peak pattern fits in about a second.

---

## Technical details (for reviewers)

Three commits on this branch:
- `6b503fe` — `fit_pattern` global fitter added to `scilink/skills/curve_fitting/xrd_profile/fit_pattern.py` (auto-discovered via the `TOOL_SPEC` registry; surfaced in orchestrator mode when `xrd_profile` is active).
- `4f949e1` — per-call `max_verification_iterations` plumbed through `CurveFittingAgent.analyze()` → `run_analysis` (mirrors `r2_threshold`; introspected forwarding), with a `<=0` bypass branch in `_fit_with_quality_control` (`curve_fitting_controllers.py`). `0`=bypass, `1`=single check, default `7`=thorough. NL mapping in the `run_analysis` tool schema.
- `6a47a36` — split (asymmetric) pseudo-Voigt as the default `peak_shape`, adaptive SNIP iteration selection (favouring R²), and method-locked series guidance.

**Design / mechanism**
- Model: sum of height-parameterised pseudo-Voigts on a linear baseline, fit with `scipy.optimize.curve_fit` (TRF) using per-parameter `x_scale` + loosened tolerances (eliminates the 15–76 s/frame thrash on busy frames → ~0.8 s/frame symmetric, ~7 s/frame split-with-SNIP-sweep).
- `peak_shape='split_pseudo_voigt'` (default) fits `fwhm_left`/`fwhm_right` per peak; returns `fwhm=(fl+fr)/2` for Scherrer/W-H plus `asymmetry=(fr-fl)/(fr+fl)`. `pseudo_voigt` (symmetric) remains selectable.
- `snip_iterations='auto'` sweeps {6,10,16,24}; peaks detected once on the most-aggressive background (stable count), best background chosen by lowest residual-rms/noise within 0.002 R² of the max (favours R² so attempt-1 clears the gate).
- Returns global `r_squared`, `residual_rms_over_noise` (the verifier's statistics), per-peak params, `peak_centers`, `fit_curve`, `background_method`, `peak_shape`.

**Validation (real IMAuCl4 VT-XRD series + synthetic ground truth)**
- Split vs symmetric on 53 real frames: rms/noise mean 3.7→2.8, R² mean 0.989→0.994, better on 53/53.
- Ground-truth recovery: symmetric synthetic peaks → asymmetry ±0.01, FWHM exact; asymmetric synthetic peaks → recovered fl/fr and asymmetry within ~5%. Generalizes (degenerates to symmetric where there's no asymmetry).
- Method-lock vs value-lock across the series: auto-detect → 0/53 frames <0.90 (mean 0.994); a peak list value-locked from a *sparse* frame → 23/53 frames <0.90 (R² down to 0.47). Method-locking is robust to which frame you start from; value-locking is not. Live-verified the planner now emits `peak_centers=None`.
- Live mode behaviour: bypass (`0`) accepts attempt-1 in ~160 s; thorough (`7`) runs the full verification + judge fallback. Backward compatible (52 relevant unit tests pass; default path unchanged).

**Known limits / follow-ups**
- Split-PV slightly raises the per-regime *anchor* verification cost (more for the verifier to scrutinise); the bulk per-frame cost rises only ~3 s. A skill-scoped speed lever was scoped (capping the annealing level, not the iteration count, to avoid compressing the annealing schedule) but deliberately deferred.
- Mirror the `max_verification_iterations` knob for Image/Hyperspectral agents — tracked in #271.
- Detection over-counts on sparse patterns (30-peak cap fitting sidelobes) — pre-existing, does not affect the real peaks' FWHM.

### DIFF
--- a/scilink/agents/exp_agents/analysis_orchestrator_tools.py
+++ b/scilink/agents/exp_agents/analysis_orchestrator_tools.py
@@ -2030,6 +2030,7 @@ class AnalysisOrchestratorTools:
             reuse_locked_script: bool = False,
             literature_file: str = None,
             r2_threshold: float = None,
+            max_verification_iterations: int = None,
         ) -> str:
             """
             Execute analysis with the selected or specified agent.
@@ -2524,6 +2525,19 @@ class AnalysisOrchestratorTools:
                             f"   r2_threshold={r2_threshold} ignored: "
                             f"{self.AGENT_NAMES.get(agent_id, 'agent')} has no R² gate."
                         )
+                if max_verification_iterations is not None:
+                    # Thoroughness / turnaround override. 0 bypasses LLM
+                    # verification (fast/in-situ), higher = more refinement
+                    # (thorough/post-experiment). Only forward to agents whose
+                    # analyze() accepts it (currently CurveFitting).
+                    import inspect as _inspect
+                    if "max_verification_iterations" in _inspect.signature(agent.analyze).parameters:
+                        analyze_kwargs["max_verification_iterations"] = int(max_verification_iterations)
+                    else:
+                        self.logger.info(
+                            f"   max_verification_iterations={max_verification_iterations} ignored: "
+                            f"{self.AGENT_NAMES.get(agent_id, 'agent')} has no verification loop."
+                        )
                 result = agent.analyze(**analyze_kwargs)
                 
                 # === Store result ===
@@ -2810,6 +2824,24 @@ class AnalysisOrchestratorTools:
                         "skill's R² gate (a warning is logged on conflict), but it "
                         "will NOT replace a skill that scores by a non-R² metric. "
                         "Leave unset for standard analyses."
+                    )
+                },
+                "max_verification_iterations": {
+                    "type": "integer",
+                    "description": (
+                        "CurveFitting agent only. Controls how thorough the "
+                        "fit-verification/refinement loop is — the speed-vs-rigor "
+                        "knob. Set it ONLY when the user asks about turnaround or "
+                        "checking depth; leave unset for the standard (thorough) "
+                        "default. Map the user's intent: a fast / quick / in-situ / "
+                        "real-time / 'good enough' turnaround → 1 (a single check, "
+                        "no refinement loop); 'skip/bypass/no verification' → 0 "
+                        "(accept the first good fit with no LLM verification at "
+                        "all); an explicit count (e.g. 'two verification steps') → "
+                        "that integer; thorough / careful / publication / "
+                        "post-experiment analysis → leave unset (defaults to 7). "
+                        "Often paired with r2_threshold (e.g. 'use R²=0.98 and a "
+                        "single verification step')."
                     )
                 }
             },
--- a/scilink/agents/exp_agents/controllers/curve_fitting_controllers.py
+++ b/scilink/agents/exp_agents/controllers/curve_fitting_controllers.py
@@ -3433,7 +3433,20 @@ Return JSON with:
 
             # --- Verification loop (for anchor spectra: first overall or first in regime) ---
             fit_was_approved = False
-            if _is_anchor:
+            if (_is_anchor and self.max_verification_iterations <= 0
+                    and best_result and best_result.get("success")):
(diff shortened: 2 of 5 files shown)


ziatdinovmax:
## Real-series scenario validation ("wait a bit for the first, fast+accurate for the rest")

Ran `analyze` on 5 frames spanning the 31→120 °C ramp (incl. the transition) with temperature metadata.

**What worked as designed:**
- The planner **segmented 3 regimes** from the metadata + scout (low-T / transition / high-T) and verified a thorough anchor per regime.
- **Locked (non-anchor) frames fit in ~14 s** off the locked auto-detect script, R²≈0.998 — fast and accurate.
- **No adaptive refit triggered** ("No flagged spectra") — the method-locked auto-detect fit every frame well, so the transition needed no outlier rescue.
- Total: ~73 min for 5 frames, but that's **3 thorough anchors** (regime segmentation); on a real 53-frame series, 3 anchors + 50 locked frames amortizes far better.

**Rough edge found (follow-up):** `fit_pattern` returns `r_squared` on the **background-subtracted** pattern, but the agent's saved-fit / visualization evaluates the model against **raw absolute intensity** → a large discrepancy (saved R²≈−0.22 vs self-reported 0.996) that cost a few verifier iterations to reconcile (the agent eventually reconstructed `SNIP_background + peak_sum` to match raw). The skill template should make the background-reconstruction explicit (return/plot `background + fit_curve` for raw-intensity comparison) so the agent doesn't rediscover it each run. Not blocking, but worth a small follow-up.

PR #277, opened 2026-06-04
Engine-neutral simulation architecture (critics + scale-agnostic pipeline + skill-bundle tools)

sarah-allec:
# Engine-neutral simulation architecture (critics + scale-agnostic pipeline + skill-bundle tools)

## Summary

Refactors the simulation (`simulate`) side so the orchestrator, critics, and
tool surface are **engine-neutral**: engine-specific knowledge (VASP INCAR
conventions, syntax rules, output parsing, generation backends) lives in skill
bundles, and the shared code contains no hardcoded engine names, filenames, or
per-engine dispatch maps. Adding a new engine (QE) or scale (molecular DFT,
MD) becomes a new skill bundle and/or foundation agent — not edits to shared
code.

VASP is the worked example throughout; the next PR (LAMMPS as a second engine)
is intended as the extensibility proof — it should touch ~zero shared code.

## What's in it

**Engine-neutral critics** (`scilink/agents/sim_agents/critics.py`)
- `InputValidator` — pre-run review. Combines three grounding sources into one
  pass: the engine's deterministic syntax check, a literature-grounded review
  (when a FutureHouse key is configured), and the active skill's `validation`
  guidance + LLM judgment.
- `RunCritic` — post-run review. Reads a finished run directory and the skill's
  `interpretation` guidance; returns a verdict and, when warranted, proposed
  input patches. Handles failed and successful runs in one call.
- Both are skill-driven and contain no engine-specific code. They consolidate
  the prior VASP-only critic classes (`IncarValidatorAgent`, `VaspUpdater`,
  `VaspQualityAgent`).

**Skill-bundle tool resolution** (`scilink/skills/_shared/_registry.py`)
- New `get_tool_function(name, active_skills=[...])` resolves a `TOOL_SPEC`
  name to its callable from the active skill bundle — completing the TOOL_SPEC
  system (specs were discoverable but never resolved to callables). Engine
  helpers now live in their bundle and are reached via the registry, not
  central dispatch or by-path imports.
- VASP bundle gains: `vasp_output.py` (`snapshot_run`), `vasp_syntax.py`
  (`check_input_syntax` / `apply_input_syntax_fixes` + INCAR primitives),
  `vasp_atomate2.py` (`generate_inputs_atomate2`). VASP physics knowledge is
  authored as prose in `vasp.md`'s `validation` / `interpretation` sections.

**Scale-agnostic pipeline** (`scilink/agents/sim_agents/simulation_pipeline.py`)
- `run_complete_workflow(scale=, software=, method=)` — a deterministic
  structure → inputs → validation sequence for any scale, routing input
  generation to the scale's foundation agent and validating via the critics.
  Replaces the per-scale orchestrator classes. `DFTOrchestrator` is reduced to
  a thin deprecation shim that delegates here (`LAMMPSOrchestrator` gets the
  same treatment in the LAMMPS PR).

**Engine-neutral orchestrator tool surface**
(`simulation_orchestrator_tools.py`, `simulation_orchestrator.py`)
- Tools: `generate_dft_inputs(software=, method=)`, `validate_inputs`,
  `analyze_output`, `apply_input_adjustments`, `submit_simulation_job
  (run_command=)`, `download_job_results`, etc.
- Generic session record: `input_files: {filename: path}` instead of
  `incar_path` / `kpoints_path`; `structure_path` instead of `poscar_path`;
  `run_command` instead of a `vasp_std` default; generic report title.

## Notable decisions

- **Literature-grounded validation is preserved and generalized.** The
  FutureHouse CROW mechanism was already engine-neutral; only the query was
  VASP-shaped. It's now folded into `InputValidator` with an engine-neutral
  query, gated on a key, degrading gracefully when absent.
- **atomate2 is a generation *method*, not an engine** — it lives in the VASP
  bundle as a tool (`generate_inputs_atomate2`); `method='llm'` is the agent
  baseline, named methods are bundle tools.
- **The old VASP-specific critics are retained as a labeled benchmark
  baseline** (module-docstring banners; off the live path) for an old-vs-new
  critic comparison in the paper benchmarks — not deleted.

## Testing

- `tests/test_vasp_syntax.py` (new) — 14 deterministic tests (no LLM/API key)
  for the syntax tools + registry resolution.
- `tests/test_simulation_orchestrator.py` — targeted (no-LLM) suite updated to
  the new tool surface; all 8 pass, including the full mocked HPC flow.
- Construction + import checks across the orchestrator / pipeline / critics.

## Scope boundaries / follow-ups

- **Engine-agnostic UI** is a deliberate separate PR — this PR only updates the
  prototype UI/CLI *reads* to the new record shape so they don't break.
- **LAMMPS (second engine)** is the next PR, branched off this one.
- **Benchmark/eval harness** now lives in a dedicated repo
  (`SciLink-Benchmarks`), depending on SciLink as a pinned install — so this
  PR is library-only.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -14,7 +14,7 @@ this is a settled architectural commitment, not a refactoring waypoint:
 |---|---|---|
 | `analyze` | `AnalysisOrchestratorAgent` | Experimental data analysis (microscopy, spectroscopy, …) |
 | `plan` | `PlanningOrchestratorAgent` | Experimental campaign design |
-| `simulate` | `SimulationOrchestratorAgent` | Computational simulations (DFT today, LAMMPS later) |
+| `simulate` | `SimulationOrchestratorAgent` | Computational simulation (DFT, classical MD, MLIP-driven MD) |
 
 Anything in scientific workflow falls under one of these three. There will
 **not** be a fourth mode. Future capability growth happens *inside* one of
@@ -164,7 +164,7 @@ maintenance pain. We have not hit it.
 When building `SimulationOrchestratorAgent`, copy the structure of
 `AnalysisOrchestratorAgent`. Don't refactor the other two.
 
-## What the simulate orchestrator looks like
+## How the simulate orchestrator works
 
 Structure-centric, iterative, two-surface. **Different from analyze mode**
 in three ways: no data file required to start, structure-centric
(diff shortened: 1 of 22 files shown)


ziatdinovmax (changes requested):
Looks great! Thank you for all the work! Claude and I had a debate on whether the issue below should be addressed in this PR or the follow-up with LAMMPS and eventually Claude convinced me it should land in this PR. Can you please address it?

#### `_resolve_engine` doesn't derive domain from the engine
```python
if software:
    _engine, scale = self.orch.active_skill_and_domain()
    return software, (scale or "periodic_dft")   # domain not derived from software
```
An explicit `software=` override keeps the named engine but takes the domain
from routing (or a hardcoded `periodic_dft`). So `software="lammps"` while
unrouted yields `("lammps", "periodic_dft")` — LAMMPS dispatched to
`PeriodicDFTAgent`. Currently latent (only `vasp` is live, and it really is
`periodic_dft`). Fix with an engine→scale lookup (inverse of `_DEFAULT_ENGINE`)
so an explicit engine resolves its own scale.

**Why here and not the LAMMPS PR:** the hardcoded `periodic_dft` default is
shared-code engine bias. Deferring it forces the LAMMPS PR to edit shared
code — against its stated ~zero-shared-code goal. Landing it now keeps that
proof clean.

sarah-allec:
Addressed in e3b9b318. `_resolve_engine` now derives the scale from where the engine's skill bundle lives (scoped to the known sim scales) rather than a hardcoded default: `software="lammps"` → `molecular_dynamics`, `"vasp"` → `periodic_dft`, with no engine bias in shared code.

ziatdinovmax (approved):

Issue #256, opened 2026-06-01
Let individual agents auto-select MULTIPLE skills (not just the orchestrator passing a list)


ziatdinovmax:
## Problem

Multi-skill **loading** is fully supported end-to-end — `_load_skills_to_state(skill: str | List[str] | None)` (`base_agent.py:549`) loads a list into `skills_loaded`, and `analyze(skill=…)` / the `run_analysis` tool param both accept `str | list[str]`. But **automatic selection of multiple** skills only happens one way today: the **orchestrator LLM** choosing to pass a list to `run_analysis`. The agents' own built-in auto-selectors each pick a **single** skill:

- `ImageAnalysisAgent` inspects the image and asks the model for `{"skill": "<name>"}` — one skill only (`image_analysis_controllers.py:~1010`).
- Sim agents' `_auto_select_skill` match by engine / file-extension / force-field — one skill (`sim_agents/base_agent.py:143`, `force_field_agent.py:220`, `md_simulation_agent.py:222`).
- `CurveFittingAgent` has no internal auto-select at all — its skill always comes from the caller/orchestrator.

So a single agent invoked directly (or an image/sim run that relies on internal auto-select) can never engage more than one skill on its own, even when a dataset genuinely spans correlated techniques or needs complementary method skills.

## Proposal

Give the individual agents the ability to auto-select **multiple** relevant skills, mirroring the loading path that already supports it:

1. **Image agent auto-select → return a list.** Change the auto-select prompt/parse from `{"skill": "<name>" | null}` to `{"skills": [..]}` (possibly empty), and feed the result into the existing `skills_loaded` machinery. Keep it conservative (only clearly-relevant skills).
2. **Sim agents** — allow `_auto_select_skill` to accumulate more than one match where it makes sense (it already builds a `matched` list in `force_field_agent.py:257`).
3. **Optional shared relevance selector** — a small helper that scores available skills (by description / declared technique) against the data + metadata and returns the relevant set, reusable across agents, so selection logic isn't duplicated per agent.

## Constraints / guards

- Stay conservative — prefer **zero or one** skill unless multiple are clearly warranted; spurious multi-loading bloats the per-call prompt and tool surface.
- Respect the technique-match principle from #251 (don't pull in a technique-mismatched skill).
- The cross-modality `_MODALITY_DOMAINS` guard (#249) and per-skill tool-visibility gating already constrain the active set; multi-select must compose with them.

## Related
[...]

Issue #251, opened 2026-05-31
Skill selection should be technique-aware (and add a curve_fitting/raman skill)


ziatdinovmax:
## Problem

The orchestrator's skill selector picks a domain skill by **description match**, and when no skill matches the actual **technique** it reaches for the nearest-sounding one, pulling in technique-specific rules that don't apply.

Two observed cases:
- **Raman data → `xrd_profile` selected.** `curve_fitting` has only `epr`, `xps`, `xrd_profile` — **no Raman skill**. For a Raman/BaTiO₃ spectrum the selector chose `xrd_profile` because its description ("per-peak pseudo-Voigt fits with FWHM, position, intensity") reads as generic peak-fitting — but its content is XRD-specific (Scherrer, Williamson-Hall, 2θ windowing), which is "physically meaningless" for cm⁻¹ data. The agent's conformance check correctly flagged the mismatch and emitted "justified plan deviations" (kept SNIP/pseudo-Voigt, dropped the XRD rules), so the fit wasn't derailed — but the wrong skill is noise and it triggered a (separately-fixed) buggy validation path.
- **EELS (1-D core-loss) → `hyperspectral/eels` selected** — a *cross-modality* leak, fixed by #249's modality gate. The within-domain wrong-technique case (above) is NOT caught by that gate (`xrd_profile` is curve_fitting).

## Proposed direction

1. **Make skill selection technique-aware.** Select a skill only when it matches the data's *technique/measurement type*, not merely "has peaks"; proceed on the agent's baseline when nothing truly fits. The `run_analysis` skill-param guidance already says "auto-select if relevant" — the selector is over-selecting. Options: tighten the selection prompt to key on technique (Raman vs XRD vs XPS vs EPR), and/or have the agent confirm technique-fit (e.g. axis units / metadata) before activating a skill.
2. **Author a `curve_fitting/raman` skill.** Raman is a major curve-fitting technique with no skill today; its absence is what makes the selector reach for `xrd_profile`. A proper Raman skill closes the immediate gap.

## Notes
- Composes with #249 (cross-modality gate) — this covers the *within-domain wrong-technique* case.
- The plan-validation robustness fixes (decouple-from-skill, scout overlay, regime guard) are separate and already in a PR.

ziatdinovmax:
most of the items - except for Raman skill - are addressed via #255 #258 #261 #281. Deferring the addition of Raman skill for now.

PR #281, opened 2026-06-06
Agent-side multi-skill auto-selection (#256)

ziatdinovmax:
## Summary

Today an analysis agent could only ever engage **one** domain skill, and only when the orchestrator picked it from metadata. This PR lets the three analysis foundation agents (image, curve-fitting, hyperspectral) **choose the right skill(s) by looking at the actual data** — and use **several complementary skills together** when a dataset genuinely needs them.

In practice:

- **The agent selects skills from the data it can see** — image pixels, the fitted curve, the spectrum metadata — instead of relying on the orchestrator's metadata-only guess.
- **Multiple complementary skills compose.** An AFM image of touching grains can use the AFM skill (flatten / physical-unit discipline) *and* the overlapping-objects skill (segment the grains) in one analysis — each contributing its stage to the generated script.
- **Curve fitting stays single-and-authoritative.** A 1-D spectrum is XPS *or* EPR *or* XRD, never a blend, so the agent picks the one matching technique (or none) and never substitutes a nearest-sounding skill.
- **Uploaded custom skills are auto-selected too.** A skill added via the UI or `--skills` is now picked up by the agent automatically when it fits the data — not only when you name it explicitly.
- **The orchestrator suggests, the agent decides.** The orchestrator can pass a non-binding hint, but the agent confirms it against the data, augments it, or overrides it. A skill you *explicitly* request is still honored exactly.

Validated end-to-end against a real model (Bedrock opus-4-8) on real and synthetic data, including full analysis runs that execute the generated code.

Implements #256. Advances #251 (technique-aware selection / no-substitution), minus the new Raman skill bundle that issue also requests.

---

### Shared selector — `scilink/skills/_shared/_skill_selector.py` (new)
- `build_skill_catalog(domain, *, include_provisional=False, custom_skills=None)` — mirrors the orchestrator's catalog convention (frontmatter `description` + `[technique: …]` tag, provisional excluded). Folds in `custom_skills` ({name: path}), skipping a custom skill whose frontmatter declares a *different* `_MODALITY_DOMAINS` domain.
- `select_relevant_skills(*, model, parse_fn, domain, context_parts, max_skills=3, exclusive=False, hint=None, custom_skills=None, logger=None)` — ranked, validated, possibly-empty list; never raises. `exclusive=True` → single best technique match, capped to 1 (`_EXCLUSIVE_GUIDANCE`). `hint` injects a non-binding "Orchestrator suggestion" block (filtered to valid names).

### Per-agent wiring (all route selection → `_load_skills_to_state`, `base_agent.py:549`, already list-capable)
- **Image** — rewrote `SkillSuggestionController` (`controllers/image_analysis_controllers.py`): fixed the bypass that hand-set the singular `skill_*` fields; now `select_relevant_skills(...)` → `skills_loaded`. `exclusive=False`.
- **Curve** — new `CurveFittingSkillSuggestionController` (pipeline step 1.7, gated on a `load_skills_fn` passed from the agent so the legacy factory is unchanged); `exclusive=True`.
- **Hyperspectral** — `_auto_select_skills()` in `analyze()` (metadata-only context — see Known limitations).

### Multi-skill prose + codegen composition
- Prose stages: curve `_append_skill_context` and hyperspectral `_render_skill_block` now iterate `skills_loaded` (image already did).
- Codegen recipe: `_collect_codegen_recipe(state)` / `_render_codegen_recipe(recipes)` source the `implementation`/`analysis` section from **all** co-active skills (labeled per skill, "apply each to its stage in the plan's order"), across the main + correction codegen sites in image (`:2014/2208` pre-edit) and curve (`:2294/2435` pre-edit). **Single-skill output is byte-identical.** This fixes a real silent-drop: the top-ranked skill can carry *no* recipe (e.g. `afm` authors only overview/planning/validation) while a lower-ranked skill owns it — the old `state["skill_sections"]`-only read returned empty.
- Ordering is plan-driven and LLM-mediated (the planning stage, now multi-skill, emits the ordered pipeline; codegen follows it). Not code-enforced.

### Orchestrator (`analysis_orchestrator.py`, `analysis_orchestrator_tools.py`)
- `run_analysis` gains `skill_hint` and forwards both `skill_hint` and the orchestrator's `_custom_skills` to `analyze(...)` via signature-introspection (same pattern as `max_verification_iterations`).
- Guidance (tool-param description + system prompt) generalized: **defer skill choice to the agent by default**; `skill` (authoritative) only on explicit user request / custom; `skill_hint` (non-binding) for the orchestrator's own guess; curve techniques are mutually exclusive (pass at most one).
- Precedence: an authoritative `skill` pre-loads `skills_loaded` → the agent's selector skip-guard short-circuits (honored as-is). A `skill_hint` does *not* pre-load → the selector runs with it as a prior. Custom-skill names selected by the agent are resolved back to paths before loading (customs aren't on the loader's search roots).

### Architecture notes
`CLAUDE.md` updated: three-rung reliability ladder (prose → code-in-markdown → `TOOL_SPEC`), multi-skill composition, per-domain `exclusive` policy, authoritative-`skill`-vs-`skill_hint`, and custom-skill auto-selectability.

### Tests
- `tests/test_skill_selector.py` — 35 offline unit tests (catalog, exclusive cap, hint injection/non-binding, multi-recipe sourcing incl. top-skill-without-recipe, custom-skill folding + foreign-modality filtering).
- `tests/test_multiskill_autoselect_live.py` (new) — comprehensive 3-tier live suite (selector / agent-e2e / orchestrator), real + synthetic data, argparse-selectable. Run: `UNSAFE_EXECUTION_OK=true AWS_BEARER_TOKEN_BEDROCK=… AWS_REGION_NAME=… python tests/test_multiskill_autoselect_live.py --tier all`.
- **Live (Bedrock opus-4-8):** selector 21 pass / 1 soft, agent-e2e 4/4 (two-skill composition, auto-select, custom-skill e2e fit), orchestrator 4/4 (defer / explicit / skill_hint). 0 hard failures.

### Known limitations / follow-ups
- **Hyperspectral selector is metadata-only** (no datacube inspection), so it's the weakest signal path and partly redundant with the orchestrator — the live suite's one soft result is the EELS *datacube* returning `[]` (the EELS *spectrum*, with an explicit technique tag, selects `eels`). Upgrade to read the mean spectrum when a second hyperspectral skill lands.
- **Custom skills remain markdown-only** — `TOOL_SPEC` tools are discovered only from skills inside the installed package (`_registry` walks `_SKILLS_DIR`). Out of scope here.
- The `#251` Raman curve-fitting skill bundle is not added in this PR.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -460,6 +460,17 @@ by construction — they compose via the prose / code-in-markdown rungs, not
 `TOOL_SPEC`. So code-in-markdown is the highest reliability rung a custom
 skill can reach without a package contribution.
 
+A user-registered custom skill (UI uploader / `--skills` → `register_skill`,
+held in the orchestrator's `_custom_skills` as `{name: path}`) is still
+**auto-selectable by the agent**: `run_analysis` forwards `_custom_skills`
+to `analyze(custom_skills=…)`, and `build_skill_catalog` folds them into the
+per-domain catalog (skipping a custom skill that explicitly declares a
+*different* analysis modality), so the agent's selector treats them like
+built-ins. A selected custom name is resolved back to its path before
+loading (customs aren't on the loader's search roots). This means the
+orchestrator does NOT need to pass an uploaded skill authoritatively just to
+make it usable — it pre-loads `skill` only for an explicit user request.
+
 **Multi-skill composition.** When several skills are co-active, each may
 own a different pipeline stage (e.g. one skill's preprocessing recipe and
 another's analysis recipe), so codegen injects the `implementation`
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -429,6 +429,82 @@ Domain scientists who write markdown only can ship a skill as a single
 `<name>.md` and never touch Python; the engineer-maintained helpers
 co-locate as siblings.
 
+This yields a three-rung reliability ladder for a skill's
+`implementation` recipe — prose → code-in-markdown → `TOOL_SPEC` helper:
+
+- **Prose recipe.** Narrative guidance; the codegen LLM improvises the
+  code. Lowest fidelity, zero packaging.
+- **Code-in-markdown.** A concrete snippet in the `implementation`
+  section, injected verbatim into the codegen prompt as the recipe to
+  follow; the LLM transcribes/adapts it into the generated script (it is
+  *not* imported or run directly). **Use when** you want to pin the exact
+  algorithm / params / library (much stronger than prose), the op is
+  short-to-medium, and you're fine with the LLM adapting it to the data.
+  This is the right default for most scientist-authored skills —
+  concrete, no Python packaging, composes automatically. The agent's
+  verification loop (sandbox run + compile-check + quality gate)
+  backstops transcription errors, so a mangled snippet is corrected, not
+  silently wrong.
+- **`TOOL_SPEC` helper.** A sibling `.py` callable surfaced in the
+  codegen tool inventory and *called* by the generated script (byte-exact,
+  deterministic, testable, reusable). **Promote to this when** the code
+  must run verbatim/deterministically, it's long or numerically
+  sensitive, or it's a reusable stage you want tested and called
+  identically every time (and you can contribute it to the package).
+
+Note the packaging boundary: `TOOL_SPEC` tools are discovered only from
+skills *inside the installed package* (`_registry` walks `_SKILLS_DIR`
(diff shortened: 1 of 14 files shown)


sarah-allec (approved):
Looks good! One optional nit: `_collect_codegen_recipe/_render_codegen_recipe` are duplicated across the image and curve controllers - might be worth lifting into `_shared` before a third copy shows up.

PR #283, opened 2026-06-07
Staged + parallel MD campaigns (sequential multi-phase + parameter sweeps)

sarah-allec:
Extends the engine-neutral refinement loop (#280) to run **multi-stage MD campaigns**, in both shapes:

- **Sequential multi-phase** — an optimization → equilibration → production chain runs as ordered stages sharing a run directory, so restart files chain and the per-phase refinement loop fires on each phase.
- **Parallel fan-out** — a parameter sweep (e.g. temperature) expands one parameterized script into N independent member runs in isolated directories, each refined independently; a failing member doesn't abort its siblings.

Both flow through one new `run_campaign` over a `Stage` model (sequential step | parallel fan-out | combine); `run_refinement` becomes a thin wrapper (a single sequential stage), so existing single-phase behavior is unchanged. What members differ by lives in their generated `input_files` — the loop stays engine-neutral (no engine names or filenames in shared code). The sweep placeholder and expander live in the LAMMPS skill bundle, so adding an engine is a bundle, not loop changes.

Generation: `MDSimulationAgent` auto-detects a sweep from the plan and emits a fan-out campaign; `generate_staged_simulation` emits a normalized sequential campaign (reachable via `run_complete_workflow(..., staged=True)`). The combine stage (run-once post-processing, e.g. WHAM) is already supported by the data model for a future umbrella-sampling skill.

Validation:
- 35 loop / stage / generation unit tests + the pipeline-integration suite green.
- Parallel sweep verified live end-to-end against real `lmp` (generation 8/8, execution 4/4), including on a PNNL HPC compute node via `LocalExecutor`.
- Sequential multi-phase verified live against real `lmp` (5/5): a 2-stage run where the production stage has only `read_restart` succeeds, proving restart genuinely chains through the shared run dir — and the per-phase refinement loop fired on the equilibration stage before it passed.
- The `LAMMPSOrchestrator` benchmark baseline is preserved.

### DIFF
--- a/scilink/agents/sim_agents/lammps_orchestrator.py
+++ b/scilink/agents/sim_agents/lammps_orchestrator.py
@@ -183,9 +183,9 @@ class LAMMPSOrchestrator:
             self.logger.error(f"Failed to generate simulation: {e}")
             return self._failed_result(f"Simulation generation failed: {e}")
         
-        stages = sim_info.get("stages", [])
         stage_scripts = sim_info.get("staged_scripts", {})
-        
+        stages = list(stage_scripts.keys())
+
         if not stages or not stage_scripts:
             # Fallback to single script
             self.logger.warning("No staged scripts generated, using single script")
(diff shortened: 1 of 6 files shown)


ziatdinovmax (approved):
Approved 👍 . A quick thought: now that planning mode works for both simulation and experimental planing, it would be good to start thinking about wiring sim agents into the meta agent to enable `planning→simulate->planning` chains that will also take advantage of the foundation optimization agent in the future.

PR #284, opened 2026-06-07
ClusterExecutor: run refinement-loop phases on an HPC scheduler

sarah-allec:
Adds `ClusterExecutor` — the HPC counterpart to `LocalExecutor` — so the engine-neutral refinement loop can run on a scheduler instead of a local subprocess, without the loop (or anything downstream) knowing the difference.

A `run()` does: upload the phase's input files → submit a batch job → poll to a terminal state → download the outputs into the local `run_dir` → persist the same engine-neutral `run_stdout.log` / `run_stderr.log` / `run_returncode.txt` files `LocalExecutor` writes. Because the outputs land in the local `run_dir`, the post-run `RunCritic` reads a cluster run's snapshot byte-for-byte the way it reads a local one — the executor swaps out, nothing else changes.

Design:
- **Same `Executor` contract** — `run_complete_workflow(..., executor=ClusterExecutor.connect(...))` works through the existing seam; no pipeline changes were needed.
- **Scheduler-neutral batch building** — a new `Scheduler.batch_directives(resources)` method translates an engine-neutral resource dict into `#SBATCH` / `#PBS` / `#BSUB` lines, and `scilink/hpc/batch.build_batch_script` wraps the command to capture its outputs. Adding a scheduler is one `Scheduler` subclass, nothing else.
- **Placement / layering** — `ClusterExecutor` lives with the loop in `sim_agents` (it implements the loop's contract), and imports the HPC backend lazily, so `scilink/hpc` stays self-contained and importing the loop doesn't pull in paramiko.
- **`ClusterExecutor.connect(...)`** — one-call factory that assembles the connection (inferring password vs key auth) and returns a ready executor.

Scope (v1): one job per `run()` call, blocking. The loop calls `run()` serially per phase and per fan-out member, so concurrent submission across members is a clean future seam (`run_many`), not a contract change.

Validation:
- 14 offline unit tests — a fake connection + scheduler exercise the full upload → submit → poll → download → persist flow, plus submit-failure, wall-clock-timeout (cancel), returncode fallback, the `connect` factory, and per-scheduler directives. No SSH, no paramiko, no cluster.
- Live on a real SLURM cluster (PNNL Deception): password-auth SSH → `sbatch` → poll the real queue to completion → outputs downloaded locally with `returncode 0`.

### DIFF
--- a/scilink/agents/sim_agents/cluster_executor.py
+++ b/scilink/agents/sim_agents/cluster_executor.py
@@ -1,0 +1,206 @@
+"""Run refinement-loop phases on an HPC scheduler.
+
+``ClusterExecutor`` is the cluster counterpart to ``LocalExecutor``: it
+satisfies the same :class:`~scilink.agents.sim_agents.refinement.Executor`
+contract, so the refinement loop drives it without knowing a run happens on a
+remote scheduler rather than a local subprocess. A ``run`` uploads the phase's
+input files, submits a batch job, polls until the job reaches a terminal state,
+downloads the outputs back into the local ``run_dir``, and persists the same
+engine-neutral ``stdout`` / ``stderr`` / ``returncode`` files the local executor
+writes — so the post-run critic's snapshot is identical regardless of executor.
+
+This lives with the loop (not in ``scilink.hpc``) because it implements the
+loop's ``Executor`` contract; ``scilink.hpc`` stays self-contained and is
+imported lazily, so importing the loop does not pull in paramiko.
+
+v1 runs one job per ``run`` call and blocks until it finishes. The refinement
+loop calls ``run`` serially per phase (and per fan-out member), so a campaign's
+parallel members are submitted one after another. Concurrent submission across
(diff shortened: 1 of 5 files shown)


ziatdinovmax (commented):
Verified locally: 14+5 tests green, API contracts match, all schedulers implement the new abstract method.

**Blocking**
- **PBS writes outputs to the wrong dir.** No `cd` in the wrapper; PBS/Torque starts jobs in `$HOME`, not the submit dir (needs `cd $PBS_O_WORKDIR`). Only SLURM was live-tested. Fix: a scheduler `workdir_prelude()` in `build_batch_script`, or scope the PR to SLURM.

**Minor**
- "Offline" tests still hard-require paramiko (optional dep) via `hpc/__init__.py` → use `importorskip`.
- `_ensure_connected` can't recover a password-auth session (arg-less `connect()`); harmless in happy path.

ziatdinovmax (approved):

PR #286, opened 2026-06-09
Add MolecularQCAgent: molecular/finite-system QC scale (NWChem first)

sarah-allec:
## What

Adds a third simulate-mode scale, **`molecular_qc`**, for finite-system electronic structure (isolated molecules, ions, small clusters) alongside `periodic_dft` and `molecular_dynamics`. The discriminator from `periodic_dft` is *finite vs. periodic* — **not** the method or basis. `MolecularQCAgent` is therefore software-, method-, and basis-agnostic (DFT / HF / MP2 / CCSD(T) / CASSCF are a deck-keyword choice inside one engine), mirroring `PeriodicDFTAgent` exactly; all engine specifics live in per-engine skill bundles under `skills/molecular_qc/<engine>/`.

## NWChem first

The `nwchem` bundle is the first wired engine:
- **`nwchem.md`** — `overview / planning / implementation / interpretation / validation` guidance for deck generation (DFT/HF/post-HF, geometry optimization, analytic frequencies + thermochemistry, COSMO/SMD implicit solvation), plus a `detect:` block matching the existing vasp/qe/lammps convention.
- **`nwchem_output.py`** — a `snapshot_run(output_dir)` parser built on **cclib**, registered via `TOOL_SPEC` (`agents=["simulation"]`). cclib reads NWChem / ORCA / Gaussian / PySCF uniformly, so this one snapshot contract serves any `molecular_qc` engine added later — the molecular analog of the pymatgen-based periodic critic. cclib is imported lazily so the module (and its `TOOL_SPEC`) import cleanly without it.

The agent's pre-submit syntax pass and output snapshot are dispatched through the skill registry by conventional tool name (`check_input_syntax` / `apply_input_syntax_fixes` / `snapshot_run`) — the agent never names an engine or imports a bundle by path, so a new engine is a drop-in sibling bundle.

## Wiring

- `simulation_router.py` — engine discovery (`discover_scale_agents`) + scale description; renames the placeholder `molecular_dft` to `molecular_qc` to reflect the method-agnostic scope.
- `simulation_pipeline.py` — `_generate_inputs` branch + default engine (`molecular_qc → nwchem`).

## Tests

`19 passed, 1 skipped`. Coverage: output parsing (layered so the spec/import and empty-dir cases run without cclib; the real-`.out` parse auto-skips until a fixture is captured — DFT was never run in the source archive), the generation pipeline, and router discovery. `tests/fixtures/nwchem/README.md` documents how to capture a real fixture to activate the skipped layer.


ziatdinovmax (commented):
No blocking issues.

One step before merge: please update the branch from main (merging main in / the "Update branch" button is fine — no need to rewrite history) so CI runs against the current tree; the refinement path has evolved since June. Heads-up on the one thing that will break: `RunCritic.assess` is now called with an `input_files=` kwarg, so `FakeRunCritic.assess` in `test_molecular_qc_pipeline.py` needs to accept it (I ran your tests against a trial merge with current main — that stub signature is the only failure, everything else stays green). Once CI passes I'll approve and merge.

Non-blocking follow-up worth filing: a deterministic geometry consistency check after deck generation (atom count + element list vs. the structure file) — the deck's inline `geometry` block is currently LLM-transcribed, and a silent transcription error is the one failure mode nothing downstream would catch.

sarah-allec:
Thanks! Merged main in and updated the `FakeRunCritic.assess` stub to take `input_files` (and `check_observables` while I was there) — the rest of the suite stayed green against the merged tree, so CI should be clean now. Re-requesting review.

Good call on the geometry consistency check — that inline `geometry` block is LLM-transcribed, so a silent transcription error really is the one thing nothing downstream would catch. Filed as a follow-up (#400): a deterministic post-deck check of atom count + element list against the structure file.

ziatdinovmax (approved):

PR #293, opened 2026-06-13
Meta agent: parallel multi-dataset fan-out (gate + full-mesh + fusion)

ziatdinovmax:
## Summary (for scientists)

This adds **parallel multi-dataset analysis** to the meta agent. When you give the assistant two or more complementary measurements of the *same* sample — e.g. a STEM image **and** an XPS spectrum, or DSC **and** TGA — it can now analyze them **at the same time, each informed by the others**, and then **fuse** the results into one cross-dataset interpretation.

How it behaves:

- **It only offers this when the datasets are genuinely complementary.** A complementarity gate checks that the inputs measure different-but-related aspects of one system (e.g. spatially-resolved morphology vs. bulk chemical state). If they're not complementary, it falls back to normal one-at-a-time analysis.
- **It asks before launching.** You get a clean confirmation panel (verdict, the shared "join axis", the branch list, a one-line rationale) and a Cancel / Launch choice.
- **Each analysis sees the others as auxiliary context** (full-mesh), so the XPS fit knows what the image found and vice-versa.
- **The branches run truly in parallel** (threads), with a "still running" heartbeat so a slow branch doesn't look like a hang.
- **The fused report** reconciles the datasets (e.g. "STEM oxide area = 40% and XPS Ti⁴⁺ = 38% agree"), is **honest about disagreements** (it will say a quantitative cross-check *cannot* be made if one fit didn't converge), carries a structured **caveats** section, embeds each branch's figure, and emits an **HTML report**. Literature **novelty is assessed once on the fused claims**, not per branch.

Also included: a `starting_annealing_level` control so a **re-run** of a curve/image fit can start from a higher "temperature" (skip the constraint stages a prior run already showed were inadequate), plus a batch of fan-out UI fixes surfaced during live testing.

Validated end-to-end on Bedrock opus-4-8 across image+spectrum, image+hyperspectral, DSC+TGA, and 3-way scenarios.

---

<details>
<summary><b>Technical details (for reviewers)</b></summary>

### Architecture
The fan-out is a **meta-agent primitive**, not a fourth mode. It lives in `scilink/agents/meta_agent/fanout.py` and is exposed through three tools on `MetaOrchestratorTools` (`assess_complementarity`, `run_fanout`, `fuse_delegations`). It consumes the mode orchestrators only through their duck-typed `run_task(task, context, autonomy)` contract — no base class.

- `assess_complementarity` / `run_fanout` — complementarity gate via a **tool-free** `_structured_model` (avoids empty completions on Bedrock when the chat model carries delegation tool schemas) + `_llm_json` with retry and multimodal `extra_parts`. Gate rubric credits *spatially-resolved-vs-bulk-of-same-sample* complementarity (criterion #3) — fixed an image+spectrum flip-flop.
- `run_fanout` — `ThreadPoolExecutor` over branches; each branch runs an ephemeral child analysis orchestrator at `AUTONOMOUS`. `_mesh_task` composes each branch's self-contained task with **full-mesh companions** passed via `run_analysis`'s `auxiliary_data`/`auxiliary_label`, forwards the primary's metadata (so branches don't synthesize it), and injects a per-branch NOTE suppressing per-branch novelty.
- `fuse_delegations` — drives `HOLISTIC_EXPERIMENTAL_SYNTHESIS_INSTRUCTIONS` (generalized for spatial / shared-axis / length-scale / complementary-observables joins, with an anti-spurious principle + `caveats` output field), feeds per-branch key figures to the multimodal fusion, writes `fusion_report.{json,html}`, and runs `_assess_fusion_novelty` once on the fused claims (inline `novelty_score` / `novelty_explanation` per claim, keyed on `has_anyone_question`; clean no-op without a FutureHouse backend).

### Heartbeat
Plain-text only (no `\r`/ANSI — branch children write to the same stdout from worker threads and output is often captured/ANSI-stripped). Polls every 5 s for prompt completion announcements; prints a "still running" tick at most every 60 s; a completion resets the tick timer. Elapsed via `time.monotonic()`.

### `starting_annealing_level` (curve + image)
`analyze(starting_annealing_level=None)` on `CurveFittingAgent` / `ImageAnalysisAgent`; controllers compute `_start_level = clamp(state["_starting_annealing_level"] or 0)` and seed the initial level, loop level, and `_previous_annealing_level = max(start-1, 0)` (so starting AT hot still fires fresh generation). `run_analysis` exposes it, forwarded via signature-introspection (like `max_verification_iterations`). **Default/None is byte-identical** to prior behavior `(0,0,0)`; verified.

### Shared-stack fixes (each a strict no-op on the success path)
`8a92dd3`: Bug A salvage `AttributeError`, Bug B `NoneType` subscript in `analysis_orchestrator_tools.py`, Bug C 8000px image cap, + sidecar metadata backstop.

### UI
`scilink/ui/app.py` `_render_fanout_confirm` (markdown panel; rationale inline, no expander); equal-size Cancel/Launch via a scoped `fanout-actions-anchor` + dark-CSS height pin in `theme.py`; `use_container_width`→`width="stretch"` across components; theme-toggle "Cached ForwardMsg MISS" fixed by raising `minCachedMessageSize` to 50 KB in `.streamlit/config.toml` (theme CSS is 17–23 KB, was being reference-cached and evicted); white background on the fusion HTML so it renders in dark hosts.

### Tests
`tests/test_meta_fanout_live.py`, `test_meta_fanout_robustness.py` (24/24 offline), `test_meta_fanout_hyperspectral_live.py`, `test_meta_fanout_3way_live.py`, `test_meta_fanout_stress_live.py`. `FANOUT_TEST_FINDINGS.md` records the validation log.

### Notes
- Targets the current `origin/main` (which already has #290/#291).
- Demo data (`fanout_demo_data/`) and figures are intentionally left untracked.

</details>

### DIFF
--- a/FANOUT_TEST_FINDINGS.md
+++ b/FANOUT_TEST_FINDINGS.md
@@ -29,6 +29,21 @@ bugs found, and what was fixed vs. flagged.
   agents' own QC correctly rejected degenerate fits on the synthetic cube. A
   clean re-run with proper energy metadata confirms the EELS branch. ✅
 
+### 3-WAY end-to-end (image + datacube + spectrum, one sample) ✅
+HAADF→`ImageAnalysisAgent`, EELS cube→`HyperspectralAnalysisAgent`,
+XPS→`CurveFittingAgent` — **three different agents concurrently**, all 3
+produced output. Gate complementary 0.90 over a set with TWO join axes
+(spatial HAADF↔EELS + chemical-state ↔XPS). Fusion found the real correlation
+(bright HAADF domains = Ti⁴⁺ TiO₂ domains, ~45/55% spatial coincidence) **and
+explicitly flagged the unresolved dark phase** — honest demarcation on a
+3-dataset synthesis.
+
+### DSC + TGA end-to-end (same-modality, two CurveFitting branches) ✅
+Fusion reconciled events point-for-point on the shared temperature axis:
+dehydration (DSC endotherm + coincident TGA mass loss) vs a mass-neutral
+crystallization exotherm (DSC-only) — the planted physics — and flagged a
+DSC-vs-TGA quantitative discrepancy, prioritizing the direct mass measurement.
+
 ### Stress battery — data-type breadth (gate, one LLM call each) — 6/6 graded
 | Scenario | Verdict | Note |
 |---|---|---|
--- a/FANOUT_TEST_FINDINGS.md
+++ b/FANOUT_TEST_FINDINGS.md
@@ -29,6 +29,19 @@ bugs found, and what was fixed vs. flagged.
   agents' own QC correctly rejected degenerate fits on the synthetic cube. A
   clean re-run with proper energy metadata confirms the EELS branch. ✅
 
+### Stress battery — data-type breadth (gate, one LLM call each) — 6/6 graded
+| Scenario | Verdict | Note |
+|---|---|---|
+| **DSC + TGA** (thermal, shared T axis) | complementary 0.92 | join = "temperature at 10°C/min" ✅ |
+| **Raman + FTIR** (same calcite) | complementary 0.86 | join = vibrational wavenumber ✅ |
+| **PL + UV-Vis** (same QDs) | complementary 0.90 | join = wavelength / one band gap ✅ |
+| **image + results-table** (mis-routed) | unrelated 0.85 | correctly rejected planning data ✅ |
+| **Raman(polymer) + DSC(metal)** | unrelated 0.95 | different systems ✅ |
+| **two FTIR, same sample** | redundant 0.95 | replicate, clustered ✅ |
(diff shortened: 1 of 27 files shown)


sarah-allec (changes requested):
Overall looks great and very helpful! My only concern is about the honesty guarantee: The protection against inventing a fake "these datasets agree!" conclusion is the complementarity gate — it refuses to pair datasets that aren't genuinely related. But `fuse_delegations` can be called directly on any two finished analyses, and it doesn't re-run that gate. So someone can analyze two unrelated things and fuse them, and it'll produce exactly the manufactured correlation the gate exists to prevent. A couple options: either route direct fusions through the same gate, or clearly mark in the report when a fusion wasn't gated.

sarah-allec (approved):

PR #297, opened 2026-06-16
Robust multi-column & instrument-export handling for curve fitting

ziatdinovmax:
## Summary

Curve fitting often starts from real instrument exports — files that mix a metadata header block with a multi-column data table (temperature, time, signal, mass, …). This PR makes that path reliable in three everyday situations that previously went wrong silently:

1. **When you correct the agent's column choice, the correction now sticks.** Previously, if the agent picked the wrong column, you pointed it out in the feedback step, it showed you a corrected plan, you accepted — and then it fit the *original* (wrong) column anyway. The accepted correction now actually drives the fit.

2. **Column names are no longer lost for common instrument files.** Many exports put the column-name row behind a comment marker (`#`, `##`, `%`). The old reader threw that row away, so the agent saw nameless columns and could pick an instrument axis (like a time channel) as the signal. Names are now recovered, so the agent knows which column is which.

3. **Files the loader couldn't read no longer fail outright.** A metadata block above a plain column header used to make loading fail entirely. Such files are now automatically separated into a clean data file + metadata and loaded successfully.

The guiding idea: let the model handle the *interpretation* of messy files (which column is what, where the data starts, what the names are), while deterministic code keeps doing the things it does reliably — extracting the numbers and verifying the result. This replaces brittle, format-specific parsing rules with a more general, model-driven path.

**Validation:** a cross-technique live stress matrix covering thermal, photoemission, diffraction, vibrational, magnetic-resonance, and optical spectroscopies (21 cases, synthetic data) — every case loaded correctly and locked the intended signal column.

---

## Technical details

Five commits, each scoped to its broken path with no-regression discipline.

### 1. Refinement honors the human-feedback column correction
`controllers/curve_fitting_controllers.py` — `HumanFeedbackRefinementController._refine_plan`
- The refinement path updated the narrative plan (`physical_model`/`fitting_strategy`) but never re-extracted `column_mapping`, so `_resolve_column_mapping` locked the *original* selection. Now re-extracts `column_mapping`/`column_mapping_note` (guarded by `column_info`, falls back to prior when omitted), surfaces the current mapping in the plan summary, and injects the column-structure block into the refine prompt so the model can actually change the X/Y choice.

### 2. Recover comment-hidden column headers
`skills/_shared/curve_fitting_tools.py` — `sniff_column_names` + `_recover_commented_header`
- `pd.read_csv(comment="#")` discards a header row that itself begins with a comment marker. Added a fallback — reached **only** when the primary read finds no usable header — that locates the header row adjacent to the numeric data block and strips leading comment punctuation (`#`/`##`/`%`/`;`/`//`). Format-agnostic (any delimiter / column count). A first-cell comment-marker guard defers `%`/`;`/`//` headers (which `comment="#"` misreads) to the same recovery. Primary path unchanged → currently-working files return byte-identical results.

### 3. Show the raw file head to the split codegen
`utils/file_prep.py` — `_build_prompt` + `_raw_text_head`
- The lossless split codegen only received a structural probe (`{n_columns, dtypes}`) — never the metadata block or column-name row — so it wrote splits blind and could not preserve names. `_build_prompt` now appends the first ~60 raw lines of text-like files (capped by lines AND chars — head, not bulk values). Binary containers (`.npy`/`.npz`/`.h5`/`.mat`/`.tif`) keep the structural probe (their metadata is already in it).

### 4. Safe X/Y fallback — prefer a non-monotonic signal
`skills/_shared/curve_fitting_tools.py` — `_choose_xy_indices`
- The monotonic branch used the first monotonic column as X then blindly took the next column as Y — often a second axis (e.g. a time channel beside temperature). Now picks X = first monotonic column and Y = first **non-monotonic** column, falling back to "next column" only if all candidates are monotonic. Last-resort deterministic fallback (the model owns this via `column_mapping` when names exist); reasons about column shape, not names → technique-agnostic.

### 5. LLM-normalizer load fallback
`agents/exp_agents/curve_fitting_agent.py` — `_normalize_messy_file` + first-spectrum load
- When deterministic `load_curve_data` fails (e.g. a metadata block above a plain non-`#` header that `np.loadtxt`'s fixed `skiprows∈{0,1}` attempts can't skip), the agent routes the file through the lossless, round-trip-verified split (`utils/file_prep.prepare_inputs`) into a clean data file, then retries the load. The normalized path replaces the input downstream so per-spectrum series loads use the clean file. On any failure the original load error is returned unchanged. Embodies "model owns interpretation, code owns value-extraction + verification" — rather than grow the deterministic attempt-ladder per format.

### Validation
- Unit + full orchestrated end-to-end (refinement binding through to the locked fit, R²=1.0).
- 11-case sniff regression+recovery matrix (plain/headerless/`#`/`##`/`%`/gnuplot/`savetxt`, 2..N columns).
- 21-case cross-technique live stress matrix (thermal, photoemission, diffraction, vibrational, magnetic-resonance, optical): 21/21 loaded correctly and locked the intended Y channel; includes normalizer fallback, safe headerless heuristic, and name-based channel selection (e.g. real-vs-imaginary, a named channel among several, signal-vs-baseline).
- Binary-format split regression (HDF5 attrs / `.npz` keys) confirmed unaffected by the raw-head change.

### Known limits (deliberately left to the model-normalizer rather than chased deterministically)
Numeric/wavelength column headers read as headerless; EU `;`-delimiter; spaces-in-names under whitespace delimiters; comma-data saved as `.txt`. Also: split name-capture is prompt-encouraged but not yet asserted post-split (round-trip proves losslessness only) — a candidate follow-up.

### DIFF
--- a/scilink/agents/exp_agents/controllers/curve_fitting_controllers.py
+++ b/scilink/agents/exp_agents/controllers/curve_fitting_controllers.py
@@ -1953,6 +1953,13 @@ class HumanFeedbackRefinementController:
             f"Parameters: {', '.join(state.get('parameters_to_extract', []))}\n"
             f"Strategy: {state.get('fitting_strategy', 'N/A')}"
         )
+        # For >2-column inputs, show the currently selected columns so the
+        # feedback (e.g. "use the other columns") is anchored to what is wrong.
+        if state.get("column_info") and state.get("column_mapping"):
+            current_plan += (
+                f"\nColumn Mapping: {json.dumps(state['column_mapping'])}"
+                f" — {state.get('column_mapping_note', '')}"
+            )
 
         prompt = [
             self.instructions,
@@ -1964,6 +1971,7 @@ class HumanFeedbackRefinementController:
             f"\n## User Feedback\nAdjust the plan based on this feedback: \"{feedback}\"",
         ]
 
+        _append_column_structure(prompt, state)
         _append_objective_context(prompt, state)
         _append_fit_domain_guidance(prompt, state)
 
@@ -2006,6 +2014,21 @@ class HumanFeedbackRefinementController:
         state["parameters_to_extract"] = result.get("parameters_to_extract", state.get("parameters_to_extract", []))
         state["fitting_strategy"] = result.get("fitting_strategy", state.get("fitting_strategy"))
         state["literature_query"] = result.get("literature_query", state.get("literature_query"))
+
+        # Re-extract the column decision so a feedback-corrected X/Y choice is
+        # honored at lock time. Mirrors _plan_analysis; guarded so a refinement
+        # that omits column_mapping keeps the prior selection rather than nulling
+        # it (the locked fit reads state["column_mapping"] via _resolve_column_mapping).
+        if state.get("column_info"):
+            state["column_mapping"] = result.get("column_mapping", state.get("column_mapping"))
+            state["column_mapping_note"] = result.get(
+                "column_mapping_note", state.get("column_mapping_note", "")
+            )
+            if state.get("column_mapping"):
+                self.logger.info(
+                    f"  Column mapping (refined): {state['column_mapping']} "
+                    f"— {state['column_mapping_note']}"
+                )
 
         # Re-extract series plan (may have been updated or removed)
         self._extract_series_plan(state, result)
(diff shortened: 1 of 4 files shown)


sarah-allec (approved):
Looks good! One small non-blocking recommended change for a follow-up: `_normalize_messy_file` pulls the private `_probe_file` from `meta_agent`, so an analysis agent now depends on the meta layer; since `prepare_inputs` already lives in `utils.file_prep`, moving the probe helper there too would keep the dependency pointing down at utils.

PR #299, opened 2026-06-16
structure-gen: emit engine-neutral coordinates instead of a VASP POSCAR

sarah-allec:
## Summary

Structure generation produced a VASP **POSCAR** even for systems bound for other engines — e.g. a condensed-phase box destined for classical MD (LAMMPS/GROMACS) was written in a VASP-specific format. Structure generation produces *coordinates*; the engine-native input (a VASP POSCAR, a LAMMPS data file, …) is built downstream by the engine step. So the generation stage should stay engine-neutral.

This makes the output format track structure type rather than defaulting to one engine's format:

- **crystal / condensed → extended XYZ** (`.extxyz`) — carries the cell + PBC, and for crystals any selective-dynamics constraints survive as a per-atom column (which CIF cannot represent).
- **molecular → `.xyz`**, **biomolecular → `.pdb`** — unchanged; already neutral and natural to those types.
- No structure-generation class emits a VASP POSCAR anymore.

## Changes

- `structure_generation/crystal` + `condensed` skills: `output_format` → `extxyz` (frontmatter + Output prose; the prose now tells codegen to write extended XYZ and *not* a POSCAR, since the engine-native input is downstream).
- `structure_pipeline`: engine-neutral default `extxyz` (was `POSCAR`) when a class declares no `output_format`; docstring updated.
- `md_simulation_agent.analyze_system`: try a generic `ase.io.read` first (handles extxyz / xyz / cif / pdb / POSCAR) before the LAMMPS-specific reader and the LLM fallback, so neutral coordinates are parsed deterministically; system-info construction factored into `_info_from_atoms`.
- `simulation_orchestrator_tools`: defensive structure-path fallback uses the neutral default filename, not POSCAR.
- `val_agent`: stats header no longer says "computed from POSCAR".

The periodic-DFT agent needed **no change** — its prompt already embeds the structure as format-agnostic "structure file content" (its docstring accepts xyz/CIF), so the LLM builds the VASP inputs from extended-XYZ coordinates just as it did from POSCAR.

## Verification

- Offline suites green (structure backends / matching / router / output-format); new `tests/test_structure_output_format.py` locks the "no class emits POSCAR" decision.
- **Live smoke** (opus-4-8): a crystal (diamond Si) and a condensed water box both generate valid `.extxyz` with cell + PBC that ASE reads back — confirming the prose-guided codegen emits the new format correctly.
- The one affected live test (`test_mp_resolver` full-workflow) is now format-agnostic; the generator-direct e2e case keeps its explicit POSCAR request and is unaffected.

### DIFF
--- a/scilink/agents/sim_agents/md_simulation_agent.py
+++ b/scilink/agents/sim_agents/md_simulation_agent.py
@@ -132,31 +132,45 @@ class MDSimulationAgent(SimulationAgent):
             except Exception as e:
                 self.logger.warning(f"Tool analysis failed: {e}")
 
+        # Generic ASE read first — handles the engine-neutral coordinate formats
+        # structure generation emits (extxyz / xyz / cif / pdb) as well as POSCAR.
+        try:
+            from ase.io import read as _ase_read
+            atoms = _ase_read(structure_file)
+            if len(atoms) > 0:
+                return self._info_from_atoms(atoms)
+        except Exception:
+            pass
+
+        # LAMMPS data files (ASE can't auto-detect these by extension).
         try:
             from ase.io.lammpsdata import read_lammps_data
             atoms = read_lammps_data(structure_file, style="full", units="real")
-            ec = {}
-            for s in atoms.get_chemical_symbols():
-                ec[s] = ec.get(s, 0) + 1
-            return {
-                "atom_count": len(atoms),
-                "elements": sorted(ec.keys()),
-                "element_counts": ec,
-                "box_dimensions": atoms.get_cell().diagonal().tolist(),
-                "has_water": (
-                    "O" in ec and "H" in ec
-                    and ec.get("H", 0) >= 2 * ec.get("O", 0)
-                ),
-                "has_ions": any(
-                    x in ec for x in ["Na", "Cl", "K", "Ca", "Mg"]
-                ),
-                "has_organic": "C" in ec,
-                "system_category": "unknown",
-            }
+            return self._info_from_atoms(atoms)
         except Exception:
             pass
 
         return self._llm_analyze_structure(structure_file)
+
+    @staticmethod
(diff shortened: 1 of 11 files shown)


ziatdinovmax:
Thanks for this — the engine-neutral direction looks right. One question on the default DFT path (`method="llm"`): previously the generation step wrote a file named `POSCAR`, which was the structure VASP consumed directly, so input-gen only needed to add INCAR/KPOINTS. Now that generation emits `structure.extxyz`, what produces the POSCAR that VASP requires?

The DFT prompt embeds the structure content with "use as-is, do not modify" and its canonical-filename example lists "INCAR + KPOINTS for VASP" (no POSCAR), so it's not obvious the input-gen LLM is reliably instructed to convert the extended-XYZ coordinates into a POSCAR. Did the live smoke exercise a full crystal→VASP `run_complete_dft_workflow` (`method="llm"`) and confirm a valid POSCAR landed in the output dir? The verification notes mention validating the `.extxyz` readback from the generation stage, but I wasn't sure whether the downstream VASP input-generation step was covered end-to-end. (The `atomate2` path looks fine — it reads via ASE and writes POSCAR deterministically.)

sarah-allec:
Fixed in 91c4c031: the engine's skill now declares its separate coordinate file in frontmatter (VASP → `structure_file: POSCAR, structure_format: vasp`), and the agent reads the generated structure via ASE and writes that file deterministically, overwriting any LLM copy, so positions are exact and the file is guaranteed. The prompt now tells the model the coordinate file is already prepared (generate only INCAR/KPOINTS). Engines that embed geometry in their main input (QE) declare nothing and are untouched, so it stays engine-neutral. Same treatment on `apply_improvements`. Confirmed the llm path lands a deterministic POSCAR, plus an offline test (test_dft_native_structure) that the written POSCAR round-trips to the exact cell/atoms.

ziatdinovmax (approved):

PR #320, opened 2026-06-29
NMR curve-fitting + relaxation skills: new fitters, honesty guards, and generalizable decision rules

ziatdinovmax:
Improvements to the `nmr` and `nmr_relaxation` curve-fitting skill bundles, developed and validated against synthetic spectra and a set of experimental 1D NMR series (solution and solid-state MAS, spin-½ and half-integer quadrupolar nuclei, plus variable-temperature relaxation). **All changes are confined to the two skill bundles** — no agent/pipeline code touched, so other curve-fitting skills cannot regress.

### New tools / diagnostics
- `fit_jcoupled_multiplet` — constrained first-order J-multiplet (doublet/triplet/quartet…), 6 parameters regardless of line count; returns J directly. Stress-tested 80/80 across noise seeds × multiplet shapes; recovers the correct coupling on real coupled spectra.
- `assess_detection` — low-SNR abstention guard (Rose / quantitative-NMR thresholds + F-test + contiguous-feature requirement). Abstains on noise / blank acquisitions; recovers weak but real resonances.
- `fit_quad_czjzek` — Czjzek (GIM) distribution-of-EFG central-transition lineshape for disordered half-integer quadrupolar solids; round-trips its width parameter; does not false-beat single-site `fit_quad_ct` on ordered lines.
- `peak_region_r2` now also returns residual-structure diagnostics, including the SNR-robust `resid_rms_frac` and a `residual_structured` flag.

### Decision-rule reframes, each validated to not over-fire
- **Default to the physical model; fall back on the tool's own flag.** Run the quadrupolar lineshape by default on solid half-integer lines (let `Cq_resolved` decide whether C_Q is reportable); fit stretched-β by default on disordered-solid relaxation (let β decide). On a held-out set of solid quadrupolar spectra, physical-model adoption improved substantially with equal-or-better fit quality, and the agent discriminated disordered (Czjzek) vs ordered single-site automatically.
- **Series: lock the form, not the count.** Let the component count follow the data across a series rather than locking it; a component appearing/disappearing along the sweep is the signal. Re-runs: a composition sweep now reports a second component emerging partway through (single elsewhere); a variable-temperature sweep captures a peak splitting on cooling.
- **Relaxation series: variable- and regime-agnostic.** Don't assume the series variable is temperature; for a temperature sweep don't assume one Arrhenius line or two — report however many regimes (1…K) or continuous curvature the data justify.

### Known limitation (documented)
Multi-regime Arrhenius detection remains an *operational-prompt* gap — the agent should run the piecewise model-selection in its generated code (it is fully codegen-able; no tool needed). Fully deterministic escalation would require the agent's verification loop, which is outside these skill bundles.


ziatdinovmax:
Superseded by #321 — NMR skill work re-based onto main as a clean, single-commit, NMR-only PR (this branch was cut from the OptimizationAgent/BO feature branch and pulled in unrelated commits).

Issue #245, opened 2026-05-31
Curve-fit: a successful-but-degenerate anchor fit (R²<0.1) skips verification and is locked with no recovery


ziatdinovmax:
## Summary

In the curve-fitting series pipeline, the anchor-spectrum verification loop is gated by:

```python
# curve_fitting_controllers.py  (~line 3128)
if _is_anchor:
    if not best_result or not best_result.get("success") or best_r2 < 0.1:
        # "Initial fit failed or R² too low (...), skipping verification"
    else:
        # full verification loop: verifier critique + residual diagnostics
        # + adaptive annealing + under-resolution clause + refinement
```

The problem is the **`or best_r2 < 0.1`** clause combined with what counts as "success". `_fit_single_spectrum` returns `success=True` whenever the script *executes* and emits `FIT_RESULTS_JSON` + a viz — **regardless of the R² value**. So a script that runs cleanly but produces a **degenerate fit (R² ≈ 0)** is `success=True` with `best_r2 < 0.1` → the entire verification loop is **skipped**, and the (bad) script is then **locked** for the regime with **no regeneration**. For a series, that locked script is reused across every spectrum in the regime → the **whole series comes out R²=0**.

The verification machinery that exists precisely to catch "this fit is completely wrong" — the verifier's physics critique, the residual diagnostics (#242), adaptive annealing, the under-resolution clause — is **skipped exactly when the fit is worst**.

## Two recovery loops, and the gap between them

1. **`_fit_single_spectrum` internal script-correction loop** — handles code that fails to *parse* or *execute*. In the observed case it *did* run (rejected a malformed codegen attempt, generated a correction), but the corrected script **executed** and produced R²≈0, which it counts as a successful run.
2. **Outer verification loop** — refines a working fit. It is **skipped** for `best_r2 < 0.1`.

A "ran-but-degenerate" fit falls **between** them: loop 1 considers it executed fine; loop 2 refuses it as too-low. Neither regenerates → the bad script is locked and (in a series) propagated.

## How it was observed

On Bedrock **opus-4-6**, the codegen for a 3-peak series (planner modeled it as 2 Lorentzians + constant baseline) returned a collapsed-indentation script; #240's compile-check correctly rejected it, the correction loop produced a script that executed to R²=0, verification was skipped, the script was locked, and all 3 spectra came out R²=0 (3/3 reproducible).

[...]

PR #328, opened 2026-07-05
XRD: identification (fingerprint COD library, indexing/Le Bail/Rietveld) + in-situ mixture & series quantification

ziatdinovmax:
## Summary

This PR gives SciLink's XRD analysis the two workflows crystallographers actually use, end to end:

**1. Identifying an unknown sample whose phase is probably known** (the everyday case). The agent now works the way a diffractionist does: correct the 2θ scale against an internal standard, extract the measured lines, and match them — intensities included — against a precomputed reference library of 440,378 patterns derived from the Crystallography Open Database, searched all at once in a tenth of a second, offline, with no AI model in the loop. A match is then confirmed by simulating the candidate structure against the measurement and, when precision matters, Rietveld-refined. On a 20-mineral benchmark of real measured RRUFF patterns spanning all six crystal systems, this identifies **16/20 blind** (no composition given; 14/20 before the retrieval fixes below) and **18/20 when the chemistry is known** — with the remaining blind misses being the textbook cases (structurally identical phases that only elemental analysis can tell apart, plus two patterns the tool itself flags as unconvincing rather than over-claiming).

**2. Characterizing a genuinely new phase** (the discovery case). The pipeline follows standard practice: autoindex the peak positions into candidate unit cells; validate each cell with a structure-free Le Bail whole-pattern fit ("does this cell account for every peak?" — with leftover peaks reported as impurity evidence); read the systematic absences to narrow the space group; then Rietveld. Each step was validated against ground truth: Rietveld recovers RRUFF reference lattices to 0.01–0.09%, Le Bail separates true cells (fit 0.96) from wrong cells and subcells (0.21–0.47), and absence analysis ranks the true space group first for the test cases. Structure *solution* (the final step for a truly new compound) is deliberately out of scope and documented as such.

The reference library ships as a one-time ~300 MB download (`scilink fetch-xrd-library`) — no user ever rebuilds it; a committed pipeline script regenerates it from a COD mirror for maintainers and air-gapped sites, and works on any private CIF collection. Everything is layered so the everyday identification workflow needs no compilers; only the GSAS-II-backed tools (profiles, indexing, Le Bail, Rietveld) require the optional Fortran-built `gsas` extra, with the verified install recipe documented.

Also included: a GSAS-II full-profile simulation engine (drop-in alternative to the pymatgen peak-list engine, adapted with permission from collaborator code and independently verified against Bragg positions, systematic absences, structure-factor intensities, and wavelength scaling), and one agent-level bug fix (below) found during live validation.

**Full-dress validation** — all 20 benchmark minerals run through the real agent exactly as a user would (blind, fetched library, one command each): **zero machinery failures across 20/20 runs**; 13 correct identifications, 4 honest non-identifications (the agent declined or hedged and proposed concrete next steps), 3 confident false positives on position-degenerate phases. That last failure mode motivated the final addition: each match now carries **absent-strong-lines evidence** (a candidate whose strong predicted lines are missing from the measurement is an impostor coinciding on the rest), reported per candidate with the agent directed to compare near-ties — shipped evidence-first (ranking unchanged by default) because real-data validation showed a blind penalty would mis-fire on phases with polytype library entries.

**Known limitations, stated plainly:** isostructural phases remain genuinely ambiguous without chemistry (the absent-lines evidence helps the clean cases; an EDX/composition tie-break is the robust answer and is supported as a match filter); autoindexing is single-engine and honestly unreliable for triclinic cells; the reporting gap where a high-FOM match's phase name failed to propagate into the final payload was traced to its root (the interpretation stage reads only two payload fields) and fixed via the emit contract — the final dress runs show zero such gaps.

**3. Mixtures and in-situ series** (added on the same branch after the identification stack stabilized). Real in-situ experiments are mixtures by definition — during any transformation two phases coexist — so the skill now covers the full arc:

- `identify_mixture` — blind multi-phase identification by classical sequential subtraction (identify the dominant phase, subtract its scaled lines with a shared-line guard against over-removal, re-search the residual), confirmed by one joint MILP whose per-phase activation becomes a `confirmed` flag. Validated on summed pairs of real RRUFF patterns against the full 440k library: 5/6 both-phase identifications, with the one miss being the documented sparse-line degeneracy case.
- `track_phase_series` — deterministic phase-evolution tracking for a frame series (no AI model in the per-frame loop): identify each phase where it is purest, then one joint MILP per frame against that fixed endmember set → per-frame shares, the transition window, thermal-expansion drift, and residual alerts that flag frames the endmembers cannot explain. Endmembers can be a database entry, a simulated pattern, or — essential in practice — a measured pure frame itself (validated on a real 38-frame lab dehydration series where the product phase exists in no database: the transition window is recovered exactly, corroborated by an independent MCR decomposition to a mean share difference of 0.026, and the tool runs at 0.16 s/frame).
- `refine_rietveld_multiphase` / `refine_rietveld_series` — quantitative weight fractions, the standard in-situ deliverable. Single patterns refine all phases jointly under a staged protocol; series runs warm-start each frame from its last converged predecessor. Validated against **certified ground truth** — the IUCr CPD quantitative-phase-analysis round robin (real lab patterns, weighed three-phase compositions): **MAE 1.4 wt% over 24 phase determinations**, on par with the original round robin's human participant median, with 1 wt% trace phases recovered to a few tenths. The one divergent sample (95% dominant + two traces) self-flags and is recovered by the exposed `initial_fractions` knob.

Getting the quantification trustworthy surfaced two failure classes that only certified real mixtures could catch, both fixed at the source: **CIFs declaring ionic species (Ca2+/F−, ubiquitous in COD) silently poisoned every GSAS-backed refinement** (identification never noticed — the kinematic engine is immune; the canonicalizer now strips oxidation states, with the pure-fluorite round-robin pattern flipping from Rwp 81% to 34%), and **COD's id-ascending retrieval surfaced century-old entries** (idealized coordinates, no displacement parameters) as Rietveld starting models — retrieval now prefers the newest deposition within the same composition and stoichiometry, after which certified-mixture quantification works first-try with top-ranked entries (33.8/33.0/33.3 wt% vs weighed 31.4/34.2/34.4).

**Final full-dress validation** (all 20 benchmark minerals through the real agent, blind, on the exact merged surface): **17–18/20 correct across the two final runs (best ever recorded; the pre-branch baseline was 12–13), zero machinery failures, zero reporting gaps**, with the hard-tier silicates going 6/6 including the first-ever correct kyanite identification. The remaining misses are the documented position-degeneracy class (e.g. fluorite vs elemental silicon — F-cubic cells 0.03 Å apart). Live mixture and series runs through the real agent confirm tool routing end to end; the 20-mineral benchmark itself now ships as a re-runnable package (data + runner + slug-stripped grader + gate + run history) so any future skill change can be regression-tested with one command.

---

## Technical details

*For AI review agents; dense references.*

### New tools (all TOOL_SPEC-registered under the `xrd` skill, `scilink/skills/structure_matching/xrd/`)

| Tool | File | Notes |
|---|---|---|
| `search_match_pattern` + `build_fingerprint_library` + `fetch_fingerprint_library` | `fingerprint.py` | Hanawalt keying: candidate's top-`n_key_lines` d-lines (stored wavelength-free) must land within `tol_deg` of the query's top-`n_query_lines` measured lines, vectorized over the whole parquet; shortlist scored by existing `score_xrd_match_robust(algorithm='hanawalt')` via the `exp_peaks` path (the profile path re-extracts peaks and returns none for a peak-list input). Library build curation: `max_sites=500` (kinematic cost ~ sites × reflections; one uncapped COD chunk burned 11+ CPU-h), `min_lines=3`, dedup by (reduced formula, rounded sorted edges, volume). Loader resolution: explicit path → `SCILINK_XRD_FINGERPRINT_DB` → `~/.scilink/xrd_fingerprints/`. Top matches materialize `structure_path` via `fetch_cod_cif` (COD ID; `_backends/cod.py`) because the distributed artifact's `cif_path` column points at the deleted build tree by construction. |
| `calibrate_zero` | `calibrate_zero.py` | NIST SRM cells (Si 640f, LaB₆ 660c, corundum 676a); reference lines from lattice+extinctions (positions-only); fits Δ2θ = zero + disp·cosθ. Documented caveat: the two terms trade off over a typical angular range — `corrected_peaks` (the sum) is the trusted output; `zero_offset` alone only from `fit_displacement=False`. |
| `index_pattern` | `index_pattern.py` → `_gsas_engine.index_pattern` | GSAS-II ITO (`GSASIIindex.DoIndexPeaks`; peaks as 9-col Index-Peak-List rows sorted by descending d; controls `[flag, zero, Nc/No, V₀]`). Default **staged symmetry descent** (cubic → tet/hex/trig → ortho → mono; stop at M20 ≥ `stage_stop_m20`=10 with X20=0) — M20 is not comparable across symmetry levels; a spurious monoclinic V≈560 Å³ cell at M20≈97 outranked true tetragonal V=71.5 at M20≈19 in a live trace. `alias_m20_factor=2` bucket (de Wolff): within 2× M20, smallest cell wins (a √5×√5×2 supercell inflates M20 via extra parameters). Physical floors: volume ≥ 15 Å³, min edge ≥ 2 Å (GSAS `halfCell` emits arithmetic subcells below its own 2.5 Å search floor). `seed` (stdlib `random`, state saved/restored) — 6/6 seeds recover the true cassiterite cell from real measured peaks. |
| `validate_cell_lebail` | `validate_cell.py` → `_gsas_engine.lebail_fit` | Le Bail via per-histogram HAP flag; one normal cycle first (builds the reflection list — flag-before-first-cycle raises `'RefList'`), then background → zero → **mustrain (load-bearing: true-cell corr 0.65→0.96 without/with)** → cell → `extra_cycles` extraction iterations. Dummy one-atom CIF with generic SG per Bravais (`_BRAVAIS_TO_SG`) — intensities are free so atoms are irrelevant, but the SG carries the centering extinctions. Verdicts on real cassiterite: true 0.963 (cell refines to 0.01% structure-free) / wrong 0.467 / half-subcell 0.214; supercells fit by construction (superset) → smallest-fitting-cell rule. `observed_peaks` → `unaccounted_peaks` read from the post-fit reflection list (caught a genuine non-SnO₂ line at 24.04° in RRUFF R040017 plus a planted 31.0°). |
| `determine_space_group` | `determine_space_group.py` | Pure numpy (reciprocal metric tensor). Assign observed peaks to P-lattice hkl at the validated cell (`tol_deg=0.1`, tight because the cell is refined); grade International-Tables conditions: centerings I/F/A/B/C/R-obv, axial screws (h00/0k0/00l × 2n/3n/4n), zonal glides (0kl/h0l/hk0). VIOLATED only when every plausible assignment of an observed peak is forbidden; `min_evidence` testable reflections for "obeyed"; unassignable peaks quarantined. Output = evidence + frequency-ranked consistent common SGs (`_COMMON_SGS`), explicitly non-exhaustive (extinction-symbol twins undecidable from powder). Si → F-family (I/R violated); real cassiterite → all centerings violated, P4₂/mnm ranked first, impurity line quarantined. |
| `refine_rietveld` | `refine_rietveld.py` → `_gsas_engine.rietveld_refine` | Staged: background+scale → Zero → isotropic **mustrain only** (mustrain+size is over-parameterized — cubic fluorite collapsed with both, converges with mustrain alone) → cell. CIF canonicalized **with** symmetry (`_canonicalize_cif(symprec=0.1)`): a P1 phase silently refuses cell refinement in GSAS-II (varyList gets no `::A*` terms) — caught by a perturbation test after the first implementation "recovered" cells that were never being refined. `converged` = final profile-corr ≥ 0.5 and not degraded vs best prior stage; `profile_corr` computed from final Yobs/Ycalc (never None; diverged ⇒ ~0). Perturbation sweep: ±0.5% converges to truth, +1–2% correctly flagged diverged; adversarial wrong-phase 3/3 flagged; RRUFF cells recovered to 0.01–0.09% (sorted-edge compare). `converged` documented as necessary-not-sufficient (triclinic feldspar passed at 39% cell error with corr 0.85). |
| `simulate_xrd_pattern(engine='gsas')` | `simulate_xrd.py` + `_gsas_engine.simulate_gsas` | Registered in `_ENGINES`; returns the pymatgen contract keys (peak-picked) + `profile_two_theta/intensities`. Knobs `crystallite_um`, `peak_rel_height` exposed through the dispatcher (pymatgen engine absorbs them). Physics verified independently: Si/ZnO/anatase positions ≤0.02°, Fe-BCC absences, NaCl I(200)/I(111)=10.9, Cu/Mo/Co wavelengths, broadening knob 0.04°→0.20° FWHM. Origin-ambiguity fix: symbol-only I4₁/amd CIF gave Fcalc(200)=0 until canonicalized via pymatgen (P1 for simulation; symprec for Rietveld). |

### Backends / infrastructure

- `QuerySpec` accepts chemistry-free specs when `lattice_param_ranges` present (`_backends/_base.py`); local-CIF + COD serve cell-only queries (COD via REST cell windows + exact post-load `_lattice_within_ranges`, which now also accepts permutation-invariant `volume`); MP skips cell-only (chemsys-keyed). Cell-only COD ranking by volume-proximity over a 3×top_n pool (fewest-elements ranking made a blind agent conclude elemental Te).
- Library pipeline `scripts/build_cod_fingerprints.py`: shard-at-a-time rsync → 8-worker fingerprint (workers×8 chunks, `chunksize=1` — per-entry cost varies by orders of magnitude) → shard parquet → delete CIFs; resumable via `state.json`. Full run: 9 shards, 448k fingerprinted, 440,378 after cross-shard dedup, 300 MB, 0 chunk errors.
- Artifact: GitHub release `xrd-fplib-v1` (sha256 pinned in `fingerprint.py`); `scilink fetch-xrd-library` CLI (`cli/fetch_data.py`) — explicit download by design; fetch validated end-to-end against the live release.
- `docs/xrd_identification_setup.md`: tier model (structure-matching → library → gsas/Fortran), tool→dependency matrix, verified GSAS build recipe incl. every observed pitfall (conda-run activation/`CONDA_BUILD_SYSROOT`, numpy ABI ordering, graceful degradation), licensing notes (COD open w/ citation; no ICDD).

### Agent fix (`controllers/curve_fitting_controllers.py`, one guarded line)

`if r2 > best_r2:` → `if best_result is None or r2 > best_r2:` at the initial-fit ranking. Matching-type skills mandate no R²; the residual-diagnostics backfill then attaches a deeply negative overlay-R² (observed −4.28 — a stick overlay is not a curve fit) which lost to the −1.0 sentinel, silently discarding the run's only successful result ⇒ "no successful result" ⇒ plan-conformant scripts failed while nonconformant ones passed. Inert unless `best_result is None and r2 ≤ −1` (previously the pathological state); also repairs the latent curve-fit case per the adjacent #245 rationale. Verified: stash-compared offline sweep (646 passed; 18 failures + 20 errors byte-identical without the fix — missing `sim` extras in the test env); live blind rerun 3/3 through the previously fatal conformance→correction path.

### Benchmarks (real measured RRUFF patterns, stratified 20 across all six crystal systems)

- Fingerprint blind, deterministic, full library: **14/20 top-1 / 15/20 top-5** (hydrate/occupancy-tolerant grading), 1.4 s cold / 0.12 s warm per query. Chemistry-known agent workflow: 18/20. LLM-driven *indexing* chain across five controlled rounds: 0/20 → 1/20 (why the fingerprint route is primary for known phases; indexing remains the new-phase path).
- Live end-to-end full-dress (real `CurveFittingAgent`, blind, `skill=xrd`, bedrock, all 20): machinery 20/20 clean (0 crashes / 0 discarded results / verifier on `figure_of_merit` throughout); identification 13 correct / 4 honest declines (incl. Kyanite correctly rejecting its own 0.62 marginal match) / 3 confident false positives (Fluorite→FeAgS₂ 0.89, Sphalerite→ErGdO₃ 0.97, Wollastonite→Pb-organic 0.73) — all position-degeneracy, matching the offline profile.
- Absent-lines evidence (final commit): `frac_strong_lines_absent`/`absent_strong_lines` per match + `adjusted_score` with `absent_lines_penalty` (default 0 — validated on real data: kills the FeAgS₂ impostor (0.86 absent vs truth 0.00) but inverts on polytype-rich truths (ZnS wurtzite entries), hence evidence-first + prose-directed relative comparison, not a blind cutoff). Offline benchmark re-ran identically with the default (14/20).

### In-situ / mixture machinery (same branch, commits `4657bec..a55ea45`)

| Tool | File | Notes |
|---|---|---|
| `identify_mixture` | `identify_mixture.py` | Sequential subtraction over the fingerprint library: per-iteration `search_match_pattern` on the residual; accepted phase's normalized predicted intensities scaled by the median matched ratio and subtracted; a matched peak SURVIVES with its remainder when ≥ `shared_line_floor` of the peak (validated on the genuine NaCl(111) 27.37°/rutile(110) 27.45° coincidence). Acceptance tie-break: within `absent_tiebreak_band` of the top FOM, fewest absent strong lines wins — real-data eval showed dense-line junk (organics whose line forests blanket any residual) out-FOM the true minority phase by ~0.05 while showing 0.1–0.25 of their own strong lines absent vs truth's 0.0 (fixed 3/6 → 5/6 on summed-RRUFF pairs). Ensemble confirmed by `score_xrd_match_multiphase`; per-phase `confirmed` flag from the joint activation. `fom_threshold` default 0.5 documented as deliberately laxer than single-phase 0.7 (the FOM's coverage term is depressed by the other phases). |
| `track_phase_series` | `track_phase_series.py` | One `score_xrd_match_multiphase` per (strided) frame against a fixed endmember set; three reference forms normalized internally (`source_id` → library row, `sim_two_theta/sim_intensity`, empirical `positions/intensities`). Residual per frame = 1 − Σ shares (exp peaks matched at most once ⇒ disjoint). `phase_events` (onset/final/max), `coexistence_frames`, `residual_alert_frames` ≥ `residual_alert_frac`. The scorer's per-phase lattice pre-fit was mixture-corrupting (exp-intensity weighting let a 70/30 mix stretch the minority 3.8% onto the majority's strong lines, deactivating it in the MILP) — `_fit_lattice_scale(weight_by='sim')` on the multiphase path only, single-phase path untouched, regression-locked. Real-data: 38-frame dehydration series transition at the known window; MCR cross-validation mean Δshare 0.026; both transition detectors independently name the same frame. Documented limits: shares are screening proxies; position-based tracking cannot separate order-disorder transitions; profile-level crystallinity loss (melt) is invisible to normalized peak lists. |
| `refine_rietveld_multiphase` / `refine_rietveld_series` | `refine_multiphase.py` → `_gsas_engine` | Staged single-histogram protocol (background+scale → zero → phase fractions [all but first; N fractions + histogram scale is rank-deficient] → mustrain → cells ONE PHASE AT A TIME dominant-first — a joint cells stage let a 10 wt% phase collapse to a=0.008 Å while global corr stayed 0.993). Weight fractions w_i = f_i·M_cell,i/Σ (`_phase_cell_mass` from GSAS atom records). Per-phase >10%-axis cell-runaway guard reports the input cell + `cell_runaway` flag. Series = per-frame loop warm-started from the last CONVERGED frame (fractions via `initial_fractions`, lattices via rewritten starting CIFs); GSAS-II native sequential mode was prototyped and REJECTED with evidence (refines all flagged parameters at once with no staging — Dij exploded to −10⁶; the cubic HStrain 'eA' term is degenerate by construction and got whole phase blocks dropped as singular). Round-trips through GSAS's own forward calculation: single-pattern fractions ~10⁻³/cells <10⁻⁴ Å; 3-frame crossfade with programmed thermal expansion: fractions <0.03, cells <2×10⁻³ Å, monotone expansion recovered. |

### Engine & backend fixes surfaced by certified-mixture validation

- **`_canonicalize_cif` strips oxidation states** (`_gsas_engine.py`): ionic species declarations send GSAS-II down a charged-scattering-factor path with a catastrophically wrong intensity model — isolated by single-variable mutation on the round-robin's pure-fluorite pattern (identical structure: corr 0.73/Rwp 81 with the oxidation loop, 0.97/34 without). pymatgen's kinematic engine is immune, so identification was blind to it. Regression test asserts charged fixtures canonicalize neutral.
- **COD ranking newest-first within ties** (`_backends/cod.py`, both `_search_db` and `_search_web`): id-ascending tie-break surfaced early-1900s entries (no ADPs, idealized coordinates) as refinement starting models — a 1911 zincite entry collapsed its phase to 0 wt% where the modern entry lands within ~1 wt%. Composition-tightness and stoichiometry-simplicity ranks above the tie-break untouched (the Magnéli-class identification fixes preserved); no-regression: 60 tests + offline blind byte-identical + certified mixture now first-try.
- **Emit contract** (`xrd.md`): the synthesis stage builds its context from only `parameters` and `fit_quality` in `FIT_RESULTS_JSON` — guidance that routed the identity into `best_match` produced high-FOM runs whose reports could not name the phase (1 → 3 → 11 cases as adoption spread; generated scripts verified correct, loss was downstream). Identity now goes in `parameters.identified_phase` (+ space group, database id, runner-up); dress-run gap count: 0. The benchmark grader classifies `UNNAMED (match found, identity not reported)` separately from wrong/declined so this class trips the gate loudly.

### Quantification benchmark (IUCr CPD QPA round robin, certified weighed compositions)

Fetched from the archived IUCr data kit (samples 1a–1h, corundum/zincite/fluorite, real lab CuKα 5–150°, weighed truth; `ground_truth.json` + `RESULTS.md` alongside the data). Literature starting structures, default knobs: 8/8 samples refine, **MAE 1.43 wt% over 24 determinations** (original round-robin participant median ≈1–2 wt%); traces at the 1 wt% level recovered to 0.1–0.7 wt%; the +3 wt% corundum systematic in balanced mixes reproduces the round robin's own documented microabsorption bias (Brindley correction noted as a future knob). The 95%-dominant sample diverges from an all-equal start, self-flags (corr 0.82 vs ~0.97), and is recovered by `initial_fractions` (94.5/1.3/4.2 vs weighed 94.3/1.4/4.3).

### Live agent validation of the new tools

Blind summed-RRUFF mixture: the agent routes to `identify_mixture` unprompted, finds the dominant phase family, and reports the unexplained remainder as evidence of a second phase rather than overclaiming. Real dehydration series: endpoints-first behavior with the transition window bracketing the known truth, the low-T hydrate composition surfaced, and honest declines where organic database degeneracy defeats naming. Quantification: the agent now reaches the Rietveld tier, builds polymorph-locked candidate pools, implements the prescribed swap-retry loop, and reports fractions with quality caveats; weak-phase accuracy remains sensitive to database-entry measurement conditions (e.g. a high-pressure corundum determination as the starting model) — documented below as the follow-up, with the deterministic loop's certified first-try result as the reference behavior.

### Re-runnable benchmark package

`benchmarking_for_paper2/xrd_id_benchmark/`: the 20 stratified RRUFF datasets, live runner, slug-stripped text-only grader (with the unnamed-vs-wrong classifier), 1-minute deterministic fingerprint eval, the 4-case mixture/series live suite, and a README carrying the merge gate (20/20 machinery-clean, ≥13 correct, 0 unnamed, ≤3 known-class FPs) plus the full run-history table including the emit-channel regression and its fix.

### Known limits / follow-ups

Isostructural/position-degenerate phases still need a chemistry tie-break (EDX/composition as a match filter — the top queued follow-up); database-entry *condition* sanity for Rietveld starts (reject high-pressure/high-T determinations by start-cell vs ambient lattice — queued); quantitative fractions assume every crystalline phase modeled, no amorphous content, minimal microabsorption (documented per-tool); multi-engine indexing and structure solution deferred with rationale in-skill; `pymatgen<2026` pin unchanged.

---

## Added since initial review — in-situ coupling, figure-grounded readout, fast quantification

Three capabilities were built and validated on top of the identification stack above.

**In-situ / operando reconciliation.** For a temperature- or time-series, two complementary passes now run and are *joined*: profile fitting (peak positions, widths, areas — *how* the structure evolves, database-independent) and phase identification (*which* phases, database-dependent). A technique-agnostic reconciler attributes the fitted-feature trends to the identified phases, cross-checks the transition each pass finds independently, and produces one coupled HTML report with an agreement verdict — *consistent* when the two corroborate, and an honest *one-sided* / *divergent* flag when identification can't name a phase (organics, novel products) or disagrees with the robust profile fit. Validated end-to-end through the agent on real synchrotron series: ibuprofen dehydration (profile 46 °C, corroborated) and an IMAuCl₄ precursor ramp (profile locates the transition; identification honestly one-sided where the products aren't in any database — precursor → metallic gold).

**Figure-grounded interpretation.** The coupled report carries an LLM interpretation that reads the actual reconciled figure, not just the numbers — which surfaced and fixed a latent orchestrator bug: images returned by tools (e.g. `preview_image`) were handed to the model as plain-text base64 and never actually seen; they are now delivered as real image content on providers that support it (verified on Bedrock/Claude).

**Fast phase quantification (NNLS).** A new `quantify_phases_nnls` tool gives phase *fractions* over a candidate shortlist by non-negative least squares — the fast rung between mixture identification (*which* phases) and full Rietveld QPA (rigorous weight fractions). It deconvolves overlapping mixtures in one joint fit, screens absent candidates toward zero weight, and abstains when a phase is missing rather than fabricating fractions. Live-validated: the agent plans and calls it on a six-candidate mixture screen.

<details><summary>Technical details</summary>

- **Reconcile core** `skills/_shared/_reconcile.py` (technique-agnostic — reusable by NMR/Raman/XPS; XRD consumer `structure_matching/xrd/reconcile_series.py`): layer-1 feature tracking with auto-scaled tolerance (no technique-specific x-unit default) + a single-crossover transition model (named seam for a future multi-regime model); `reconcile_analysis_dirs` orchestrator seam; `render_reconcile_report` self-contained HTML with an attributed Interpretation section. Orchestrator tools `reconcile_series` + `finalize_reconcile_report`. Deterministic join (no interpreter LLM); the narrative is authored by the orchestrator and re-rendered without recompute.
- **Orchestrator image vision** `utils/tool_media.py`: an image-bearing tool result becomes a multimodal message on capable providers (Claude/Bedrock, Gemini), with a plain-string fallback (no regression on OpenAI-compatible path); saved history is sanitized (no base64 on disk); over-large images are capped to the shared `_MAX_IMAGE_DIM`. Wired into the analysis + meta orchestrator loops.
- **NNLS** `structure_matching/xrd/quantify_nnls.py`: pymatgen-simulated + pseudo-Voigt-broadened design-matrix columns, SNIP background, ZMV-corrected weight-fraction estimate, and a worst-99.5th-pct-residual-over-MAD-noise abstention. Adoption required routing the skill's *planning* quantification branch through NNLS (tool-in-inventory + codegen recipe + code-in-markdown were each insufficient — adoption is set at the planning layer).
- **Tests**: `test_shared_reconcile.py`, `test_xrd_reconcile_series.py`, `test_orchestrator_reconcile.py`, `test_tool_media.py`, `test_xrd_quantify_nnls.py`.

</details>


tkavinchakravarthy-del (approved):
All sounds good! Eager to see how they perform on the real XRD data :)

PR #343, opened 2026-07-09
Add `exafs_simulation` skill bundle

jenna1701:
## Summary

- Adds `scilink/skills/exafs_simulation/` as a new simulation-domain skill bundle covering end-to-end EXAFS spectrum computation from crystal structures.
- Adds `scilink/skills/machine_learning_potentials/SKILL.md` as a router that directs an agent to the correct backend-specific MLIP skill.

### `exafs_simulation/` skill bundle

Three sub-skills, following the standard `<domain>/<name>/<name>.md` + sibling `.py` layout:

**`exafs_workflow`** — the primary skill. Covers the full six-stage pipeline:
1. Input validation and MD engine selection
2. Cell relaxation (`FrechetCellFilter` + `FIRE`)
3. NVT molecular dynamics
4. FEFF input generation (carving local environments from trajectory frames)
5. Batch FEFF execution
6. chi(k) averaging and convergence plotting

The skill is **simulation-engine-agnostic**: the MD calculator for stages 2–3 is provided by whichever simulation skill is co-loaded at runtime (`machine_learning_potentials/*`, `molecular_dynamics/lammps`, or `periodic_dft/{vasp,qe}`). No backend-specific calculator code lives in the EXAFS skill itself, so adding a new MLIP backend requires zero changes here.

**`generate_feff_input`** — FEFF card parameter guidance (EDGE/HOLE selection, SCF, RMAX, CONTROL, CORRECTIONS) and batch generation from MD trajectories. Can be loaded standalone when working from an existing trajectory.

**`feff_tools.py`** — three registered `TOOL_SPEC` helpers for the `simulation` agent surface:
- `carve_out` — extracts a local atomic cluster from a periodic structure for FEFF input
- `generate_feff_inputs_from_trajectory` — samples frames at regular intervals, carves environments, and writes `feff.inp` for each
- `average_chi` — reads all `chi.dat` outputs from a FEFF batch run and returns a converged mean chi(k)

**`SKILL.md`** — domain router that lists the two sub-skills and provides a co-loading table mapping use case (magnetic system, large cell, AIMD benchmark, etc.) to the correct combination of skills.

The `planning` section includes FEFF binary detection and installation guidance covering all available versions (FEFF6-lite and FEFF8-lite are free and sufficient for EXAFS; FEFF7–9 require a license). A `find_feff_binary()` snippet in the `implementation` section resolves the binary at runtime via `$FEFF_BIN`, `$FEFF_DIR`, `$PATH`, and a site-default fallback, raising with the download URL if nothing is found.

### `machine_learning_potentials/SKILL.md`

A top-level router for the existing MLIP skill family. Summarises all six sub-skills (`general`, `mace`, `chgnet`, `deepmd`, `uma`, `orb`) with a priority-ordered selection guide, so an agent landing on the `machine_learning_potentials` domain can immediately identify the right backend skill without reading all six files.

## Test plan

- [ ] `python -c "from scilink.skills.exafs_simulation.exafs_workflow.feff_tools import carve_out, generate_feff_inputs_from_trajectory, average_chi; print('OK')"` — confirms TOOL_SPEC helpers import cleanly
- [ ] `load_skill('exafs_simulation/exafs_workflow')` and `load_skill('exafs_simulation/generate_feff_input')` both return all five canonical sections with no `extras` warnings
- [ ] End-to-end: load an Fe₂O₃ CIF, co-load `exafs_simulation/exafs_workflow` + `machine_learning_potentials/chgnet`, run relaxation + short NVT, generate FEFF inputs, confirm directory structure matches expected layout


sarah-allec (changes requested):
The shape of this is perfect: the calculator coming from whichever sim skill is co-loaded is the right call, and the FEFF card guidance is very useful. A few requested changes before merging:

1. `average_chi` averages χ by array index rather than by k: `np.array(k_all)` → `nanmean(axis=0)` → `emits k_arr[0]` as the grid. A different-length k-grid raises on the ragged array; a same-length-but-different one silently averages misaligned spectra. The `if k[0] == 0.05: chi = [np.nan] + chi` branch is already shimming exactly that. Can we interpolate onto a common grid instead? 

2. Both `SKILL.md` files are dead: the loader only resolves `<domain>/<name>/<name>.md` (`loader.py:230-234`, `:334`) and nothing in the package reads `SKILL.md`. You don't need them — `build_skill_catalog(domain)` in `_shared/_skill_selector.py` already routes per-domain off each bundle's frontmatter `description`, automatically and without drifting like a hardcoded list. I'd drop both and lean on good descriptions.

Smaller:
- `carve_out` silently drops all hydrogen with no way to disable it — worth an `include_hydrogen=False`.
- `_supercell_repeats` sizes off cell-vector lengths and `_check_distance` uses a Cartesian bounding box; for non-orthogonal cells the guard can pass while the cluster is truncated. Hematite is rhombohedral, so your test case may hit it.

Thank you!

ziatdinovmax:
A quick clarification on the skill format.

SciLink skills are based on the Anthropic’s agent skill idea but with some customization to make them more targeted.
Specifically, Anthropic uses `SKILL.md + siblings`, whereas SciLink uses `<name>.md + siblings`. The Anthropic SKILL.md "router" convention is not supported by SciLink. The routing job is done automatically by SciLink's  build_skill_catalog(domain) + the agent-side selector reading each bundle's frontmatter description.

jenna1701:
I've addressed the requested EXAFS changes, including removing the SKILL.md files. Also wired the simulation orchestrator dispatch tools to access MLIP simulation and EXAFS workflows.

**The MLIP tool does not run simulations itself and EXAFS tool doesn't run FEFF itself — consistent with the orchestrator's design for DFT (it generates inputs and the user runs the calculation).** The feff_tools.py TOOL_SPEC helpers (generate_feff_inputs_from_trajectory, average_chi) are registered via the skill registry and callable by the simulation agent when the exafs_simulation skill is co-loaded.


### Full Summary (from Claude):

> average_chi averages χ by array index rather than by k: np.array(k_all) → nanmean(axis=0) → emits k_arr[0] as the grid. A different-length k-grid raises on the ragged array; a same-length-but-different one silently averages misaligned spectra. The if k[0] == 0.05: chi = [np.nan] + chi branch is already shimming exactly that. Can we interpolate onto a common grid instead?

Changed to grid interpolation:
- k_min = max(all starting k) / k_max = min(all ending k) — restricts to the intersection so no spectrum is extrapolated beyond its data range.
- dk = median(all per-file median dk) — robust to the occasional file with a slightly different step; np.arange with a half-step tolerance on the upper bound avoids float rounding cutting off the last point.
- np.interp (linear) is appropriate — FEFF's ~0.05 Å⁻¹ step spacing means interpolation error is far below any physical signal. The interpolation is onto the common grid, not back onto the original, so the resulting array is always rectangular and np.mean (no nan needed) is correct.

Skips files with len(k) < 2 — a file with a single data point can't meaningfully contribute and would break np.diff in the dk calculation.

> _supercell_repeats sizes off cell-vector lengths and _check_distance uses a Cartesian bounding box; for non-orthogonal cells the guard can pass while the cluster is truncated. Hematite is rhombohedral, so your test case may hit it.

- _perp_heights(atoms) — new shared primitive. Computes h_i = |V| / |a_j × a_k| for each of the three cell directions. Returns the true inscribed-face distances, which equal the vector norms only for orthogonal cells.
- _supercell_repeats — now divides by h_i instead of |a_i|. Guarantees the sphere fits.
- _check_distance — now checks min(h_i)/2 >= rmax (inscribed-sphere radius of the centred parallelepiped) instead of the Cartesian bounding-box minimum. The old test was geometrically incorrect for any non-orthogonal cell.

In order to test the workflow I had to wired the orchestrator's dispatch tools to run MLIP simulations and the EXAFS workflow

Refactoring mlip_tools.py to include added MLIPs :
  1. Has per-backend _probe_*() functions — each imports only its own package, inside a try/except ImportError
  2. Has a _BACKEND_PROBES registry mapping backend names to probe functions
  3. Has a check_backends(backends=None) that accepts an optional filter list
  4. Adds _probe_uma (fairchem) and _probe_orb as previously-missing backends

  simulation_pipeline.py
  - Added "machine_learning_potentials": "mace" to _DEFAULT_ENGINE so the pipeline knows its default engine
  - Added a new if scale == "machine_learning_potentials": branch to _generate_inputs that constructs MLIPAgent, infers system_info from the structure file via ASE, calls deploy_pretrained,
  and normalizes the result to the common {status, input_files, entry_file} shape the rest of the pipeline consumes

  simulation_orchestrator_tools.py
  - Added run_mlip_simulation (tool 16): a direct one-call MLIP run setup — selects backend automatically or uses an explicit override, generates the ASE run script, records the structure in
  the session, returns the script path with next-steps guidance
  - Added run_exafs_workflow (tool 17): a multi-stage setup tool that deploys the MLIP, locates the absorber index, detects any installed FEFF binary, maps the edge to the HOLE card number,
  and returns a fully pre-filled step-by-step plan (run MD → generate FEFF inputs → run FEFF → average_chi)
  - Fixed route_simulation description to replace the "LAMMPS/MLIP wiring is in progress" caveat with the actual dispatch table

Updated _ase_runner.py to allow for any ASE-compatible extensions
  - Replaced the unconditional from ase.io.lammpsdata import read_lammps_data import and read_lammps_data(...) call with a format-detecting _load_structure() that uses ase.io.read for everything except files with
  .data/.lammps/.lmp extensions, which are LAMMPS data files that ASE can't auto-detect by extension
  - The LAMMPS path is moved inside an if branch, so CIF, XYZ, EXTXYZ, POSCAR, PDB, and any other ASE-readable format all go through ase.io.read directly
  - The element-remapping logic (needed for LAMMPS type→symbol mapping) is preserved in the LAMMPS branch where it's still relevant

sarah-allec:
Re-reviewed after the updates — the two things that mattered most are genuinely fixed. `average_chi` now interpolates every spectrum onto a shared uniform k-grid before averaging, so the by-index averaging bug is gone; both dead `SKILL.md` files are removed; and the new orchestrator wiring (`run_exafs_workflow`, `run_mlip_simulation`) means the skill is actually reachable now rather than orphaned. Good progress.

A few things surfaced in the newly-added orchestrator code that weren't in the original — none blocking now that the correctness bug is fixed, so your call on address-now vs. follow-up:

The one I'd weight highest is the absorber selection. `run_exafs_workflow` picks `target_atom = absorber_indices[0]` — a single absorber atom, then samples its environment over MD frames. That captures thermal disorder, but EXAFS is an ensemble average over *all* crystallographically equivalent absorbers, and for a multi-site or larger system one atom under-samples the configurational average. For Fe₂O₃ there are many Fe; averaging over frames of Fe #0 is not the same measurement as averaging over the Fe sublattice. Either sample several absorbers (the `carve`/`generate_feff_inputs_from_trajectory` path already loops over frames — extending to a few absorbers is natural) or document that the workflow characterizes a single absorber's thermal environment, so nobody reads the output as a full site-averaged spectrum.

`average_chi`'s parsing still keys on a byte-exact `header_line` and increments `skipped` silently on a mismatch. The k-fix addressed *how* it averages; this is *which* files it accepts — a FEFF version whose `chi.dat` header differs by whitespace yields `n_samples: 0` with no error surfaced. Since it already tracks `skipped`, the low-cost fix is to fail loudly when `skipped > 0 and n_samples == 0` ("parsed 0 of N chi.dat files — header mismatch?") rather than return an empty result quietly.

And there's still no test coverage. `carve_out` and `average_chi` are pure functions, and the k-grid interpolation you just added is exactly the kind of numeric code that silently regresses — a couple of tests on those two would be worth having, especially now that the correctness fix lives there.

Minor, take or leave: `run_exafs_workflow` is really a setup-and-plan tool (it deploys the MLIP, writes the MD script, and returns next-steps instructions rather than executing FEFF) — the name reads as if it runs the whole thing; and FEFF-binary detection is now duplicated between the skill markdown's `find_feff_binary` snippet and the orchestrator's inline `_FEFF_NAMES` walk.

The correctness blocker is resolved, so this is in good shape — let me know if you want to address any of the above in this PR or in a follow-up. Thank you! 😊

sarah-allec (approved):

Issue #359, opened 2026-07-15
Hyperspectral: size-gated decomposition-first staging (mask or skip the per-pixel fit)


ziatdinovmax:
## Motivation

For a very large hyperspectral cube, the expensive stage is per-pixel fitting — and it is frequently attempted before anything establishes *where* (or *whether*) fitting is worthwhile. Observed live (see #358): a fluorescence branch railed a full-frame two-peak fit five times because the emission was weak and confined to a tiny region — information the **decomposition stage already had** (localized abundance) but that the fitting stage never used. The fit was doomed everywhere except a small mask, and the pipeline kept regenerating it.

Decomposition is cheap relative to iterative per-pixel fitting and it answers exactly the two questions that should gate fitting: **where is the signal, and is there fittable structure at all?** Using it to gate the fit — for large data especially — turns a doomed full-frame fit into either a cheap masked fit or an honest decomposition-only result.

## What already exists (so this is gating, not new machinery)

The pipeline (`pipelines/hyperspectral_pipelines.py`) is already `Decomposition -> Interpretation -> refinement decision -> Dynamic Analysis (fitting)`, and:
- fitting is **conditional**: `SelectRefinementTargetController` can return `refinement_needed: False` and no per-pixel fitting runs (`controllers/hyperspectral_controllers.py:1703, 2190`) — so **decomposition-only is already a valid terminal outcome** the agent can choose;
- the reverse `skip_decomposition` mode exists (direct fitting, no decomposition);
- the decomposition contract (abundance maps, component spectra, high-purity regions) already flows into the fitting stage;
- follow-up-reuse already lets a later `run_analysis` consume a prior run's outputs as materials.

What is missing is making the refinement decision **size-aware** and letting decomposition **scope** the fit.

## Proposal: size-gated decomposition-first staging

Above a size threshold (pixels x bands — reuse the size-guard bands), make the refinement/target decision commit to one of:

1. **fit-within-a-mask** — derive a fitting mask/ROI from the decomposition (e.g. the high-abundance region of the relevant component) and require the generated `custom_code` to fit ONLY there; the size-guard target rule ("state which pixels the fit touches") already wants this — here decomposition supplies the mask instead of the LLM guessing it.
2. **decomposition-only** — if decomposition finds no component with fittable per-pixel structure (flat/pure-noise/artifact components), return the decomposition + interpretation as the deliverable and do NOT attempt a per-pixel fit. Honest and cheap.

[...]

PR #362, opened 2026-07-15
Vetted parallel per-pixel fitting tool for hyperspectral codegen (fit_per_pixel)

### LINKED ISSUE #356: Vetted parallel per-pixel fitting tool for hyperspectral codegen (map-over-pixels harness)
## Motivation

Commercial hyperspectral cameras produce cubes an order of magnitude larger than the academic SIs the hyperspectral agent was tuned on. On a live run over a commercial-camera-scale cube (704×704×260 ≈ 129M values, ~495k pixels), one analysis pass completed normally, then a second pass sat in unbounded in-process compute for hours: nothing in the stack scaled the plan or the per-pixel fitting to the data size.

The first half of the fix is the **size guard** (now merged on `feature-fusion-numerics`): a `spatial_bin_factor` knob in the preprocessing strategy, a size-scaling rule for `custom_code` targets, and a SIZE BUDGET block in the codegen prompt. The guard's philosophy is *don't pay the cost when the science doesn't need it*.

This issue is the second half: **pay the cost efficiently when the science DOES demand full resolution.** A per-pixel map over a full frame is embarrassingly parallel (hundreds of thousands of independent, few-ms fits), but the generated code cannot parallelize it today, and no library flag substitutes:

- The hyperspectral custom-code runs in a restricted in-process `exec()` sandbox (`RunDynamicAnalysisController`, `controllers/hyperspectral_controllers.py` — in-process by documented design: no cube serialization per attempt, numpy-array returns).
- `exec`-defined functions cannot be pickled, so generated code can never use `multiprocessing` itself.
- lmfit/scipy parallelize only *within* a single fit (`emcee`, `differential_evolution(workers=)`) — the wrong axis for many small independent fits; and scipy's `workers=` uses a fork-based Pool, a deadlock hazard because fan-out branches run in threads.

## Proposal: a vetted map-over-pixels harness, injected into the sandbox

**1. `TOOL_SPEC` tool** (e.g. `scilink/skills/_shared/parallel_pixel_fit.py`):

```
fit_per_pixel(data, axis, model, mask=None, init="moments", bounds=None,
              n_jobs=-1, chunk_size=2048, time_budget_s=None)
  -> {"maps": {...}, "err_maps": {...}, "coverage": ndarray, "notes": [...]}
```

The load-bearing engineering, all of it portability-driven:

- **Module-level chunk workers** (picklable — mandatory for Windows spawn, and the only way anything crosses a process boundary), dispatching ~2k-pixel chunks to amortize IPC.
- **loky backend**: spawn-like semantics on all three OSes; safe to invoke from worker threads (fan-out branches are threads — `fork` there is a deadlock hazard).
- **Zero-copy input** via joblib memmapping so a 0.25–1 GB cube is not copied per worker (explicit memmap folder + cleanup-with-retry for Windows file-locking quirks).
- **BLAS pinning per worker** (`threadpoolctl` / env) — without it, `n_jobs × BLAS-threads` oversubscription makes "parallel" slower than serial. (macOS Accelerate exposes no thread control; harmless here since per-fit linear algebra is tiny.)
- **Soft time budget**: stop dispatching chunks at the deadline, return partial maps + a coverage mask — degrade instead of dying at the sandbox's exec cap.
- **σ as a first-class output**: per-parameter uncertainties from the fit covariance as `_err` maps, matching the "uncertainties are features, not prose" contract.
- **Declarative model spec**, not callables: compositions of registry primitives (Gaussian/Lorentzian/Voigt/pseudo-Voigt peaks; constant/linear/power-law backgrounds; sigmoid/edge onsets; exponential decays). Exec-defined models can't cross process boundaries, so this is a hard design constraint, not a style choice. lmfit remains the per-fit engine inside workers.
- Every knob exposed and documented (what it does + which way to turn it), robust no-arg defaults; `n_jobs` default capped (e.g. `min(cores, 16)`).

**2. Sandbox injection** (~20 lines): pre-import the vetted callable into the exec globals alongside `np`/`scipy`/`sklearn`/`lmfit`. The sandbox stays import-free; the tool enters under the same trust model as the pre-imported libraries.

**3. Prompt awareness** (~5 lines): list under PRE-IMPORTED FUNCTIONS and reference from the SIZE BUDGET block ("for unavoidable full-frame iterative fits, call `fit_per_pixel(..., n_jobs=-1)`").

**4. Tests**: parallel-vs-serial map equivalence; speedup smoke test; partial-budget return + coverage mask; knob-changes-behavior; called-from-a-worker-thread; NaN/saturated-pixel guards. If Windows is a deployment target this tool's suite is the first candidate for a CI OS matrix.

**5. Live validation**: a full-resolution cube with an objective that legitimately requires single-pixel resolution (the case where the size guard steps aside), showing a full-frame fit completing in minutes at near-linear core scaling.

## Scope boundaries (stated in the tool description)

- Parallelizes **independent per-pixel fits only**. Coupled/global fits (parameters shared across pixels, spatial regularization, MCR-ALS-style constraints) are not embarrassingly parallel and are out of scope.
- Models outside the registry fall back to serial in-sandbox fitting — acceptable, since the size-guard rules already force masking / coarse-to-fine there.

## Relationship to existing work

- Complementary to the size guard (same branch): guard avoids unnecessary cost; this tool makes necessary cost efficient.
- Orthogonal to `docs/hyperspectral_codegen_relocation.md` (preprocessing/denoising ownership; decomposition stays in-loop) and to #221/#219 — no ordering dependency either way.
- Follows the standing tool conventions: adaptive logic in the tool, exposed tunable knobs, deterministic/testable module code as the top rung of the reliability ladder.

ziatdinovmax:
## Summary

When a hyperspectral objective genuinely needs a fitted value at **every pixel of a large frame** — a peak-position map at native resolution, say, because the feature of interest is only a pixel or two wide — the agent has had no good option: the code it generates runs single-process, so a 160k-pixel lmfit loop takes tens of minutes and hits the execution cap. The size guard and the decomposition-first gate (merged earlier) handle all the cases where that cost is *avoidable* (mask it, bin it, decide from decomposition). This PR handles the case where it isn't: a vetted `fit_per_pixel` tool the generated code can call, which fans the independent per-pixel fits out across CPU cores.

In live validation, the agent was given a 400×400×140 cube containing a tilted defect stripe just ~2 pixels wide with a shifted emission line, and an objective forbidding binning. The planner correctly reasoned that the objective overrides the size gate, named `fit_per_pixel` as the method of record, the generated script called it, and the full analysis finished in **6.8 minutes** with the stripe cleanly resolved in the peak-position map — a hand-rolled loop at the same scale is the regime that previously ran for hours.

Fits report their **uncertainties as first-class outputs**: every parameter map comes with a σ map from the fit covariance, plus an R² map and a coverage mask. A soft time budget stops dispatching at a deadline and returns partial maps (with coverage marking what was fit) instead of dying at the sandbox cap — same degrade-never-hang philosophy as the wall-clock budgets.

## Technical details

- **Why a vetted tool and not generated code**: the hyperspectral sandbox is in-process `exec()` by documented design, and exec-defined functions cannot be pickled — generated code can never use multiprocessing itself. `fit_per_pixel` (`scilink/skills/_shared/parallel_pixel_fit.py`) therefore takes a **declarative model spec** (data, not callables) built from registry primitives — gaussian/lorentzian/voigt/pseudo_voigt peaks; constant/linear/quadratic/power_law backgrounds; exponential decay; step_logistic/step_erf onsets — and rebuilds the lmfit composite inside each worker. lmfit remains the per-fit engine.
- **Portability engineering** (per the issue spec): module-level picklable chunk workers (~2k px/task, Windows-spawn-safe); **loky** backend via joblib — spawn-like on all three OSes and safe from worker *threads* (fan-out branches are threads, where fork pools deadlock); explicit once-only **memmap handoff** for cubes > 50 MB (path crosses the process boundary, not the array) with retry-on-rmtree for Windows file locking; **BLAS pinning** per worker via threadpoolctl. All dependencies (joblib/loky/threadpoolctl) already ship transitively with sklearn — nothing new.
- **Initialization**: `init='auto'` (default) runs one serial fit of the field-mean spectrum to seed widths/backgrounds, then snaps each peak's center to the pixel's local maximum — robust to field-wide drift; `'moments'` and `'spec'` variants exposed. Every knob (`mask`, `bounds`, `n_jobs`, `chunk_size`, `time_budget_s`, …) is documented in the TOOL_SPEC with which-way-to-turn guidance; `n_jobs=-1` caps at 16.
- **Wiring is registry-native**: `TOOL_SPEC(agents=["hyperspectral"])` auto-lists the tool in the planner's tool-awareness block and the codegen prompt's REGISTERED TOOLS, and pre-loads the callable into the sandbox globals — zero new plumbing. The single prompt edit is the SIZE BUDGET block (`hyperspectral_controllers.py`), which now routes unavoidable large per-pixel fits to `fit_per_pixel` instead of declaring parallelism impossible; HS goldens regenerated for that deliberate change (curve/image goldens untouched, verified 17/17).
- **Scope boundaries** (stated in the spec): independent per-pixel fits only — coupled/global fits (shared parameters, spatial regularization, MCR-ALS) are out of scope; models the registry can't express fall back to the existing mask/bin coarse-to-fine rules.
- **Tests**: offline `tests/test_parallel_pixel_fit.py` — 42/42: spec normalization (incl. loud unknown-type failure in the parent), serial-vs-parallel **bit-identical** maps, planted-truth accuracy + σ maps, mask scoping, budget partial-return contract (NaN + coverage consistency), dead/NaN/saturated-pixel guards, knob-changes-behavior (bounds clamp; init `auto` 0.61 nm vs `spec` 2.88 nm mean error on a drifting field), called-from-worker-threads with agreeing results, TOOL_SPEC registration. Scale check: 62.5k pixels in 23 s at 2742 fits/s (6.3× the serial rate on 14 workers; a 704×704 commercial frame extrapolates to ~3 min). Live `tests/test_parallel_pixel_fit_live.py` — 5/5, adoption verified from the **saved script**, not the result dict.
- **Known limits**: wave-dispatch checks the deadline between waves, so at most one wave (~`n_jobs × chunk_size` pixels) lands after the budget expires; speedup is sub-linear (spawn warmup + wave barriers + hyperthreading) but the win compounds with cube size; the loky executor persists between calls by design (reused, not leaked).

Closes #356


ziatdinovmax:
Stress-tested with four adversarial live scenarios (all on Bedrock Opus; suite committed as `tests/test_parallel_pixel_fit_stress_live.py`):

| Scenario | What it stresses | Result |
|---|---|---|
| mask_comp | #359 gate verdict → `fit_per_pixel(mask=...)` composition on a 122k-px frame with a ~1000-px blob | 5/5, 10.0 min — script called the tool **mask-scoped**; emission localized, 620 nm recovered |
| multipeak | two-component declarative spec (fixed 520 nm + drifting ~610 nm band, per-pixel ratio, no binning) | 6/6, 5.3 min — both bands recovered through a two-Gaussian spec |
| noise_demand | the dishonesty trap cheap fitting creates: pure noise, per-pixel two-peak maps demanded at native res | honest both rounds — see below |
| fanout_thread | loky called from concurrent fan-out branch threads (the real deadlock hazard), 90k-px cube + curve series | 7/7, 15.6 min — no deadlock, both branches productive, drift + 85 °C transition recovered |

**One real finding, fixed in this branch.** In noise_demand round 1 the codegen's measurability pre-check passed on a statistically flat spectrum (σ estimated from first-differences makes the spectrum's own ripples register as "prominence"), `fit_per_pixel` obligingly returned bound-railing artifact maps, and only the per-map visual review caught it — the run still ended in an honest null, but after rejected retries (7.3 min). Fix: one principle added to the MEASURABILITY GATE — *a statistically flat field-mean spectrum is not_measurable regardless of ratio-based tests*. Round 2 post-patch: the generated code declared the judged honest null on **attempt 1**, 2.0 min. HS goldens regenerated for the prompt change.

**One check corrected, not the agent.** In fanout_thread the cube branch skipped the tool in favor of vectorized argmax + parabolic sub-channel refinement — which is the *right* call for a single dominant peak (the SIZE BUDGET ranks vectorization first; the tool is for unavoidable iterative fits). The test's adoption check was overfit and is now a cost-honesty check: parallel tool OR vectorized estimator, never a hand-rolled per-pixel iterative loop.

PR #369, opened 2026-07-16
Engine-neutral force fields: the OpenFF Interchange path for classical MD

sarah-allec:
Adds an engine-neutral force-field path for classical MD, so a packed box of
arbitrary molecular species can be parameterized and handed to any MD engine
without the pipeline knowing which engine or which FF backend is in play.

## The contract

`ParameterizedSystem` (P1) is the neutral descriptor between the FF producer
and the engine consumer — the same N+M shape as `DeployedPotential`. Adding a
second FF backend is one skill bundle; adding a second engine is one writer.

```
components manifest → ForceFieldAgent.parameterize() → ParameterizedSystem → engine inputs
                        └─ OpenFF branch → build_interchange (skill bundle)
```

- **P1** — `ParameterizedSystem` + Interchange export helper.
- **P3** — the OpenFF skill: `build_interchange` as the engine-neutral FF
  backend, resolved through the skill registry (`TOOL_SPEC`), not imported by
  the agent.
- **P4** — `ForceFieldAgent.parameterize()` OpenFF dispatch (routes by input
  shape, not by engine) + the components manifest and force-field step wired
  into the MD pipeline.

Plus a **parameterization completeness gate**: `create_interchange` silently
leaves unsupported atoms (e.g. a divalent metal Sage has no vdW for) with no
Lennard-Jones term — a zero-repulsion point charge that collapses into its
neighbours. This refuses to emit such a system and names the offending species.
A bundled metal-ion vdW supplement covers the cations Sage lacks.

## Scope note — this bundles more than the FF refactor

Worth flagging for review rather than burying: producing a *runnable* deck from
an Interchange-typed data file surfaced several real defects, fixed here
because the FF path does not work without them:

- `lammps skill`: styles-before-`read_data` for fully-typed FF data files;
  coeffs-in-data-file deck template ordering (the root cause); infer required
  bond/angle/dihedral styles from the data file.
- `md`: orthogonal box + kspace-after-`read_data` for Interchange data files.
- `critics`: RunCritic gets the input files so its fixes patch the real deck;
  RunCritic never returns a vacuous (all-comment) fix.

One commit is arguably its own feature and could be split if preferred: the
**engine-neutral dry-run setup-validation gate** (`31e64950`) — a cheap engine
dry-run that converges the deck to a clean setup before the expensive run.
It is skill-provided (`prepare_dry_run`) and fail-open, so an engine that does
not offer it is unaffected.

## Validation

- Produces **runnable LAMMPS inputs** from the OpenFF path (verified with a
  real `lmp`).
- Carried a full multi-component aqueous Zn(OTf)₂ / sulfone-cosolvent MD
  campaign end to end on HPC (SLURM), including the metal-ion and multi-species
  cases the completeness gate was built for.
- Existing refinement/pipeline test suites updated and green.

## Not included

- **P2** — an AMBER backend (ParmEd-direct; the `from_openmm` bridge is
  fragile).
- **P5** — the full-pipeline benchmark.

Both are additive behind the same `ParameterizedSystem` contract.


ziatdinovmax (approved):
Looks good to me. Claude had several non-blocking comments/findings which I appended below for your Claude to review.

## Findings

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | Minor (latent) | `critics.py:_has_runnable_content` | Comment markers `#`/`!` aren't actually engine-neutral. GROMACS (exporters wired in this PR) uses `;` for comments and `#include`/`#ifdef` as directives — a `.top` fix would be misjudged (real `#include` counted as comment, `;`-comments as runnable). Dormant today (GROMACS not on the executor path); the "engine-neutral" claim overreaches. Scope the comment to wired engines, or handle `;`. |
| 2 | Minor (convention) | `build_interchange` TOOL_SPEC | `nagl_model` is a documented param and in the `signature` string, but omitted from `TOOL_SPEC.parameters`, so the LLM can't set it. CLAUDE.md rule: expose every meaningful tunable. One released NAGL model today, so defensible — either surface it or drop it from the public signature. |
| 3 | Minor (acknowledged) | `refinement.py:_select_deck_fix` | `len(v) >= len(real_deck)` guard rejects a legitimate *shorter* mis-keyed fix (e.g. one that removes an offending line). Docstring owns this; direct-key path is preferred. Failure is silent (`unfixed`) rather than surfaced. |
| 4 | Trivial (perf) | `refinement.py:_stage_dry_dir` | Copies the entire `run_dir` (== `output_dir`) every gate cycle — all loose files (structure, manifest, `system_interchange.json`, logs) into `_dryrun/`, up to 5×. Harmless but real I/O if the Interchange JSON is large. |
| 5 | Cosmetic | `force_field_agent.py:_parameterize_openff` | `backend` provenance string is inconsistent: default records `"openff-2.2.0"` (no `.offxml`) while an override passes through *with* `.offxml`. Provenance-only. |

PR #376, opened 2026-07-18
UI Stop button: actually stop the run, resume cleanly after

ziatdinovmax:
## Summary

Pressing Stop in the UI looked like it stopped the agent, but the run kept going in the background — and if you then continued the session, the next message could fail with a Bedrock error about missing tool results. This PR makes Stop actually stop everything (the orchestrator, delegated specialists, best-of-N candidates, and running fit scripts), makes the session safely resumable afterward, and cleans up how an interrupted delegation is reported in a meta session.

What changes for users:

- **Stop halts the run within seconds**, including scripts already executing and parallel best-of-N candidate attempts.
- **The session continues normally after a Stop** — no more `tool_use ids were found without tool_result blocks` error on the next message. The agent is told the tool call was interrupted so it can re-run it if needed.
- **In a meta session, an interrupted delegation is marked "interrupted"** in the delegation history instead of appearing to run forever, and it is excluded from later cross-dataset fusion.
- Resetting a session from the sidebar now also stops a running agent (previously it kept running in the background).
- Small best-of-N judge wording fix from a live trace: when the judge prefers a candidate that does not have the best numeric metric, it must say so explicitly instead of misstating which R² is highest.

All paths live-validated on Bedrock (details below).

---

## Technical details

**1. Stop signal was swallowed (`1208f312`)**
- `AgentStoppedError` now subclasses `BaseException` (scilink/ui/output_capture.py:8). It was an `Exception`, so any of the ~640 `except Exception` recovery handlers in the agent stack (tool dispatch, retry loops, the orchestrator `chat()` catch-all at analysis_orchestrator.py:1365) caught the stop, fed it back to the LLM as a tool error, and the run continued.
- The UI log-handler filter also raises on stop (scilink/ui/app.py:404) — logging-heavy phases abort on their next log call instead of waiting for a `print()`. Filters run in the emitting thread, and unlike emit errors they are not swallowed by `logging`.
- `kill_subprocesses_for_thread` also kills subprocesses spawned by registered fan-out worker threads (best-of-N candidates), via the `log_context.effective_thread` mapping (scilink/executors.py:39).
- 12 bare `except:` sites (which catch `BaseException`) converted to `except Exception:` so they cannot swallow the stop.
- Sidebar `_reset_session` now signals a running agent to stop (scilink/ui/components/sidebar.py:1159); previously it only unblocked a pending feedback prompt.

**2. Resume after Stop failed with Bedrock BadRequestError (`bec78ac2`)**
- A stop firing inside `execute_tool` leaves the history ending with an assistant `tool_calls` message and no tool results; Anthropic/Bedrock reject such histories on the next turn. History trimming could slice a pair apart the same way.
- `repair_dangling_tool_calls` (scilink/utils/tool_media.py) inserts a synthetic "interrupted — re-run if needed" tool result for every unanswered tool_call id and drops orphan/duplicate tool messages. Called at the top of both chat handlers (OpenAI + LiteLLM) in all four orchestrators — analysis, planning, simulation, meta — covering in-memory continuation, `run_task` delegation, and checkpoint-restored histories. `structure_agent` builds a fresh local message list per call and needs no repair.

**3. Meta ledger entries stuck "running" (`15a5b99e`)**
- `_open_delegation` appends a provisional `running` entry before the child runs; a stop bypasses `_close_delegation`, leaving it `running` forever — misleading the UI delegation tree, `summarize_session_state`, and later `fuse_delegations`.
- `_sweep_interrupted_delegations` (meta_orchestrator.py:1053) runs at every meta chat-turn start: nothing runs between turns, so any `running` entry is dead → closed as `interrupted` with an explicit warning. `interrupted` is a non-success status, so fusion's usable-branch predicate (fanout.py:1068) already excludes it.

**4. Best-of-N judge prompt (`287b29b8`)**
- One-sentence principle added to `BEST_OF_N_JUDGE_PROMPT` (curve_fitting_controllers.py:2790); trace specifics in the commit message.

**Known limit**: an in-flight LLM HTTP call is not interruptible; the abort lands on the next print/log after it returns (~seconds in practice).

**Validation**
- Offline: `tests/test_ui_stop.py` — 12 tests (stop semantics through `except Exception` handlers, logging-filter propagation, no-write-after-stop, worker subprocess kill, dangling-pair repair incl. parallel tool calls and orphan drop, meta sweep idempotency).
- Live on Bedrock (opus-4-8), three scenarios: (a) stop mid-`run_analysis` — thread exited 4 s after stop, zero output growth, no phantom result (5/5 checks); (b) stop → resume — history contained a real dangling `tooluse_…` id, follow-up turn succeeded with a correct session summary (6/6); (c) meta: stop mid curve-fitting delegation → full-stack unwind, ledger `running → interrupted` at next turn, meta correctly reported the interrupted delegation (7/7).

### DIFF
--- a/scilink/agents/exp_agents/analysis_orchestrator.py
+++ b/scilink/agents/exp_agents/analysis_orchestrator.py
@@ -19,7 +19,8 @@ from datetime import datetime
 from enum import Enum
 
 from ...auth import get_internal_proxy_key
-from ...utils.tool_media import (build_tool_message, provider_supports_tool_image,
+from ...utils.tool_media import (repair_dangling_tool_calls,
+                                 build_tool_message, provider_supports_tool_image,
                                  sanitize_history_images)
 from ...wrappers.openai_wrapper import OpenAIAsGenerativeModel
 from ...wrappers.litellm_wrapper import LiteLLMGenerativeModel
@@ -1589,6 +1590,9 @@ class AnalysisOrchestratorAgent:
             timeout=120.0  # 2 minute timeout
         )
         
+        # Repair any tool_use left unanswered by a mid-run user Stop
+        # (or a trim slicing a pair apart) before extending history.
+        self.messages = repair_dangling_tool_calls(self.messages)
         self.messages.append({"role": "user", "content": user_input})
         
         if len(self.messages) > 120:
@@ -1730,6 +1734,9 @@ class AnalysisOrchestratorAgent:
         import litellm
         from ...wrappers.litellm_wrapper import litellm_completion
 
+        # Repair any tool_use left unanswered by a mid-run user Stop
+        # (or a trim slicing a pair apart) before extending history.
+        self.messages = repair_dangling_tool_calls(self.messages)
         self.messages.append({"role": "user", "content": user_input})
         
         if len(self.messages) > 120:
--- a/scilink/agents/exp_agents/controllers/curve_fitting_controllers.py
+++ b/scilink/agents/exp_agents/controllers/curve_fitting_controllers.py
@@ -3314,7 +3314,7 @@ Your guidance: '''
                     # Try tab-delimited
                     try:
                         return np.loadtxt(data_path, delimiter='\t', skiprows=1)
-                    except:
+                    except Exception:
                         raise ValueError(f"Could not parse text file: {data_path}")
         else:
             # Generic attempt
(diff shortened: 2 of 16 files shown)


tkavinchakravarthy-del (approved):
This one is good.

PR #378, opened 2026-07-20
Meta agent: delegate_to_simulation + an executing engine-neutral one-shot

sarah-allec:
Completes the meta agent's third delegation seam: the meta orchestrator can now
delegate computational simulation to the simulation orchestrator, the same way
it already delegates analysis and planning. Builds on #369 and #374 (merged).

## `delegate_to_simulation`

The meta previously declared this as a deferred lazy seam because
`scilink.agents.sim_agents` hard-imports the optional `ase` dependency and the
meta module must stay importable without it. This wires it up while preserving
that guarantee:

- The tool does a **guarded import inside the function** — an `ase`-free build
  returns a clean "install the `[sim]` extra" error instead of failing at
  import. Verified: the meta module still imports with `ase` blocked, and does
  not eagerly load the sim orchestrator.
- The routing inventory now **probes the simulation child** (guarded), so
  `delegate_to_simulation` and its tools appear in the meta LLM's
  SPECIALIST CAPABILITIES block — an `ase`-free build simply omits the section,
  keeping the inventory honest about what the build can do.
- The meta keeps **one persistent simulation child** at `<meta>/simulation/`,
  structure-centric (no `data_dir`), run under the meta's own autonomy mode via
  `run_task` — identical in shape to the analysis and planning children.

## `run_simulation` — an engine-neutral one-shot that *executes*

The simulate orchestrator could generate + validate inputs but had no tool that
ran them, so the meta→sim path stopped at input files. `run_simulation` closes
that: route (scale, engine) from the description, source the run command from
the engine skill's own `default_run_command` (user-supplied overrides), build a
`LocalExecutor` when a command is available and no HPC connection is attached,
and drive `run_complete_workflow` with executor + run command so the simulation
runs and refines. Engine specifics stay in the skill bundle — `default_run_command`
is added to the LAMMPS skill (returns `<lmp> -in {script}`), declared via
`TOOL_SPEC` and resolved through the registry; no engine name or launch command
in shared code.

## Validation

- 3 tests (`tests/test_meta_simulation_delegation.py`): the tool is registered
  with the right schema; a delegation routes to the sim child under the meta's
  autonomy and records a ledger entry; and the meta imports with `ase` blocked
  without loading the sim orchestrator.
- `examples/run_md_oneshot.py` drives the path end to end (directly and through
  the meta in AUTONOMOUS). Live-validated on HPC: from a natural-language goal,
  the meta delegated to simulation, which built a water box, typed it, generated
  a LAMMPS deck, and ran it to a converged density — the full
  goal → structure → FF → deck → run → refine chain, no human in the loop.

## Notes

- v1 scope in CLAUDE.md is updated: simulation delegation is now built, not a
  deferred seam.
- The add/remove of a site-specific `.sbatch` (my hardcoded cluster paths) and
  an accidental run-artifact commit net to nothing in the diff; a `.gitignore`
  covering engine outputs / large binaries / session dirs is included so it
  cannot recur.


ziatdinovmax (changes requested):
All looks good , except for the `.gitignore`, where many valid entries got deleted. I suggest keeping the original standard block and appending the new sim-artifact + session-dir section rather than swapping it out.

sarah-allec:
Good catch — the merge aligned on the shared top-of-file entries and dropped upstream's standard packaging/coverage sections. Rebuilt `.gitignore` from upstream/main as the base with the sim-artifact + session-dir additions appended, so nothing valid is lost (`774809bd`). Verified no upstream entries are missing.

ziatdinovmax (approved):
It maybe a good idea to add one or two tutorial-like examples for sim agents, similar to the analysis agent examples.

PR #379, opened 2026-07-20
Hypothesis generation: best-of-N candidates, lab/ideation modes, and cross-domain cross-pollination (#377)

### LINKED ISSUE #377: Best-of-N hypothesis generation in plan mode: LLM judge selection with human override
## Summary

Plan mode currently generates exactly one experimental plan per `generate_initial_plan` call — but only by prompt mandate ("a list containing exactly ONE experiment plan", `scilink/agents/planning_agents/instruct.py:27`). The pipeline is already plural end to end (`proposed_experiments` is a list; `write_experiments_to_disk`, the orchestrator counter, and the HTML renderer all iterate it), so a best-of-N mode is a prompt + selection change, not a schema change.

Proposal: generate N candidate plans, have an LLM judge pick one, and let the human override the pick in co-pilot/autopilot. Adapts the analysis-side best-of-N pattern (PR #290, shipped in **both** the image and curve-fitting agents) to hypothesis generation.

### Generation — sequential, diversity-conditioned (not independent sampling, not "maximize novelty")

- Independent sampling doesn't work here: hypothesis prose over a fixed retrieved context is mode-seeking, so N independent calls converge on near-identical plans. (Image/curve best-of-N works because stochastic *codegen* genuinely varies across attempts.)
- Blanket "be different from the previous plan" also fails: by candidate 3–4 the model drifts to exotic, weakly-grounded hypotheses just to satisfy the constraint — candidates engineered to lose at the judge, in tension with the grounding safety rule.
- Instead: sequential calls, each seeing prior candidates' one-sentence hypotheses plus one principle-level instruction — test a **different mechanistic hypothesis or strategic approach**; standard practice (equipment, safety, characterization) may repeat; if the provided evidence cannot support another distinct grounded approach, return the error JSON.
- **"Evidence" means the full bundle, not the docs KB alone**: KB retrieval + literature context (`external_context` from `search_literature`/`query_molecules`) + primary data + images + `additional_context` (how database-query and knowledge-tool outputs arrive). A campaign with no docs KB but rich literature context and data can still support several distinct grounded candidates.
- **N is a cap, not a quota** — the decline discipline doubles as the early stop when the evidence only supports k < N distinct approaches.
- **Fallback parity, decided once per run, never mixed**: the single-plan flow doesn't hard-error on insufficient KB — it drops to fallback mode (`HYPOTHESIS_GENERATION_INSTRUCTIONS_FALLBACK`: foundational plan from general knowledge + data + literature, with a forced warning). Best-of-N must not be stricter. The tier is decided once, at candidate 1, by the same sufficiency test: an evidence-sufficient run stays grounded throughout; a fallback run authors ALL candidates under the fallback instructions (distinctness still applies, every candidate carries the warning, judge told all are fallback-tier). No mixed tiers — a general-knowledge candidate next to grounded ones is auto-killed by the judge's groundedness criterion, i.e. a paid-for engineered loser.
- **Retrieve once, author many**: all candidates share one `retrieved_context` (via `perform_science_rag(return_context=True)`), saving N−1 retrievals and making candidates comparable for the judge.
- Clamp N to 1–4; default `n_candidates=1` on `PlanningAgent.generate_plan` and the `generate_initial_plan` tool → byte-identical behavior when unset.

### Judge

New selection prompt mirroring the analysis agents' judges. Sees all candidate plans plus the same evidence the authors saw (retrieved context, primary data, images). Criteria: groundedness, testability, actionability, feasibility, expected information gain. When the pick doesn't top every criterion, the judge states that plainly (house rule from 287b29b8).

### Interaction with the plan critic

Context: live experience showed that an LLM *acting* on the critic's findings reliably **overcorrected** — which is why the critic is advisory-only today (caveats recorded, plan unchanged, human decides). Best-of-N must not reintroduce automatic action-on-criticism through a side door. The three doors, all kept shut:

- **No auto-refine on caveats** — refinement fires only on human free text, as today.
- **No auto-switch on critical findings** — if the critic flags the judge's pick, the caveat is shown and the human override index is the mechanism; the critic never overturns the judge.
- **No criticism in the generation loop** — sequential candidates see prior *hypotheses* only; the critic runs after generation, on the pick, so no candidate is authored in reaction to criticism of a sibling.

Division of labor: judge = comparative selector (no caveats); critic = per-plan adversarial annotator (no ranking, no veto). Only `critic_findings` flow into caveats/warnings — one caveat channel, unchanged.

Sequencing — **critique the pick, lazily**: critic runs on the judge's pick before the human prompt (so caveats inform the decision, as today); on override, the newly chosen candidate gets its own critic pass. Invariant: every plan that becomes `current_plan` was critiqued; nothing else is. O(1) critic calls (2 on override).

Per-candidate critique before judging is deferred (O(N), overlaps judge criteria) — and the overcorrection history is a standing constraint on it: if ever revived, findings reach the judge as information and the human override stays.

### Human review — CLI and UI

Rides the existing `input()` contract (works in CLI, Streamlit UI, and meta AUTOPILOT delegations — the UI already bridges `input()` via `FeedbackRequest` and renders the analysis best-of-N as a radio picker with the judge's pick preselected; the existing parser handles both curve `R²=` and image `score=` formats).

Two prompts, not one:

1. **Selection**: per-candidate **summary cards** — extractive, assembled verbatim from `experiment_name` / `hypothesis` / `expected_outcome` / `justification` + judge scores/reasoning + pick marker with caveat count (no extra LLM call, no summary drift). One stdout block serves both surfaces: CLI reads it directly; a new UI parser (sibling of `_parse_bestofn_review`, keyed on its own header to avoid stale-buffer hijack — the analysis regex's `metric=value` shape doesn't fit judge scores anyway) splits it into markdown cards above a radio. Enter accepts the pick; a digit overrides. Each card links to its full HTML report.
2. **Refinement**: the existing plan-approval prompt on the winner, unchanged — refining a runner-up = override in stage 1, free-text feedback in stage 2.

Per-candidate HTML reports at `plan_candidates/candidate_<i>.html` (reusing `HTMLReportGenerator._render_experiment`), **persisted, not temporary** — the `discard_plan` runner-up fallback, autonomous-run audit, and checkpoint restore all need rejected candidates readable later. Winner still becomes the canonical `plan.html`. UI surfaces them via the existing new-HTML sweep.

### Plumbing

- Candidate table stored as `state["plan_candidates"]` (`{candidates, judge, selected_index, human_override}`), mirrored into results like `anchor_candidates`/`anchor_judge`, and checkpointed.
- Runner-ups are an asset: on `discard_plan`, offer the scored runner-up (with a critic pass at adoption) before regenerating from scratch.
- Downstream contract untouched: selected plan becomes `current_plan`; scalarizer, BO, codegen, `campaign_state`, and the meta `run_task` shape are unchanged.
- Autonomous mode: judge pick final; table + reasoning recorded for audit.

## Deferred

- Per-candidate critic passes before judging (see overcorrection constraint above).
- Adaptive N (escalate only on thin judge margins — planning twin of the image struggle-gate; the curve twin of that gate is itself still a follow-up).
- Best-of-N at the refinement stage (`refine_plan_with_results`).
- Cross-candidate synthesis (judge picks, doesn't merge).
- Foundational-wildcard flag (one clearly-labeled fallback-tier candidate alongside evidence-grounded ones, exempt from the groundedness ranking) — only if users ask for breadth beyond what the evidence supports.

## Open questions

- Should sequential conditioning also see prior candidates' `optimization_params` (search-space diversity), or hypotheses only? Start hypotheses-only.
- Default N per autonomy level: start at 1 everywhere (explicit opt-in) — matches the curve-fitting precedent from PR #290 (curve shipped opt-in at default 1; image defaulted to 3 with calibrated escalation); revisit after live validation.

ziatdinovmax:
Closes #377.

## What this does

When you start a new planning campaign, SciLink no longer commits to the first experimental strategy it thinks of. It now drafts **several genuinely different candidate plans** — each testing a different mechanistic approach to your goal — has an **LLM judge compare them** against the same evidence (your papers, data, images, and lab constraints), and picks the best one. In co-pilot/autopilot you review a compact card for each candidate (hypothesis, expected outcome, justification, judge scores) and can **override the judge with one click/keystroke**; each candidate also gets its own full HTML report saved alongside the plan, so rejected strategies remain readable later.

This happens **by default for a campaign's first plan** (best-of-3) — because even a precisely specified goal usually admits several routes (e.g. precipitation vs. sorption vs. extraction), and the first plan is exactly when no strategy has been committed yet. Follow-up refinements stay single-plan. Asking for "a single plan" or a specific number of candidates always wins.

Honesty is preserved throughout: candidates must be grounded in your evidence, and generation **stops early when the evidence cannot support another distinct approach** — N is a cap, not a quota. A run whose knowledge base lacks method knowledge will say so and produce fewer candidates rather than invent chemistry.

## Two modes: lab and ideation

Planning serves two genuinely different jobs, and one selection policy cannot serve both. A `selection_profile` parameter (default `lab`) switches between them per call:

- **`lab`** — the campaign is headed for actual execution. Candidates must be grounded in the provided evidence, and the judge weights feasibility and actionability most heavily: a plan that cannot run on the available equipment cannot generate information. This is the default and preserves the original behavior.
- **`ideation`** — research ideation. The judge weights information gain and mechanistic novelty first (feasibility becomes a tiebreaker, not a veto), and — the load-bearing half — candidate *authors* receive grounding latitude: where the provided context lacks the decisive mechanism, they may propose one from general scientific knowledge, explicitly marked `(speculative)` in the justification, with fabricated numerics still forbidden. Without this, the evidence-grounding discipline (correct for lab work) structurally suppresses exactly the inspired-leap hypotheses that ideation exists to produce.

Same candidates surface either way, same scoring criteria, same human override — only the authoring latitude and the pick's weighting change. Ideation requires `n_candidates >= 2` — with a single plan there is nothing to select, and the profile is documented as a no-op at N=1. The meta agent switches to ideation when the user is brainstorming rather than planning an executable run.

## Cross-pollination: literature from *other* fields

Scientific literature search previously did one thing: retrieve the problem's
own subfield — established methods, parameter ranges, known pitfalls. That is
what a runnable plan needs, but it also anchors every candidate to what the
field already does.

`search_literature` now offers a second retrieval mode, `cross_domain`, which
asks for **mechanisms from adjacent and unrelated fields that could transfer**
to your problem, and explicitly refuses to produce a topical review. Asked
about flexible heat-harvesting devices, for instance, it returns things like
counter-current heat exchange in fish vasculature and island–bridge
architectures from stretchable electronics — analogies whose *mechanism* might
carry over, which is where genuinely new ideas usually come from.

**Several search types can now be requested at once and run in parallel**
(`search_type="hypothesis_context,cross_domain"`), merging into one document
with labelled sections — "ESTABLISHED IN THIS FIELD" versus "TRANSFERABLE
MECHANISMS FROM OTHER DOMAINS (analogies, NOT established results)". The
labelling matters: it lets candidates use field literature as constraints and
cross-domain material as mechanism inspiration, instead of treating everything
as equally established. Because the searches run concurrently, the pair costs
roughly the wait of a single search.

**It is deliberately scoped rather than on by default**, and paired with a
fix for something literature does to constrained plans. Reading literature —
any literature, not just cross-domain — makes a plan more likely to *omit* a
stated requirement: it stays on-topic and scientifically stronger, but
quietly drops the less glamorous constraints, like a durability check or a
secondary characterisation. The plans do not violate constraints; they just
fail to mention them.

Withholding literature would fix that by giving up its benefit, so instead,
whenever your constraints and literature are both in play, the planner is
required to tie every constraint to a named step that says which requirement
it satisfies. In testing this recovered nearly all of the lost coverage with
no measurable cost to plan quality. So cross-domain retrieval is scoped to
ideation because that is what it is *for* — not because it is unsafe under
constraints.

Two adjacent fixes that surfaced during live testing are included:
- **Plan caveats now always describe the current plan.** Previously, when the orchestrator fixed an issue the critic had raised (e.g. adding the ICP-MS dilution step the critic asked for), the report still showed the old caveat. Revisions are now re-checked so resolved caveats disappear, unresolved ones stay, and new ones are flagged. The critic remains strictly advisory — it annotates, never acts.
- **UI first-render fix**: the mission-control landing page now shows the embedding model/API-key fields and the correct autonomy options immediately, instead of only after the first interaction.

## How to use it

- **Default**: start a campaign (UI, CLI, or via the meta agent) — the first plan generates as best-of-3 automatically. Review the cards, accept the judge's pick (Enter / green button) or pick another candidate (number / radio), then the familiar plan review follows.
- **Control the width**: say "generate 4 candidate plans" or "just one plan"; the meta forwards this.
- **After selection**: everything downstream (implementation code, scalarizer, BO) is unchanged; the runner-up candidates stay available under `plan_candidates/` and are offered if you later discard the plan.

---

<details>

### Generation (`planning_rag.py:generate_plan_candidates`)
Sequential, diversity-conditioned: candidate 1 runs the normal science RAG; candidates k>1 see prior hypothesis sentences plus `HYPOTHESIS_DISTINCTNESS_CONDITIONING` and may decline via the existing error-JSON discipline (early stop). Every author — candidate 1 included — receives `HYPOTHESIS_BESTOFN_AUTHOR_NOTE` ("you are writing ONE of N; alternatives are separate candidates"): without it, an objective phrased "explore several alternatives" (user phrasing travels verbatim through the meta) made the unconditioned first author pack all strategies into one plan, which then won at the judge on scope-matching (observed live twice). The instruction tier (strict vs fallback) is detected at candidate 1 via an opt-in `mode_key` stamp on the shared `run_rag` (`rag_engine.py`; existing callers byte-identical) and pinned for the whole run — no mixed-tier candidate sets. All candidates share one retrieval (deterministic), making them comparable for the judge.

### Judge (`planning_rag.py:judge_plan_candidates`)
Comparative selector, not an editor: scores groundedness / testability / actionability / feasibility / information-gain against the same evidence the authors saw — including `additional_context`, where lab constraints (e.g. "Opentrons Flex 2, 96-well plates") travel; feasibility is scored against them. Must state plainly when its pick lacks the top score on some criterion. Fails open to candidate 1. Emits no caveats — the advisory critic remains the single caveat channel.

### Selection UX
`display_plan_candidates` prints ONE stdout block of extractive cards (verbatim plan fields + judge scores — no summarization LLM call, no drift) serving both surfaces: CLI reads it directly; the UI's `_parse_plan_candidate_review` (sibling of the analysis best-of-N parser, gated on its own `accept plan candidate` / `PLAN CANDIDATES` markers to prevent cross-hijack) renders a radio with the judge's pick preselected. Two prompts: stage-1 selection (Enter/index), then the existing stage-2 plan approval on the winner — refining a runner-up = override, then normal feedback. Composes with meta AUTOPILOT delegations through the existing `input()` bridge. Per-candidate HTML reports persist under `plan_candidates/` (`html_generator.py:generate_single_plan`) for runner-up fallback, audit, and checkpoint restore.

### Critic interaction (advisory-only preserved)
The critic runs lazily on whichever candidate is selected (and again on an override) — every plan that becomes `current_plan` is conformance-checked and critiqued; nothing else is (O(1), 2 on override). Critical findings never auto-switch the selection; caveats are displayed at the selection prompt for the human to weigh. Sequential candidates see prior hypotheses only — criticism never enters the generation loop.

**Revision bookkeeping fix** (`planning_agent.py:_recritique_revision`): the refinement LLM echoes prior `critic_findings` into revised JSON; `refine_plan` and `adjust_plan_for_constraints` (and their human-feedback branches) now pop the echo and re-critique in resolution-check mode (drop resolved / keep unresolved / flag new), exactly like `generate_plan`'s human path. Annotation-only; pre-revision history snapshots keep their original findings. Live-validated: a planted TDS-dilution caveat was dropped after the adjustment that resolved it.

### Default policy (`orchestrator_tools.py:resolve_n_candidates`)
Omitted `n_candidates` → best-of-3 for a campaign's FIRST plan, 1 afterwards; explicit values (incl. 1) win, clamped 1–4. Rationale: objective *specificity* is the wrong trigger (upstream routing always narrows objectives before delegating, and specific goals still admit multiple strategies); the meaningful single-plan case is a *committed* strategy. The programmatic `PlanningAgent.generate_plan` API keeps its explicit default of 1 — policy lives at the LLM-tool boundary only. The meta's `delegate_to_planning` description advertises the capability; both descriptions instruct keeping the scientific objective singular (the parameter provides the multiplicity).

### Selection profiles (`selection_profile`, default `lab`)
`BESTOFN_SELECTION_PROFILE_LAB` / `_IDEATION` append the pick-weighting to the judge prompt; `IDEATION_AUTHOR_OVERRIDE` rides the generation instructions adjacent to the safety rule it supersedes, granting authoring latitude in ideation only (lab authors byte-identically). Profile recorded in `state["plan_candidates"]`; exposed on `generate_initial_plan` (enum) with the meta instructed when to switch.

### Cross-domain retrieval and parallel search
`LiteratureSearchAgent.search_for_cross_domain` asks for transferable
mechanisms from adjacent/unrelated domains and forbids a topical subfield
review. It bypasses `search_for_hypothesis_context`'s wrapper rather than
reusing it — that template ("a comprehensive review of experimental
methods… common pitfalls") reframes the task and neutralises cross-domain
intent, observed live before the fix.

`search_literature` accepts comma-separated types, runs them in a
`ThreadPoolExecutor`, and merges into one labelled document; partial failure
keeps the successful half and reports the rest in `failed_searches`; any
successful search carries the constraint-coverage caveat in its payload.
Single-type behaviour is byte-identical. The orchestrator's
`LITERATURE SEARCH WORKFLOW` block gained the pairing rule (cross-domain +
ideation profile), since the tool-parameter description alone was
discoverable only to a model reading parameter docs.

**Constraint coverage** (`instruct.CONSTRAINT_COVERAGE_NOTE`, spliced in
`planning_rag.perform_science_rag`): when `additional_context` and
`external_context` are both present, the author is required to map every
hard constraint to a named step. Injected at the point where the single-plan
and best-of-N paths converge, so every candidate carries it, and scoped to
that condition — a constraints-only run already complies, so the note stays
out of the prompt otherwise. The earlier cross-domain-specific caveat was
misattributed: grounding-only retrieval costs the same coverage, so the
caveat is now literature-general and names the remedy (map constraints to
steps) instead of advising against a search type.

Measured while validating: 4/4 concurrent Edison searches succeed with a
2.6x speedup over serial, and 7+ were observed in flight account-wide, so the
2-way pairing sits well inside capacity and no executor cap was added. Note
the concurrency pool is account-wide — a future fan-out with per-branch
literature search would want a shared semaphore.

### Validation
- **Offline**: 18 tests in `tests/test_best_of_n_planning.py` (sequential conditioning, shared evidence, decline early-stop, fallback pinning, judge fail-open, override + lazy-critique invariant, constraints-to-judge, author note, default policy, CLI→UI parser round-trips, constraint-coverage injection scoping) + 5 in `tests/test_revision_recritique.py`; existing planning suites unchanged.
- **Live orchestrator validation of the scoping** (autonomous session, two turns, verified from recorded tool arguments rather than reply text): an ideation ask produced `search_type='hypothesis_context,cross_domain'` + `selection_profile='ideation'` with both labelled sections present in the merged file; a constraint-bound ask (Opentrons Flex 2, 96-well, UV-Vis) produced field-only retrieval, with the constraints travelling as `additional_context`.
- **Live (Bedrock Opus 4.8 + Gemini embeddings)**: produced-water KB runs (fallback tier, 3 distinct mechanisms, judge tradeoff stated); Opentrons-constrained runs (diversity relocates from mechanism family to strategy lever; judge cites deck-feasibility); FutureHouse literature run (**tier flips to strict** with lit context vs fallback KB-only on the identical KB — the evidence-bundle decline design working; real `source_documents`); two full Playwright-driven Streamlit sessions (override path and accept path, goal form → cards+radio → approval → final summary narrating the override); meta-delegation runs (pass-through, strategy-committed collapse via an articulate decline, default firing unprompted); cross-domain check on solid polymer electrolytes (3 canonical distinct routes — no overfit to the produced-water example).

### Known limits
- Candidate evidence for the post-selection re-critique is the revision context (prior plan + findings + request); the original retrieval context is not persisted across calls.
- In strict tier with a KB that grounds the feed but not alternative methods, later candidates may decline and the set legitimately collapses (honest-decline discipline). A labeled general-knowledge "wildcard" candidate alongside grounded ones is a deferred opt-in (see design note).
- Deferred: per-candidate critique before judging, adaptive N on thin judge margins, best-of-N at the refinement stage, cross-candidate synthesis.

</details>

### DIFF
--- a/scilink/agents/lit_agents/literature_agent.py
+++ b/scilink/agents/lit_agents/literature_agent.py
@@ -373,6 +373,36 @@ class LiteratureSearchAgent:
         )
         return self._execute_crow_task(formatted_query, task_type="Hypothesis")
 
+    def search_for_cross_domain(self, objective: str) -> Dict[str, Any]:
+        """
+        Formats a query for INSPIRATION retrieval: mechanisms and design
+        principles from adjacent or unrelated fields that could TRANSFER to
+        this problem.
+
+        Deliberately not a topical review. ``search_for_hypothesis_context``
+        retrieves the problem's own subfield — which grounds a plan but also
+        anchors it to established approaches. This one asks for analogies
+        whose mechanism might carry over, which is where a genuinely new
+        mechanistic idea usually comes from.
+
+        Intended for ideation. Benchmarking showed cross-domain context
+        raises idea novelty and non-obviousness, but degrades adherence to
+        hard equipment/process constraints — so pair it with grounding
+        retrieval, and prefer grounding alone when a plan must satisfy
+        stated constraints.
+        """
+        formatted_query = (
+            "Survey mechanisms and design principles from ADJACENT and "
+            "UNRELATED domains that could TRANSFER to the problem below. Do "
+            "NOT provide a topical review of the problem's own subfield "
+            "methods. For each analogous system (a different chemistry, "
+            "biology, or engineering field achieving a similar function), "
+            "describe the underlying mechanism and explain why it might "
+            "transfer. Emphasize unconventional and emerging approaches.\n\n"
+            f"PROBLEM: {objective}"
+        )
+        return self._execute_crow_task(formatted_query, task_type="CrossDomain")
+
     def search_for_fitting_models(self, objective: str) -> Dict[str, Any]:
         """
         Formats a query specifically for finding mathematical equations.
--- a/scilink/agents/meta_agent/meta_orchestrator_tools.py
+++ b/scilink/agents/meta_agent/meta_orchestrator_tools.py
@@ -397,7 +397,15 @@ class MetaOrchestratorTools:
                 "with no interactive user and returns a structured JSON result "
                 "(status, summary, key_findings, files_produced, "
                 "suggested_followups, warnings, delegation_index). `task` must "
-                "be a complete, self-contained instruction."
+                "be a complete, self-contained instruction. The specialist "
+                "supports best-of-N plan generation (an LLM judge picks among "
+                "distinct candidate strategies): when the user asks for "
+                "alternatives or the design goal is open-ended, add the "
+                "instruction 'use n_candidates=N on generate_initial_plan' "
+                "(N up to 4) to the task — but keep the scientific goal "
+                "itself SINGULAR (never phrase the objective as 'propose N "
+                "strategies'; the tool parameter provides the multiplicity). "
+                "Omit for narrow or follow-up tasks."
             ),
             parameters={
                 "task": {
--- a/scilink/agents/meta_agent/meta_orchestrator_tools.py
+++ b/scilink/agents/meta_agent/meta_orchestrator_tools.py
@@ -405,7 +405,10 @@ class MetaOrchestratorTools:
                 "when the user wants a specific width — including N=1 for a "
                 "single plan. Keep the scientific goal itself SINGULAR "
                 "(never phrase the objective as 'propose N strategies'; the "
-                "tool parameter provides the multiplicity)."
+                "tool parameter provides the multiplicity). When the user is "
+                "brainstorming/ideating rather than planning an executable "
+                "run, add 'use selection_profile=discovery' to the task so "
+                "the candidate judge weights novelty over feasibility."
             ),
             parameters={
                 "task": {
--- a/scilink/agents/meta_agent/meta_orchestrator_tools.py
+++ b/scilink/agents/meta_agent/meta_orchestrator_tools.py
@@ -399,13 +399,13 @@ class MetaOrchestratorTools:
                 "suggested_followups, warnings, delegation_index). `task` must "
                 "be a complete, self-contained instruction. The specialist "
                 "supports best-of-N plan generation (an LLM judge picks among "
-                "distinct candidate strategies): when the user asks for "
-                "alternatives or the design goal is open-ended, add the "
-                "instruction 'use n_candidates=N on generate_initial_plan' "
-                "(N up to 4) to the task — but keep the scientific goal "
-                "itself SINGULAR (never phrase the objective as 'propose N "
-                "strategies'; the tool parameter provides the multiplicity). "
-                "Omit for narrow or follow-up tasks."
+                "distinct candidate strategies) and DEFAULTS to best-of-3 for "
+                "a campaign's first plan. Only add an explicit "
+                "'use n_candidates=N on generate_initial_plan' instruction "
+                "when the user wants a specific width — including N=1 for a "
+                "single plan. Keep the scientific goal itself SINGULAR "
+                "(never phrase the objective as 'propose N strategies'; the "
+                "tool parameter provides the multiplicity)."
             ),
             parameters={
                 "task": {
--- a/scilink/agents/meta_agent/meta_orchestrator_tools.py
+++ b/scilink/agents/meta_agent/meta_orchestrator_tools.py
@@ -407,7 +407,7 @@ class MetaOrchestratorTools:
                 "(never phrase the objective as 'propose N strategies'; the "
                 "tool parameter provides the multiplicity). When the user is "
                 "brainstorming/ideating rather than planning an executable "
-                "run, add 'use selection_profile=discovery' to the task so "
+                "run, add 'use selection_profile=ideation' to the task so "
                 "the candidate judge weights novelty over feasibility."
             ),
             parameters={
--- a/scilink/agents/planning_agents/html_generator.py
+++ b/scilink/agents/planning_agents/html_generator.py
@@ -247,6 +247,32 @@ class HTMLReportGenerator:
         </div>
         """
 
+    def generate_single_plan(self, plan: Dict[str, Any], output_path: str,
+                             title: str = "Candidate Plan"):
+        """Render ONE plan dict (not the session history) to a standalone HTML
+        report — used for best-of-N candidate reports, where each candidate
+        gets its own persisted page under ``plan_candidates/``."""
+        content_html = "".join(
+            self._render_experiment(exp, x + 1)
+            for x, exp in enumerate(plan.get('proposed_experiments', []))
+        )
+        html_content = f"""
+        <!DOCTYPE html><html><head><meta charset="UTF-8"><title>{html.escape(title)}</title>{self._get_css()}</head>
+        <body><div class="container">
+            <header>
+                <h1>🔭 {html.escape(title)}</h1>
+                <div class="objective"><strong>Objective:</strong> {html.escape(self.state.get('objective', '') or '')}</div>
+            </header>
+            <div class="step">
+                <div class="step-header"><span class="badge plan">EXPERIMENTAL STRATEGY</span></div>
+                {content_html}
+            </div>
+            <footer>{self.generated_by}</footer>
+        </div></body></html>
+        """
+        with open(output_path, 'w', encoding='utf-8') as f:
+            f.write(html_content)
+
     def generate(self, output_path: str):
         date_str = self.state.get('start_time', datetime.now().isoformat())
         
--- a/scilink/agents/planning_agents/instruct.py
+++ b/scilink/agents/planning_agents/instruct.py
@@ -185,8 +185,8 @@ Respond with a single JSON object:
 """
 
 
-DISCOVERY_AUTHOR_OVERRIDE = """
-## DISCOVERY MODE — GROUNDING LATITUDE (overrides the Safety Rule above)
+IDEATION_AUTHOR_OVERRIDE = """
+## IDEATION MODE — GROUNDING LATITUDE (overrides the Safety Rule above)
 This run is research ideation. The provided context may deliberately lack
 the key inspiration. Instead of declining for insufficient context: ground
 your plan in the context where possible, and where the decisive mechanism
@@ -207,8 +207,8 @@ Prefer the candidate most likely to produce a clean, interpretable result
 on the stated platform."""
 
 
-BESTOFN_SELECTION_PROFILE_DISCOVERY = """
-## SELECTION WEIGHTING — DISCOVERY PROFILE
+BESTOFN_SELECTION_PROFILE_IDEATION = """
+## SELECTION WEIGHTING — IDEATION PROFILE
 This is research ideation: the goal is the most scientifically valuable
 hypothesis, not tomorrow's protocol. Weight information_gain and
 mechanistic novelty most heavily when selecting: prefer the candidate
@@ -957,7 +957,7 @@ SAMPLE FILENAMES (first 20): {filenames}
 
 QUESTION: {query}
 
-The script below provides imports, file discovery, and reader functions.
+The script below provides imports, file ideation, and reader functions.
 Complete ONLY the TODO section (query logic).
 
 ```
@@ -991,7 +991,7 @@ SAMPLE FILENAMES (first 20): {filenames}
 QUERY: {query}
 TOP-K to retain: {top_k}
 
-The script below provides imports, file discovery, and reader functions.
+The script below provides imports, file ideation, and reader functions.
 Complete ONLY the TODO section.
 
 ```
--- a/scilink/agents/planning_agents/instruct.py
+++ b/scilink/agents/planning_agents/instruct.py
@@ -133,6 +133,16 @@ You MUST respond with a single JSON object containing a key "proposed_experiment
 """
 
 
+HYPOTHESIS_BESTOFN_AUTHOR_NOTE = """## ⚠️ BEST-OF-N AUTHORING NOTE
+You are authoring ONE candidate in a multi-candidate (best-of-N) process:
+alternative strategies are handled by SEPARATE candidate plans and compared
+by a judge afterwards. If the objective asks to explore multiple distinct
+strategies or alternatives, that request is satisfied by the process — NOT by
+this plan. Propose exactly ONE strategy: a single coherent experiment plan
+built around one mechanistic approach. Do NOT pack multiple alternative
+strategies into this plan."""
+
+
 HYPOTHESIS_DISTINCTNESS_CONDITIONING = """## ⚠️ PRIOR CANDIDATE HYPOTHESES (already proposed for this objective)
 {prior_hypotheses}
 
--- a/scilink/agents/planning_agents/instruct.py
+++ b/scilink/agents/planning_agents/instruct.py
@@ -133,6 +133,48 @@ You MUST respond with a single JSON object containing a key "proposed_experiment
 """
 
 
+HYPOTHESIS_DISTINCTNESS_CONDITIONING = """## ⚠️ PRIOR CANDIDATE HYPOTHESES (already proposed for this objective)
+{prior_hypotheses}
+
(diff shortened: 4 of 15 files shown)


sarah-allec (changes requested):
Read through this properly — it's a well-built PR. The tier pinning so a fallback run authors *all* candidates under fallback instructions, the decline-as-early-stop so N is a cap rather than a quota, the judge failing open to candidate 1, and the "lazy-critique invariant" re-running conformance+critic after a human override are all the right calls, and the 28 tests across three files cover the paths I'd have worried about (override, decline, fallback pinning, invalid judge selection, partial literature failure). The author-note-on-candidate-1 fix with the comment explaining the observed failure — the unconditioned first author packing all alternatives into one plan and then winning on scope-matching — is a nice catch.

One thing needs fixing before merge. A find-replace of "discovery" → "ideation" has corrupted two unrelated prompts in `instruct.py`: both agentic data-query codegen prompts now read *"The script below provides imports, file ideation, and reader functions"* (lines ~875 and ~909 on main). That's the structured-data query tool, nothing to do with this feature, and "file ideation" is meaningless — it'll just degrade that prompt. Should be `file discovery` in both places.

A couple of smaller things, take or leave:

In `generate_plan_candidates`, the ideation author override is dropped whenever the run falls back: candidate 1 swaps to `HYPOTHESIS_GENERATION_INSTRUCTIONS_FALLBACK` inside `run_rag`, and candidates 2+ then use the fallback instructions directly, so `IDEATION_AUTHOR_OVERRIDE` never applies in a fallback run. That's probably fine in substance — fallback already licenses general knowledge — but it means `selection_profile="ideation"` silently has author-side effect only in strict runs, with just the judge weighting applying otherwise. Worth a line in the docstring so it isn't surprising later.

The ideation override is the part I looked hardest at, since it deliberately relaxes the derivability rule. The guardrails look right to me — opt-in only, profile clamped to `("lab","ideation")` at the tool layer, `(speculative)` marking required, numeric fabrication still forbidden, and the default stays `lab`. My only reservation is that "choose the boldest hypothesis you can rigorously defend" is doing a lot of work next to a rule it's overriding; if ideation runs start producing plans whose speculative parts aren't marked, that phrasing is where I'd look first.

Last one: `resolve_n_candidates` keys "first plan" off `not planner_state.get("current_plan")`. That's the right signal, but it means a restored session with a checkpointed `current_plan` correctly gets 1 while a fresh session that failed its first plan would get 3 again on retry — which I think is what you want, just flagging that the retry-after-failure case re-pays for 3 candidates.

Happy to approve once the `file discovery` revert is in.

sarah-allec (approved):
Re-reviewed — the `discovery`→`ideation` collateral is reverted (the two data-query prompts read `file discovery` again), which was my only blocking item, and I see you'd already caught it. Approving.

The three earlier notes stay non-blocking, just flagging for later: the ideation author override is inert in a fallback run (only the judge weighting applies then — worth a docstring line), "choose the boldest hypothesis you can rigorously defend" is the phrasing I'd watch if speculative parts start going unmarked, and `resolve_n_candidates` re-pays for 3 candidates on a first-plan retry-after-failure (probably intended). Nice PR.

PR #394, opened 2026-07-24
Plan mode separates experiments, ideation portfolios and documents (+ named KB store, campaign-scoped literature, UX fixes)

ziatdinovmax:
Plan mode grew three things that were all being squeezed through one tool, and this PR separates them — alongside a named knowledge-base store, campaign-scoped literature, and a run of UX and robustness fixes found by using the thing.

## What changed, in plain terms

**Ideation is no longer an experimental plan.** `generate_initial_plan` designs bench experiments. Asking it for a *portfolio of research directions* meant the portfolio got flattened into a numbered protocol — in one real session, 56 "experimental steps" that were actually document sections, printed at a review prompt whose only question was "is this the right direction?". There is now `generate_ideation_portfolio`, which produces directions (each with its own hypothesis, rationale and novelty) under an organizing thesis, plus `refine_portfolio` to harden, drop, add, re-rank or consolidate them.

**Documents are no longer experimental plans either.** "Plan how we build the facility", "estimate the floor space we need", "merge these two documents" are requests a planning agent should answer — but they are not experiments, and forcing them through the experiment schema made the model invent fields to fill it: a build sequence as a `hypothesis`, notes-to-self as protocol steps, and thirteen fabricated optimization parameters carrying numeric ranges and citations for a facility that did not exist. `write_technical_document` authors those properly, grounded in the campaign's literature and in prior session documents.

**Knowledge bases are named, reusable artifacts.** `scilink kb create/add/list` builds a store under `~/.scilink/knowledge_bases/`, so a corpus survives the session that built it. Retrieval degrades rather than dies: if the embedding provider is unavailable a BM25 keyword tier keeps the KB usable, and a total failure falls through to no-context with a warning instead of killing generation.

**Literature searches stopped leaking across topics, and stopped hanging.** Two unrelated brainstorms in one session used to share a corpus (#396); literature is now campaign-scoped. Separately, a job that failed on the provider's end could pin an entire batch — one search held five hostage for 45 minutes — because only two spellings of "dead" counted as terminal and nothing bounded the batch.

**Ideation runs produce the artifacts people actually asked for**: a sponsor-facing white paper (now with a PDF twin), an all-candidates dossier where runner-up directions are deliverables rather than rejects, and a browser-readable portfolio report. The protocol-shaped `plan.html` no longer appears for ideation work, because it misrepresents a portfolio.

Plus UI work: the File Explorer renders markdown instead of dumping its source, PDFs preview, and light-mode code blocks stopped hiding half their text.

---

### Plan mode's three contracts

| ask | tool | payload |
|---|---|---|
| design an experiment | `generate_initial_plan` | `proposed_experiments` |
| which directions are worth pursuing | `generate_ideation_portfolio` | `directions` |
| roadmap / estimate / memo | `write_technical_document` | markdown sections, no campaign state |

Decision rule, now recorded in `CLAUDE.md`: **no hypothesis to test and nothing to measure means it is not an experimental plan.**

**One engine, two contracts.** `generate_plan(kind="portfolio")` reuses retrieval, best-of-N, the judge, the critic, campaign scoping, checkpointing and the deliverables ledger. `generate_plan_candidates` takes a `contract` that swaps only what a candidate *is*; absent one it is the experiment path unchanged — tier pinning, distinctness conditioning and early stop are shape-agnostic and stayed shared.

**Accessors, not shape checks.** `parser_utils.plan_directions` / `plan_is_portfolio` / `plan_thesis` resolve top-level `directions` → `proposed_experiments[*].concepts` → a direction synthesised from experiment fields. That ladder is what keeps pre-existing checkpoints restorable.

**Transition shim.** A portfolio carries both shapes: `directions` as the payload plus a one-entry `proposed_experiments` shim, because ~50 legacy readers treat a missing key as a *failed* plan. Consequence worth knowing: every pass that re-emits a plan (conformance, critic, refinement) edits the shim, so `resync_portfolio` — called from `_stamp_campaign` — makes the nested copy authoritative. Without it a refined portfolio silently served stale directions. `test_no_new_direct_readers_of_the_experiment_schema` ratchets the reader count at 71 so the debt cannot grow; removing the shim is deliberately deferred.

**Report emission** is one helper (`_emit_plan_report`) instead of five inlined copies. Suppression previously lived on the initial-plan path only, while four refinement paths re-rendered the protocol view regardless.

### Literature

- Campaign-scoped registry keyed on `campaign_id`, with `objectives_share_campaign` deciding continuation vs new campaign. Threshold 0.35, set from measured pairs: unrelated ≤0.20, genuine continuations ≥0.43.
- Terminal remote states matched by substring (`fail|error|cancel|crash|timeout|abort|kill|reject`) rather than a two-item list; batch deadline 1800 s with per-future timeouts returning partial results; per-task budget 1500 s.
- Per-type objectives: `hypothesis_context` and `cross_domain` get different questions — `cross_domain` needs a *function to transfer toward*, not a field to survey.
- The internal fallback that searched whenever a key was present is removed; `search_literature` is the only path.

### Knowledge bases

`scilink/knowledge/kb_store.py`, with `_swap_into_place` publishing via rename-aside so a KB is never absent from both paths (the previous rmtree-then-rename left a window where a crash lost it, and `kb list` hid the orphan). BM25 tier in `sparse_retrieval.py`.

### Known limits

- **Routing is an LLM judgment, not a guarantee.** ~16 routing decisions against the final code with no misroutes — but that is consistent with a real misroute rate up to ~19% (rule of three), the samples are correlated, and two of the three defects found in overnight testing were routing-shaped. What does not depend on routing is the containment: the anti-fabrication rule now rides every plan tier's prompt, so even a misrouted document cannot invent optimization parameters.
- The meta sometimes asks a clarifying question instead of delegating. Correct with a human present; it stalls scripted runs.
- `markdown-it-py` is now a direct dependency (it already arrived transitively via streamlit→rich). PyMuPDF's `Story` renders a subset of HTML/CSS, so it backs documents, not the richer `plan.html` reports.

### Validation

Full suite **1,464 passing**, failure set byte-identical to the pre-branch baseline at every phase boundary (the 26 failures / 24 errors are pre-existing env-dependent sim and orchestrator tests).

Live, on Bedrock Opus 4.8: the nine-turn session that motivated the ideation work was replayed end to end — **92 min, 9/9 turns, zero invariant violations**. Against that original session: fabricated optimization parameters **13 → 0**, `plan.html` for ideation work **2 delegations → none**, console protocol dump **~200 lines → 0**. Routing across the replay was 5 × portfolio, 6 × document, 0 × plan tool, which is correct because that session never asked for a bench experiment.

Separately validated live: the lab path with a refinement cycle (14 steps, 8 equipment items, critic firing, report starred as the deliverable); the escape hatch (ideate → pick a direction → `type=lab` bench plan with 11 steps); portfolio refinement verbs (harden left untouched directions byte-identical, drop removed one and kept the rest verbatim); and portfolio survival across a meta checkpoint restore, including a session whose history holds both a portfolio and an experiment.


sarah-allec (changes requested):
Big PR, so I focused on the two spots where a bug actually costs something — the KB store's safety guarantees and the campaign scoping that keeps one brainstorm's literature out of another. Both are in good shape. The no-write-back (session `kb_cache` copy) and no-re-ingest (built KBs resolve to `sources/`, imported to nothing) guards are real and tested, and recording each KB's embedding model so a mismatch warns instead of failing opaquely is a nice touch.

The one thing I'd like fixed before merge is the "atomic swap." `create`/`import`/`add`/`rebuild` all end with `shutil.rmtree(final)` then `staging.rename(final)`. The rename is atomic, but the rmtree-then-rename pair isn't — if the process dies between them, the live KB is gone and the rebuilt one is stranded at `.staging_<name>`. In principle it's recoverable, but in practice the user isn't told: `kb list` hides dot-dirs, nothing promotes the staging copy on the next run, and a natural "my KB vanished, let me recreate it" reaction runs `kb create`, which wipes the stale staging dir first — destroying the one surviving copy. So the realistic outcome is a KB that silently disappears. The usual fix closes it cleanly: rename the old one aside (final → final.bak), `staging.rename(final)`, then drop the backup — the KB is never absent from its path. The tests cover the success path but not a crash mid-swap.

On campaign scoping, the logic is solid and I don't think anything's wrong — this is more a question about which way it leans. `objectives_share_campaign` calls it the same campaign when content-word overlap over the smaller set is ≥0.25, and ties toward continuity. The two mistakes aren't symmetric, though: wrongly saying "same campaign" carries the previous corpus forward (the #396 leak), while wrongly saying "new campaign" only loses carry-forward. With a short new objective, one shared specific-ish word ("solar" in *perovskite solar cell* vs *solar wind radiation*) can clear 0.25 and read as a continuation. I know biasing the other way has its own cost — a reworded objective wrongly splitting a live campaign — and this only fires when the LLM didn't already flag the change, so it's a backstop. Just flagging that the backstop leans toward the more expensive mistake; worth a moment's thought on the threshold, not a change I'd insist on.

Small thing in that same function: `w.rstrip("s")` strips every trailing s, not a plural — "stress" → "stre", "gas" → "ga". It's applied to both sides so matching still works, but it's a sloppy fold that could cause odd collisions; a single-s strip or a real stemmer would be cleaner. Take or leave.

Nice work on the safety plumbing overall — the guards and the provenance recording are the right instincts. Just the swap-ordering I'd want before merge; the rest is yours to weigh.

ziatdinovmax:
Thanks — the swap-ordering catch is a good one, and your framing of *why* it matters (invisible staging dir → "recreate it" → `kb create` clears staging → last copy gone) is what makes it worth fixing rather than documenting. Fixed in f2682f0e.

**Swap ordering.** All four sites (`create`/`import`/`add`/`rebuild`) now go through one `_swap_into_place()` using exactly the ordering you suggested: rename the live directory aside to `<name>.bak`, rename staging into place, drop the backup only after that succeeds. `final` is at every instant either the old KB or the new one. Two additions on top:

- if the publish rename fails, the backup is restored; if that restore *also* fails, the original error still propagates while the log names where the surviving copy is — a cleanup failure should never mask the real one.
- `kb list` now warns about leftover `.staging_<name>` directories instead of hiding them, since hiding them is what turned a crash into a silent disappearance.

Tests cover crash-mid-publish (asserting the original KB is still readable), the success path, no-previous-KB, a stale `.bak`, and the leftover report.

**Threshold.** Taken, and your asymmetry argument is the right lens — a false "same campaign" carries the corpus forward (the #396 leak), a false split only loses carry-forward and the caller can override it with `new_campaign=False`. I measured it rather than guessing, over real objectives from live sessions:

| | overlap |
|---|---|
| unrelated pairs (incl. your perovskite-solar vs solar-wind) | ≤ 0.20 |
| genuine continuations | ≥ 0.43 |

Your example lands at exactly 0.200, so it didn't trip 0.25 — but that left only 0.05 of margin on the dangerous side. Moved to 0.35, which sits mid-gap and leans the residual risk toward the cheaper mistake. A test locks both sides against those real pairs.

**Plural fold.** Taken, and it was worse than sloppy: `rstrip("s")` strips *every* trailing s, so `gas`→`ga` but `gases`→`gase` — it fails at the job it was doing, which can cost a genuine match. Now a single trailing `s`, skipped for `ss`/`us`/`is` endings and short words. Still a naive fold rather than a stemmer (gas/gases stays unmatched), documented as such.

Full suite: 1,318 passing, failure set unchanged from the pre-branch baseline.

ziatdinovmax:
Founds a few minor bugs, converting this PR to draft while fixing them...

ziatdinovmax:
@sarah-allec - I know I broke all the rules of a good PR with this monstrosity, but if you and Claude could take a look at and review it, that'd be great. Thanks!

sarah-allec (approved):
Re-reviewed. Both of my earlier asks are resolved cleanly:

The KB-store swap is now a single `_swap_into_place` used at all four sites — rename the live dir to `.bak`, move staging in, drop the backup only on success, restore it if the publish fails — plus `kb list` surfacing leftover staging dirs. That's exactly the crash-safe ordering, and it's tested. And the campaign threshold moved to 0.35 with the overlap distribution actually measured (unrelated ≤0.20, continuations ≥0.43), sitting mid-gap so the residual error leans toward the cheaper mistake. Both are better than what I flagged — clearing the change request.

Honest scope note: this has grown a lot since I first looked — it now bundles the plan-mode experiment/ideation/document separation, the portfolio renderer, in-place document revision, campaign scoping, and the KB store. I confirmed my prior concerns and the KB-store safety; I did not line-by-line the full ~11k lines of new plan-mode work, but it's well covered by dedicated test suites, which is the basis I'm approving on.

One process thought, take or leave: it's a lot of independent features in one PR — easier to review and safer to revert if future work like this lands in smaller slices.

Approving.

PR #403, opened 2026-07-31
Experiment-validated prediction: reference-property validation + observable-requirements contract + validation panel

sarah-allec:
This adds the simulate-mode machinery for *validate-then-predict*: catch a model when it disagrees with what's independently known, iterate, and only then trust a prediction — all engine-neutral. It's one coherent capability built in two layers.

**Reference-property validation.** Before trusting a novel prediction, validate the model against independently-known reference quantities. `ReferencePropertyCritic` reasons over measured constituent/reference properties and attributes a converged-but-wrong result to the force field; `ReferencePropertySelector` picks the reference property per component; `TrendCritic` catches a physically-backwards trend across a swept series; `ReparameterizationAdvisor` + `run_reparameterization` drive an autonomous source→apply→re-check fix that SciLink self-validates and a human approves. Wired into the pipeline as an opt-in pre-run gate.

**Observable-requirements contract** (`parameters = f(system, observables)`). The target observables become a first-class, typed, engine-neutral input: derived from the goal upstream (`ObservableRequirementsDeriver`, opt-in), threaded into the MD/DFT Generate stage so sampling is chosen to capture them, and checked at the pre-run gate two ways — the gate verifies the declared list instead of re-inferring it, and deterministic `signal_present` / `cadence` checkers (delegating engine detection to a `detect_signal_logging` skill tool, LAMMPS realization included) run alongside the LLM layer, blocking on the union.

**Validation panel.** `run_validation_panel` judges N observables of the actual system against their references and scopes the prediction to what passed.

Design notes: `docs/proposals/{reference-property-validation,experiment-validated-prediction,simulation-analysis-agent}-design.md`.

**Opt-in machinery, one intentional default-path change:** observable derivation is opt-in (off), `required_observables` is optional, and the reference gate is opt-in — those add nothing to an existing run. The one unconditional change is on the post-run critic: `RunCritic` can now attribute a converged-but-physically-wrong run to the force field (`failure_class="force_field"`, `poor` verdict with null `suggested_fixes`), which surfaces as a `force_field_flagged` warning and, at that point, stops the refine loop rather than iterating deck fixes. This is intended (detection wired, fixer to follow) and shifts the verdict distribution on existing pipelines — flagging it explicitly. Fully unit-tested (no API keys/cluster); merged up to current `main` with the integrated suites green.

**Out of scope (follow-ups):** the simulation-analysis agent + the UC2 viscosity/T1 skills + the Deception run; Edison literature grounding (a future caveated fallback).


ziatdinovmax (changes requested):
I really like it and see as a step towards SciLink's ultimate goal, bridging experiment and theory/simulations. For example, in the future a reference may come from the experimental analysis agents and passed via meta-agent to sim agents. And if there's an ambiguity about what to trust (experiments can also be wrong!), then meta can request the planning agent to design experiment to resolve it. Amaze amaze amaze.

Below are technical comments from Claude:

## Findings

### Major (worth addressing before merge or as an immediate follow-up)

1. **The live path of the headline feature is inert, but still spends real
   compute.** `_measure_property_via_short_run` and
   `_search_force_field_parameters` are documented `LIVE-ENV SEAM` stubs
   returning `None`. Consequence with `reference_check=True` in a live run:
   selector LLM call → **real packmol packing + real OpenFF parameterization per
   component** → measurement returns `None` → everything "unmeasured" → critic
   fails open `good` → production proceeds. The gate can never catch anything
   yet, but it does burn per-component packing/parameterization work each run.
   Suggest short-circuiting before packing while the measurement seam is
   unimplemented, and saying plainly in the PR body that the live catch is
   pending the seam.

2. **"Existing runs are byte-identical" is not accurate.** The `RunCritic`
   post-run prompt change (the `force_field` failure class + "judge poor by
   physics reasoning") is unconditional, and the surfacing
   (`force_field_flagged`, warnings, `failure_class` propagation in
   `run_campaign`) is always-on. Behavior delta on the default path: a converged
   run the critic now deems physically wrong returns `poor` with
   `suggested_fixes: null`, so the refine loop **stops** where it previously
   might have iterated deck fixes. This looks intended (detection wired,
   fix later) — but the PR description's default-safety claim should be
   corrected, and the verdict-distribution shift on existing pipelines is worth
   a heads-up.

3. **LAMMPS detector: presence is conflated with interval parseability.**
   `detect_signal_logging` reports `present: False` whenever the interval is not
   a literal integer. Decks using variables — `thermo ${freq}`,
   `dump 1 all custom ${dumpfreq} …` — are common in generated inputs and would
   be flagged as *not logging* signals they do log. Fix: presence from the
   command's existence; `interval_steps: None` when unparseable
   (`signal_present` then passes, `cadence` defers).

4. **`fix ave/correlate` cadence uses Nfreq, not Nevery.**
   `_ave_output_interval` captures group 3 (Nfreq — how often the *correlation
   output* is written, typically large). For Green-Kubo adequacy the governing
   interval is Nevery (group 1) — the fix samples stress internally at that
   cadence. As written, a perfectly adequate `fix v all ave/correlate 5 100 1000 …`
   deck reports interval 1000 and fails a `max_interval_steps=10` cadence
   requirement. Nfreq is right for `ave/time` raw writes; use Nevery for
   `ave/correlate`.

5. **Deterministic findings never reach the fixer.** In `_run_dry_run_gate`,
   `det_blocking` joins `missing` for the pass decision, but the LLM critic that
   produces `suggested_fixes` never sees those messages. When only the
   deterministic layer flags (LLM says clean), `suggested_fixes` is empty → gate
   exits `"unfixed"` → production proceeds with the observable still
   unrecoverable. The union blocks the *status*, not the *run*. Suggest
   threading `det_blocking` into the next-cycle coverage prompt so the fixer can
   actually correct the deck.

   (Mitigation for 3–5: gate failure doesn't abort the campaign, so the cost is
   wasted cycles and misleading history rather than blocked runs.)

### Medium

6. **A cadence-only requirement lets an absent signal escape both checkers.**
   `_check_cadence` defers when the signal is absent ("that is
   `signal_present`'s concern"), but the deriver prompt says to emit `cadence`
   "ONLY when dense sampling is needed" — it doesn't require pairing it with a
   `signal_present` entry. Either make the deriver always pair them, or have
   `cadence` flag absence.

7. **Fragile name matching in `run_reference_check`.** Selections are keyed by
   the LLM-returned `component` string and matched exactly against
   `name or smiles`. A cosmetic mismatch ("EIS (ethyl isopropyl sulfone)")
   silently degrades to unmeasured → fail-open pass. Suggest normalized
   (casefold/strip) matching with a SMILES fallback.

8. **Panel counts unrated entries as passed.** In `run_validation_panel`, a
   judge entry with no `verdict`/`consistent` field lands in `passed` and is
   then cited as "backed by validation against: …". For a trust-scoping
   surface, unrated ≠ validated.

9. **Human approval isn't actually wired on the auto-fix path.** The design and
   PR body say "the human approves the corrected model", but the pipeline calls
   `_reparameterize` with `confirm_fn=None` → auto-approve. Moot while the
   search seam returns `None`, but when it lands, `confirm_fn` should route to
   the refinement `InteractFn`/autonomy policy rather than defaulting open.

sarah-allec:
Thanks Maxim!

Pushed fixes for the findings:

- **Detector (#3, #4):** presence now comes from the logging command existing, not a parseable interval — `thermo ${freq}` / `dump … ${n}` are no longer flagged as unlogged (interval just becomes `None`, so `signal_present` passes and `cadence` defers). And `ave/correlate` cadence now uses **Nevery**, not the correlation-output Nfreq.
- **Panel (#8):** unrated judge entries no longer count as passed or get cited as validated, and they block warranting (new `unrated` bucket).
- **#5:** deterministic gate findings now thread into the *next* cycle's coverage prompt, so a deterministic-only flag reaches the LLM fixer instead of only blocking the status.
- **#1:** the reference check short-circuits before any per-component packing while the measurement seam is a stub (an `_LIVE_MEASUREMENT_IMPLEMENTED` gate) — no more wasted packing/parameterization.
- **#6:** the deriver now pairs a `cadence` requirement with a `signal_present` one for the same signal.
- **#7:** normalized (casefold/strip) name matching with a SMILES fallback in `run_reference_check`.

You're right on **#2** — I corrected the PR body. The `RunCritic` `force_field` verdict change is unconditional, so a converged-but-wrong run now returns `poor` with null fixes and the refine loop stops there; called that out explicitly. And **#9** is moot now that #1 short-circuits the reparameterize path — when the measurement seam lands I'll wire `confirm_fn` to the refinement autonomy policy rather than defaulting open.

Re-requesting review.

ziatdinovmax (approved):

PR #405, opened 2026-08-01
Simulation-analysis agent: engine-neutral data->number via availability-gated skills

sarah-allec:
Adds the **simulation-analysis agent** — the engine-neutral "data → number" step that computes properties (viscosity, T1, RDF, …) from a finished run's output. This is the piece the validation panel consumes, and it fills the gap left when the engine-neutral refactor demoted the old `LAMMPSAnalysisAgent` to a benchmark baseline (its quality role moved to `RunCritic`; its property-computation role was never replaced).

One agent for all simulation analysis; the property × technique differentiation lives entirely in skills, gated by which output data is present — not per-scale agents, not per-data-modality. Design note: `docs/proposals/simulation-analysis-agent-design.md`.

**`BaseAnalysisAgent`** (sim-native — does not subclass the experimental base, so the modality families stay decoupled). The reusable codegen engine: `compute_property` generates a self-contained analysis script, runs it sandboxed with the inputs injected as globals, verifies the returned value is physically plausible, and regenerates from the error on failure (with a pre-run compile check).

**`SimulationAnalysisAgent`** — the concrete pipeline: classify the outputs into data kinds → find the technique skills whose required data is present (**availability gate**) → let the LLM pick which of their properties the goal wants → run each through the base engine. A `requires: trajectory` skill fires only with a trajectory, `requires: dft_output` only with a `vasprun.xml`, and an overlapping property (elastic constants from MD vs DFT) resolves by what was actually run.

**Skills** (`skills/simulation_analysis/`) — `viscosity_greenkubo` (Green-Kubo, `requires: thermo_log`) and `t1_relaxation` (¹H NMR T1 via reorientational dipolar model, `requires: trajectory`), as code-in-markdown recipes. Frontmatter fits the existing convention (`description`, `technique`) plus two new fields the availability gate reads: `computes` and `requires`.

**Live-validated:** a smoke test generated stress with an analytically known Green-Kubo viscosity, and the agent's real-LLM-generated code parses it, computes the value to the right order of magnitude with correct units, and honestly flags non-convergence. That smoke test also caught (and this PR fixes) two real base-engine bugs: the wrapper's `.text` runs a JSON extraction that mangled generated code (now uses `raw_text`), and a missing pre-run compile check. ~20 unit tests, no API keys/cluster.

**Out of scope (follow-ups):** analysis skills for the other scales (molecular DFT / periodic DFT / MLIP-MD) and the comprehensive cross-scale benchmark; wiring `run_analysis` into the validation panel (that panel is in a separate PR); the UC2 Deception runs. Complementary to the observable-requirements contract (which ensures runs *sample* the signals this agent consumes).


ziatdinovmax:
Question: does it make sense to use `SimulationAnalysisAgent` for *forward models* — EXAFS simulation today, and microscopy image simulation in the near future? Those models are built on DFT and MD outputs and compute a property from a finished run (e.g. STM LDOS maps from a DFT charge density), so they fit the agent's sahpe quite nciely.

I'm imagining a sibling skill domain `scilink/skills/forward_models/`, served by the same agent:

 ```
 scilink/skills/
 ├── simulation_analysis/       
 │   ├── viscosity_greenkubo/    
 │   └── t1_relaxation/          
 └── forward_models/             
     ├── stem_multislice/        
     ├── stm_ldos/              
     └── exafs_feff/            
 ```

This would require expanding the `compute_property`'s output type from scalar to include curve, image, and datacube. Doesn't necessarily have to be a part of this PR.

Let me know what you think.

sarah-allec:
Great idea — I think it's a genuinely nice fit and worth pursuing as a follow-up. A forward model *is* the same shape: it takes a finished DFT/MD run and computes an observable from it (STM LDOS map, EXAFS χ(k), STEM image), which is exactly what this agent does — the availability gate and codegen loop already generalize, and a `stm_ldos` skill would just `require: dft_output` and produce an image the way `band_gap` produces a scalar today.

The one real extension is the one you flagged: `compute_property` currently returns a scalar and the verification gate judges a single number, so the output type (and the plausibility check + artifact handling) needs to widen to curve / image / datacube. Agree it needn't be in this PR — I've opened #407 to track it so we design that generalization deliberately. The sibling `forward_models/` domain you sketched is exactly the structure I'd go with.

For reference, the cross-scale analysis side is now validated on real output — the agent recovers the HOMO-LUMO gap (real NWChem), band gap (real VASP), and self-diffusion (real LAMMPS trajectory) from whatever output is present, engine-agnostic, which is the same routing forward models would ride on.

ziatdinovmax:
Sounds good - and thanks for filing an issue to keep track of this. There are a few technical items that Claude has identified - I'll post them in the next comment. Please address them when you can.

ziatdinovmax (changes requested):
Technical issues to address before the merge:

### 1. The agent does not work at all on the internal-proxy path

`_llm` prefers `raw_text` and falls back to `.text`. Only `LiteLLMGenerativeModel`
exposes `raw_text`; `OpenAIAsGenerativeModel` — the `base_url` proxy path — returns
`SimpleNamespace(text=…, candidates=…)` with no `raw_text`. Its `.text` runs
`_fix_json_format`, which does `replace("```", "")` and strips only a leading `json`
tag, so a fenced ```` ```python ```` response becomes a script whose first line is the
bare word `python`. That **compiles cleanly** — the new compile check does not catch it —
and dies at runtime with `NameError: name 'python' is not defined`.

**It never recovers, because the failure is in the wrapper rather than in the code.**
Simulated end-to-end with the model returning *perfect* code every time, differing only
in being fenced (which is why `_clean_code` exists at all): all three attempts fail with
the identical `NameError` and `compute_property` returns `status: "error"`. The same
model output on the LiteLLM path returns `success, value=1.23`. Each regeneration is
re-fenced, re-stripped to a bare `python` line, and dies the same way.

So on `base_url` deployments this agent does not work, for any property. Per CLAUDE.md
the internal proxy is a first-class supported deployment, so this is not a corner case.

---

### 2. An honest "cannot compute" is treated as a crash and retried

The codegen prompt explicitly invites `{"status": "error", "message": …}` when the
property cannot be computed. But `_execute_script` returns
`{"status": "success", "code_path": …, **parsed}` — the spread puts the script's own
`status` last, overwriting `"success"` with `"error"` — so `compute_property` routes a
correct script into the refinement loop.

Confirmed: a script reporting `"stress series too short to converge Green-Kubo"` produces
**3 LLM calls**, `attempts: 3`, and a final `status: "error"`, with the refine prompt
telling the model *"The following analysis script failed. Fix it."*

The wasted calls are the minor cost. The real risk is that the model complies by
inventing a fallback value, which then reads as `success` into a gate that decides
whether to trust a force field — the one failure mode a validation-feeding agent must
not have.

---

### 3. Fan-out runs silently lose all but one member's data

`run_analysis` builds `data_files = {Path(p).name: p …}`, keyed by **basename**. A
parallel stage — a first-class shape in `run_campaign` (temperature sweeps, umbrella
windows) — writes `member_0/log.lammps`, `member_1/log.lammps`, …, which all collide.

Confirmed: three member logs on disk collapse to **1** entry; the other two are dropped
with no warning and no record.

Silent data loss in an analysis agent is the worst kind, because the computed number
still looks fine and nothing downstream can tell that two thirds of the run was never
read.

---

## Borderline — recommended, not strictly blocking

**The plausibility gate cannot see the convergence flags.** `_verify_result` formats only
`task`, `value`, and `units` into the prompt. Confirmed: a Green-Kubo result carrying
`plateau_reached: False` is judged `plausible: True` ("typical liquid range").

Arguable both ways. *Against fixing now:* the flags are present elsewhere in the result
dict, and the agent has no external callers yet, so nothing is being misled today.
*For fixing:* `verification` is the **designated** quality field, so a false green light
there misleads regardless of what else the dict carries — and it fails on precisely the
case both skills were written to catch. Cheap either way; I would fold it in with the
three above.

---

## Deliberately not blocking

The **zero-external-callers** fact demotes most of these: nothing consumes this agent
yet, so any finding whose harm is "downstream is misled" has no downstream. Full detail
is in git history for this file; kept here as a one-line ledger so nothing is lost.

- `classify_outputs` matches bare extensions against the whole filename, so
  `params.func` / `mesh.inc` classify as **trajectory**. Trivial fix, mild harm — fold in
  if convenient (`simulation_analysis_agent.py:82`).
- `_select_properties`' `return selected or eligible` computes *every* eligible property
  when the LLM names nothing, inverting the step's stated purpose. Cost, not correctness
  (`:153`).
- `_extract_json`'s regex fallback returns the first object, not the last, contradicting
  its docstring. Multi-line fallback only; the prompt asks for single-line
  (`base_analysis_agent.py:310`).
- No numeric uncertainty on the computed value — moves to #407 with the result contract.
- Neither new class is exported from `sim_agents/__init__.py` (which exports all 19 other
  agents); `_load_skills_to_state` is dead code; `in_container` is computed and never
  read; both skills' section order differs from CLAUDE.md's recommended order;
  `classify_outputs` breaks on the first matching kind, so an overlapping pattern
  resolves by skill iteration order.

sarah-allec:
Thanks Maxim — all four of the blocking/borderline ones are fixed in add933a5, plus a few of the ledger items.

The proxy codegen bug was the important catch. `openai_wrapper` now carries `raw_text` (the uncleaned content) alongside `text`, mirroring the LiteLLM wrapper, so codegen on a `base_url` deployment parses the model's actual script instead of one the JSON cleaner mangled. That's the one that would have bitten every proxy user silently.

For the honest-error-as-crash: `_execute_script` returns an `ok` flag now — script ran to completion (success *or* an in-script error it reported honestly) is `ok=True` and does not retry; only a crash / no-JSON is `ok=False` and retries. So a script that correctly reports "no plateau, can't compute" is accepted as-is rather than being re-rolled until the model invents a number. And `_verify_result` now formats the scalar result fields into the prompt and tells the verifier a false `plateau_reached`/`linear_regime`/`converged` is not trustworthy, so the designated quality field stops green-lighting the exact case the two skills exist to catch.

Fan-out data loss: `data_files` is keyed by path relative to `run_dir` now, so `member_0/log.lammps` and `member_1/log.lammps` keep distinct keys instead of collapsing to one.

From the ledger I also folded in the cheap correctness ones: `classify_outputs` matches full filename or exact extension (no more `mesh.inc` → trajectory), `_extract_json`'s regex fallback returns the last balanced object per its docstring, and both classes are exported from `sim_agents/__init__.py`.

Left as-is deliberately: `_select_properties`' fall-back-to-all when the LLM names nothing — I'd rather it compute the eligible set than return nothing, and there's no correctness harm, just cost. The remaining ledger items (dead `_load_skills_to_state`, unused `in_container`, skill section order) I'll sweep with the #407 result-contract work rather than churn this diff further.

21 analysis-agent tests green. Re-requesting review.

ziatdinovmax (approved):

PR #410, opened 2026-08-03
read_file: return literature reports whole, add offset, index truncated reads

ziatdinovmax:
## What was wrong

A planning agent asked for a literature report and only ever saw the first question in it.

Literature reports put their questions one after another, and `read_file` returns the first 200 lines by default. In the report that exposed this, question 1 ends at line 197 and question 2 begins at line 198 — so the default read stopped two lines short of everything else. The agent read the same file **seven times**, concluded from the identical results that the file "repeats" question 1, and wrote its note without questions 2–4. It marked the missing numbers `[TBD]`, as though the literature were thin.

It wasn't thin. **70% of the retrieval was never opened** — 86,000 characters and 21 of 27 papers. The answers it needed were all there.

The worst part is that this is invisible in the output. The note reads as complete and carefully hedged; only the trace shows otherwise. The agent recorded a tool failure as a property of the evidence.

## What this changes

- **Literature reports come back whole.** Up to a 250k-character cap, sized against the real corpus (89k–262k, median 169k) — a smaller cap would exclude precisely the multi-question reports that most need a whole read.
- **`offset=` reads from any line**, so the middle of a file is reachable. `tail=` and `search=` were added for an earlier instance of this same bug and only ever covered the end and presence/absence.
- **A truncated read now says so, and says what it is missing** — it lists the section headings with their line numbers and names `offset=` as the way to reach them. An agent can no longer mistake a short read for a short file, which is what happened here.

Across the session's 14 reports, question sections reachable in one default read go from **17/56 to 51/56**. The remainder exceed the cap and are navigable via the emitted outline.

<details>

**`scilink/agents/planning_agents/orchestrator_tools.py`**

- `_FULL_READ_STEMS` / `_FULL_READ_MAX_CHARS` gate the whole-file path; matched on filename, and only when `offset`/`tail` are not in play.
- New `offset` branch in the head/tail block, plus an `offset` entry in the tool's `parameters` so the LLM can see it.
- `_outline()` collects `#`-prefixed headings with line numbers (capped at 12) and is appended to both the offset and head truncation notices.
- Truncation notice reworded to state explicitly that it is a truncated read rather than the whole file.
- **Bug fix:** `truncated` was computed as `total > max_lines`, so it reported `true` even when the whole file had just been returned — the exact signal that would prompt an agent to read again.

**`tests/test_read_file_offset_and_index.py`**

Lifts the real `read_file` body out of the source with `ast` and exercises it — no orchestrator construction needed. 15 assertions across five groups:

1. the regression itself, against the exact file that failed (792 lines, 4 questions) — all four sections in one default call, `truncated=false`
2. `offset=198` and `offset=579` land precisely on questions 2 and 4
3. an ordinary long file still truncates, and the notice carries the outline and names `offset`
4. a lit report over the cap falls back to truncation but still emits the outline
5. regressions — `tail`, `search`, whole small reads, missing-file error

Run: `python tests/test_read_file_offset_and_index.py` (exits non-zero on failure).

**Base branch:** targets `fix-ui-resume-session` rather than `main` — that branch is 60 commits behind main, so a PR to main would show ~7,350 unrelated deletions. Retarget when the base lands.
</details>

### DIFF
--- a/scilink/agents/planning_agents/orchestrator_tools.py
+++ b/scilink/agents/planning_agents/orchestrator_tools.py
@@ -4844,8 +4844,23 @@ class OrchestratorTools:
         # =====================================================================
         # READ FILE (non-destructive inspection)
         # =====================================================================
+        # A literature report is read for its whole content, never its head:
+        # its questions sit in sequence, so a head-only read hides every
+        # question but the first. Live, an agent read one seven times, decided
+        # the file "repeats" Q1, and wrote the note without Q2-Q4 — 70% of the
+        # retrieval unused, recorded as [TBD] as though the literature were
+        # thin. These files get the whole body up to a cap.
+        # Cap sized against the real corpus: session lit reports run 89k-262k
+        # chars (median 169k), so a smaller cap would exclude precisely the
+        # multi-question reports that most need a whole read. Past the cap the
+        # read truncates but emits the section outline, so the agent can
+        # offset= to each question instead of being stuck at the head.
+        _FULL_READ_STEMS = ("literature_search",)
+        _FULL_READ_MAX_CHARS = 250_000
+
         def read_file(file_path: str, max_lines: int = 200,
-                      tail: bool = False, search: str = None) -> str:
+                      tail: bool = False, search: str = None,
+                      offset: int = None) -> str:
             """
             Read and return the contents of a file. Use this to inspect
             plans, protocols, configs, logs, or any text/JSON file without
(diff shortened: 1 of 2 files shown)


sarah-allec (approved):
This is a great catch — the truncated = total > max_lines bug is the kind that's nearly impossible to see from the output, since the note reads as complete and carefully hedged while 70% of the retrieval was never opened. Returning the lit reports whole, adding offset=, and emitting the section outline on a truncated read all hang together well, and I like that the test runs the real read_file body against the actual 792-line file that failed rather than a synthetic stand-in.

One small thing worth a follow-up, not blocking: an offset past EOF returns empty content with a backwards range in the note ("showing lines 1000-792 of 792"). Offsets coming from the emitted outline are always valid, but a hallucinated one lands in exactly the ambiguous-empty-read state this PR is trying to kill — a clamp or an explicit "offset past end of file (N lines)" would close it. And the 250k whole-read is a sizable context injection per report; deliberate and justified here, just worth keeping an eye on if a session pulls several big reports.

Also minor: the test hardcodes an absolute /Users/... path for both the source and the fixture, so it won't run outside your machine — worth making relative when you get a chance (same as the session-title test).

Approving — the fix is right and the regression coverage is solid.

PR #415, opened 2026-08-04
Dataset-aware metadata ownership; plan validator re-derives series regimes (#411)

### LINKED ISSUE #411: Session metadata is reused across techniques, and plan validation does not repair the series regime
When a session runs more than one technique, metadata loaded for the first
technique is silently reused for the second, and neither skill selection nor
plan validation recovers from it.

Three related defects. The first creates the wrong state; the second and third
are why it survives into the fit.

---

## 1. Session-level metadata slot is sticky across datasets

`current_metadata` is a single session-level slot
(`analysis_orchestrator.py:632`). `load_metadata` writes the loaded document
into it wholesale (`analysis_orchestrator_tools.py:756`). Nothing records
which dataset that metadata belongs to.

`run_analysis` resolves metadata per dataset **only** when the dataset carries
per-file sidecar JSONs — in that case the slot is cleared and per-file
metadata is used (`analysis_orchestrator_tools.py:2284`):

```python
if has_sidecars:
    self.orch.current_metadata = {}
else:
    # Single-file backstop: ... This ONLY runs when no metadata was
    # loaded (so it can never override an explicit load); ...
```

Otherwise it falls through to a stem-matched single-file sidecar, and finally
errors only when the slot is empty (`analysis_orchestrator_tools.py:2304`):

```python
if self.orch.current_metadata is None:
    return json.dumps({"status": "error", "message": "No metadata available. ..."})
```

So for a second dataset in the same session that has no sidecar of its own:
`has_sidecars` is false, the stem-matched backstop finds nothing, and
`current_metadata` is not `None` — it still holds the first technique's
document. No error is raised and the stale metadata is applied to the new
dataset.

The guard at line 2290 was written to stop the backstop clobbering an explicit
load. It also makes an explicit load from an earlier, unrelated dataset sticky
for the rest of the session.

**Consequence.** The planner reads the second dataset's axes through the first
technique's schema — column roles, units and control variable are all taken
from metadata describing different data. The run does not fail; it produces a
confidently wrong plan for the wrong technique, and a technique-specific skill
that would otherwise be selected can be dropped.

**Why the sidecar path is unaffected.** Per-file sidecars resolve per dataset
by construction (line 2284 clears the slot). The gap is specific to a single
combined metadata document loaded via `load_metadata`: it never enters the
sidecar path and there is no per-dataset ownership recorded.

**Suggested direction.** Make the slot dataset-aware rather than
session-global: record the `data_path` that populated `current_metadata` at
line 756, and in `run_analysis`, when the requested dataset differs from that
path and nothing else resolves, refuse to reuse it. Erroring with a message
naming the previous owner is preferable to silent reuse — a plausible-looking
fit against the wrong schema is worse than a stopped run. Reusing the existing
sidecar semantics (clear, then resolve) keeps both paths consistent.

---

## 2. Plan validation rewrites the model but not the series regime

`CurveFitting._validate_plan`
(`controllers/curve_fitting_controllers.py:2042`) can re-identify the technique
from the data and rewrite the plan. It already has an apply path for the
series plan (`:2125`):

```python
if result.get("series_analysis_plan"):
    self._extract_series_plan(state, result)
```

and the current regime block is passed *into* the prompt, so the validator can
see it.

The gap is the response schema (`instruct.py:2694`):

```
    "series_analysis_plan": null
```
```
Include series_analysis_plan only if this is a series with regimes that need revision.
```

The field defaults to `null` and its inclusion is opt-in, with nothing
indicating that a change of technique or physical model necessarily
invalidates the regimes. In practice the validator corrects
`physical_model`, `parameters_to_extract` and `fitting_strategy` while leaving
`series_analysis_plan` untouched.

Because the regime is what drives the per-frame fit, correcting the model
sections alone is not sufficient — the series still runs under the previous
technique's regime and parameters.

**Suggested direction.** State the principle in the validation prompt: if the
correction changes the technique or the physical model, the series regimes
must be re-derived rather than carried over. No code change is needed; the
apply path at `:2125` already handles a returned plan. A regression test that
feeds a plan for one technique against data from another and asserts the
regime changes would cover it.

---

## 3. Skill selection consumes the stale metadata, and is never revisited

Skill selection runs **before** planning and validation:
`CurveFittingSkillSuggestionController`
(`controllers/curve_fitting_controllers.py:1143`) builds its context with
metadata first, then summary statistics, then the rendered plot (`:1160`
onward):

```python
context_parts.append(f"Metadata: {sysinfo.strip()[:1500]}")
...
context_parts.append(f"Data statistics: {stats}")
...
context_parts.append({"mime_type": "image/png", "data": plot_bytes})
```

Curve-fitting selection is *exclusive* — it picks the single best technique
match or none — so metadata that confidently describes a different technique
suppresses the correct technique skill even though the statistics and plot
are also present and describe the real data.

`select_relevant_skills` is invoked in exactly one place (`:1194`). Nothing
re-runs selection later, so when `_validate_plan` subsequently re-identifies
the technique from the data (defect 2), the skill set is **not** revisited.

**Consequence for sequencing.** Fixing defect 2 alone is not sufficient: the
plan sections would be corrected while the run still proceeds without the
technique's skill loaded. Defect 1 is therefore the load-bearing fix — it
prevents the wrong evidence from reaching selection in the first place.

A secondary hardening question, separable from the above: selection currently
treats metadata as leading evidence even when the data statistics and plot
contradict it. Whether a disagreement between metadata and data should
downweight the metadata, or surface as a warning, is worth deciding
independently of the metadata-scoping fix.

---

## Notes

All three are reproducible on current `main` and on a fresh clone. Fix order matters: (1) is load-bearing, since it prevents the wrong evidence
from reaching both skill selection (3) and planning. (2) is self-contained and
prompt-only and can proceed in parallel. (1) touches session-state semantics
that the meta fan-out path also relies on, so it deserves the more careful
treatment.

Providing an explicit instruction in the request naming the correct technique
does reach the regime and produces a correct plan, which confirms the regime
is reachable and that the gap is in the validator's own rewrite path.

ziatdinovmax:
Fixes #411.

When one session analyzes more than one technique, metadata loaded for the first dataset silently carried over to the next one. The planner then read the new data through the old technique's schema — axes, units, and control variable all wrong — and the correct technique skill was dropped. The run didn't fail; it produced a confidently wrong plan. A second, related gap: when the plan validator re-identified the technique from the data, it corrected the model sections but left the series regimes from the old technique in place, so a series still fit under the wrong recipe.

**What changes**

- Loaded metadata now belongs to a dataset. A freshly loaded document binds to the first dataset it is used for; a later analysis of a *different* dataset re-resolves metadata on its own (per-file sidecar JSONs, then a stem-matched sidecar). If nothing resolves, the run stops with an error naming which dataset the loaded metadata belongs to — one `load_metadata` call away from recovery — instead of silently fitting against the wrong schema.
- This also repairs a stricter case: previously, once *any* metadata was loaded, sidecar detection was skipped entirely — so even a dataset carrying its own sidecars inherited the stale document.
- The plan validator's prompt now states that a technique or physical-model correction invalidates the series regimes and must return a re-derived series plan (with the regime fields named explicitly).
- Ownership survives checkpoint/restore; older checkpoints backfill it from the last data path.

**Validation**

- Offline: `tests/test_metadata_dataset_ownership.py` — 31 checks (ownership semantics, refusal naming the owner, sidecar displacement, bind-on-first-use, checkpoint round-trip, and the validator apply path with a mocked response). Neighboring suites unchanged and green (`test_run_analysis_glob`, `test_file_prep_metadata_only`, `test_metadata_conformance_energy_axis`, `test_analyze_file_shapes`).
- Live (`bedrock/us.anthropic.claude-opus-4-8`):
  - `tests/test_metadata_ownership_live.py` (9/9) — a wrong-technique plan validated against a mismatched series overlay gets both the model *and* the regimes re-derived; a two-turn session shows the refusal firing for a second dataset with no sidecar, followed by autonomous recovery and correct re-binding.
  - `tests/test_411_multimodal_meta_live.py` (10/10) — one meta session, three delegations over real instrument exports (thermal → in-situ diffraction → in-situ spectroscopy), with the later turns naming no technique: no silent reuse of the thermal document, `xrd_profile` auto-selected for the diffraction series, plan free of thermogram leakage, and the slot re-resolved per dataset every time. Dataset paths are env-configured so the harness runs against any trio of this shape.

<details>
<summary>Technical details</summary>

**Defect 1 (load-bearing).** `current_metadata` was a single session-level slot (`analysis_orchestrator.py`); `load_metadata`/`convert_metadata` wrote into it wholesale and `run_analysis` error-checked only `if current_metadata is None`, so a second dataset without a sidecar inherited the first dataset's document. The whole resolution ladder also sat inside that `None` guard, so once the slot was filled, per-file sidecar detection never ran for later datasets.

Fix: `current_metadata_owner` on the orchestrator, with **bind-on-first-use** semantics. `_replace_metadata` (the single choke point for all load paths) resets the owner to unbound; `run_analysis` binds the slot to the dataset that first consumes it. On an owner mismatch the existing ladder re-runs — per-file sidecars (clear slot, resolve per file), stem-matched sidecar backstop, then a refusal naming the previous owner. Bind-on-first-use is what keeps "load metadata, then analyze the next dataset" working — a fresh explicit load is never refused. `_dataset_key`/`_same_dataset` normalize paths and define coverage: a dir covers its member files and globs within it; a glob covers files it matches; siblings and broader paths are distinct datasets. Owner is persisted in both checkpoint writers and restored; legacy checkpoints without the field backfill from `current_data_path`.

**Defect 2.** `CurveFitting._validate_plan` already had the apply path for a returned `series_analysis_plan`; the response schema made it opt-in (`"series_analysis_plan": null`) with nothing tying regime validity to the technique. The prompt now states the principle and names the exact regime fields — the first live run showed the validator returning re-derived regimes under improvised keys (`model`/`params`), which the extractor then backfilled with placeholders.

**Defect 3** (skill selection consuming stale metadata) needed no new mechanism: selection runs before planning and reads whatever metadata reaches it, which after defect 1's fix is either the right document or none. A post-validation re-selection pass was deliberately not added.

**Known non-goal:** explicitly loading wrong metadata for the *same* dataset is still honored — user authority over their own dataset's metadata is out of scope here.

Observed while testing, not addressed here: delegations through `run_task` never write the child's `checkpoint.json`, and the delegation ledger still trusts the child's self-reported status (plan item 1.4).
</details>

### DIFF
--- a/scilink/agents/exp_agents/analysis_orchestrator.py
+++ b/scilink/agents/exp_agents/analysis_orchestrator.py
@@ -630,6 +630,11 @@ class AnalysisOrchestratorAgent:
         
         # Session state
         self.current_metadata: Optional[Dict[str, Any]] = None
+        # Dataset that current_metadata belongs to. None = unbound (freshly
+        # loaded, not yet consumed by run_analysis). Binds on first use; a
+        # run_analysis on a *different* dataset must re-resolve rather than
+        # silently reuse another dataset's metadata (issue #411).
+        self.current_metadata_owner: Optional[str] = None
         self.current_data_path: Optional[str] = None
         self.current_data_type: Optional[str] = None
         self.selected_agent_id: Optional[int] = None
@@ -1284,6 +1289,13 @@ class AnalysisOrchestratorAgent:
                 state = json.load(f)
             
             self.current_metadata = state.get("current_metadata")
+            # Older checkpoints predate ownership: backfill from the last data
+            # path so restored metadata stays bound to the dataset it served,
+            # instead of silently binding to whatever runs next.
+            self.current_metadata_owner = state.get("current_metadata_owner") or (
+                state.get("current_data_path")
+                if state.get("current_metadata") is not None else None
+            )
             self.current_data_path = state.get("current_data_path")
             self.current_data_type = state.get("current_data_type")
             self.selected_agent_id = state.get("selected_agent_id")
@@ -1552,6 +1564,7 @@ class AnalysisOrchestratorAgent:
             checkpoint_data = {
                 "timestamp": datetime.now().isoformat(),
                 "current_metadata": self.current_metadata,
+                "current_metadata_owner": self.current_metadata_owner,
                 "current_data_path": self.current_data_path,
                 "current_data_type": self.current_data_type,
                 "selected_agent_id": self.selected_agent_id,
--- a/scilink/agents/exp_agents/analysis_orchestrator_tools.py
+++ b/scilink/agents/exp_agents/analysis_orchestrator_tools.py
@@ -15,6 +15,7 @@ LLM reasoning.  If extraction fails, the user is prompted and shown the
 sidecar contents to help them specify the variable manually.
 """
 
+import fnmatch
 import glob
 import hashlib
 import json
@@ -275,6 +276,59 @@ def _resolve_glob_files(pattern: str) -> tuple[list[Path], list[Path]]:
         if sidecar.is_file():
             all_files.append(sidecar)
     return data_files, all_files
+
+
+def _dataset_key(data_path: str) -> str:
+    """Canonical form of a dataset reference for ownership comparison.
+
+    Globs keep their pattern (resolved to an absolute base) so two different
+    patterns over the same folder stay distinct datasets; real paths resolve
+    to their absolute form.
+    """
+    if _is_glob(data_path):
+        p = Path(data_path)
+        try:
+            return str(p.parent.resolve() / p.name)
+        except OSError:
+            return data_path
+    try:
+        return str(Path(data_path).resolve())
+    except OSError:
+        return data_path
+
+
+def _same_dataset(owner: str, requested: str) -> bool:
+    """Does metadata bound to *owner* also cover *requested*?
+
+    True when they are the same path, when *requested* drills into the
+    owner directory (a member file, or a glob selecting files within it),
+    or when *requested* is a file the owner glob matches. A broader or
+    sibling dataset is NOT covered — reuse must be re-resolved.
+    """
+    owner_key = _dataset_key(owner)
+    req_key = _dataset_key(requested)
+    if owner_key == req_key:
+        return True
+
+    owner_is_glob = _is_glob(owner)
+    req_is_glob = _is_glob(requested)
+
+    if not owner_is_glob:
+        owner_p = Path(owner_key)
+        if owner_p.is_dir():
+            # Member file/dir of the owner folder, or a glob over it.
+            probe = Path(req_key).parent if req_is_glob else Path(req_key)
+            try:
+                probe.relative_to(owner_p)
+                return True
+            except ValueError:
+                return False
+        return False
+
+    # Owner is a glob: a single file it matches belongs to its dataset.
+    if not req_is_glob:
+        return fnmatch.fnmatch(req_key, owner_key)
+    return False
 
 
 def _detect_sidecar_jsons(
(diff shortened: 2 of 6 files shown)


tkavinchakravarthy-del (approved):
Thanks 😁

PR #420, opened 2026-08-04
lammps validate_script: respect unfix lifecycle, skip defined template vars

### LINKED ISSUE #409: LAMMPS validate_script: two false positives on correct sequential-ensemble decks
`validate_script` in `scilink/skills/molecular_dynamics/lammps/lammps.py` reports two spurious **errors** on a physically correct single-file deck (minimize → NPT-equilibrate → `unfix` → NVT-produce). The gate is fail-open so it doesn't block a run, but it cries wolf on a valid deck, triggers wasted LLM "fixing" cycles, and erodes trust in the pre-run critic.

Surfaced by a real UC2 electrolyte deck. The *staged* two-file campaign validates clean; only the combined single-file deck trips these.

### Repro

```python
from scilink.skills.molecular_dynamics.lammps.lammps import validate_script
# deck: fix NPT all npt ...; run; unfix NPT; fix NVT all nvt ...; with a
# GK viscosity block whose *comment* mentions ${scale} and whose code
# defines `variable scale ...` and uses v_scale.
print(validate_script("run.lammps")["errors"])
# ["fix nvt and fix npt both on group(s): {'all'}",
#  "Unresolved template variables: ['${scale}']"]
```

### Bug 1 — nvt/npt overlap ignores `unfix` (lines ~872-909, ~1062)

`npt_groups` / `nvt_groups` accumulate the group of every `fix … npt/nvt` line across the whole file; the overlap check `npt_groups & nvt_groups` then fires. Nothing consumes `unfix`. Any legitimate sequential deck — NPT equilibration, `unfix`, then NVT production on `all` — is flagged as if both integrators ran at once.

Fix: model the fix lifecycle. Walk in command order, keep an active map `fix_id -> (group, style)`, add on `fix`, drop on `unfix <id>`, and flag a conflict only when an nvt and an npt fix on the same group are active **simultaneously**.

### Bug 2 — unresolved-template regex scans comments and rejects valid `${var}` (line ~1111)

```python
templates = re.findall(r'\$\{[a-z_]+\}|\{[a-z_]+\}', content, re.IGNORECASE)
```

Two problems: (a) `content` includes comment lines, so a comment that merely *mentions* `${scale}` is flagged; (b) `${scale}` is valid LAMMPS syntax when a `variable scale …` is defined earlier in the deck. In the repro, `${scale}` appears only in a comment and `variable scale` is defined.

Fix: strip comments before scanning, and don't flag a `${name}` / `{name}` whose `name` has a matching `variable <name>` definition in the deck. If the intent is to catch SciLink's own unrendered templating rather than LAMMPS variable syntax, use a marker that can't collide with `${...}`.

### Impact
- Pre-run critic reports false errors on correct decks; fail-open so runs proceed, but the noise is real and drives needless refinement iterations.
- Verified on the UC2 80-20 deck: single-file `run.lammps` → both errors; staged `run_equilibration.lammps` / `run_production.lammps` → clean.

DavidPandleton:
Fixes the two false positives in `validate_script` from #409 on a correct sequential deck.

- Track the fix lifecycle: `unfix <id>` now removes a fix, so NPT equilibration then `unfix` then NVT production on the same group no longer trips the nvt/npt overlap error. Only fixes still active at the end of the script are compared.
- The template scan now skips comment lines and treats `${name}` / `{name}` as resolved when a matching `variable name` is defined in the deck. A comment that merely mentions `${scale}` no longer fails validation.

Closes #409

### DIFF
--- a/scilink/skills/molecular_dynamics/lammps/lammps.py
+++ b/scilink/skills/molecular_dynamics/lammps/lammps.py
@@ -868,9 +868,11 @@ def validate_script(
     commands_seen: set = set()
     command_order: List[str] = []
 
-    # Track thermostat/barostat per group
-    npt_groups: set = set()
-    nvt_groups: set = set()
+    # Track the fix lifecycle: fix_id -> (group, style). `unfix` removes an
+    # entry, so only fixes active when the script ends participate in the
+    # nvt/npt overlap check below.
+    active_fixes: Dict[str, Tuple[str, str]] = {}
+    defined_vars: set = set()
 
     for line in lines:
         stripped = line.strip()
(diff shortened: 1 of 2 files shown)


DavidPandleton:
@ziatdinovmax ping, PR for #409 ready for review. Both false positives reproduced and fixed, 3 regression tests added (137 lammps tests pass locally).

PR #428, opened 2026-08-05
Add per_atom_quantile_uncertainty simulation-analysis skill

jenna1701:
Summary

  Adds a new simulation_analysis skill, per_atom_quantile_uncertainty, that computes per-atom epistemic uncertainty for an MLIP trajectory using the PNNL UQ-MLIP (https://github.com/pnnl/UQ-MLIP) method — a quantile gradient-boosting model (GBM) trained on the potential's own per-atom embeddings.

  The skill is purely post-hoc analysis: it trains the GBM from a validation set and predicts uncertainty from a saved trajectory. It does not run MD and does not attach the UQ-decorated calculator.

  What's included

  - scilink/skills/simulation_analysis/per_atom_quantile_uncertainty/per_atom_quantile_uncertainty.md — a single knowledge-bundle skill with:
    - Frontmatter: technique: quantile_gbm, computes: [per_atom_uncertainty], requires: [trajectory] — so the SimulationAnalysisAgent availability gate  selects it only when a trajectory is present on disk.
    - Overview / Planning / Implementation / Interpretation / Validation sections following the fixed section vocabulary.
    - An Implementation recipe (injected verbatim into the codegen loop) that uses the uq_mlip public API — get_extractor → extract → UQModel.fit/from_dir → uq.uncertainty — reading frames only via ase.io.read, and emitting the standard {"status": "success", "value": ...} JSON contract shared by sibling skills.
    - Install instructions for the uq_mlip package with backend-matched extras ([chgnet] / [mace] / [uma]), the e3nn incompatibility note, and platform caveats.

  Verification

  - Skill loads through scilink.skills.loader; frontmatter parses and the implementation section is exposed to the agent's codegen loop.
  - Availability gating confirmed: eligible only when a trajectory data kind is present; correctly excluded when only thermo_log exists.

Generated with Claude Code (https://claude.com/claude-code)

### DIFF
--- a/scilink/skills/simulation_analysis/per_atom_quantile_uncertainty/per_atom_quantile_uncertainty.md
+++ b/scilink/skills/simulation_analysis/per_atom_quantile_uncertainty/per_atom_quantile_uncertainty.md
@@ -33,7 +33,27 @@ pipeline; this skill orchestrates it.
 
 ## Planning
 
-Prerequisites and choices before running:
+**Install UQ-MLIP.** This skill depends on the `uq_mlip` package (the PNNL
+[UQ-MLIP](https://github.com/pnnl/UQ-MLIP) repo). Install it with the extra that
+matches the MLIP backend you will use — the extractor for a backend is only
+importable once that backend's dependency stack is present:
+
+```bash
+pip install -e ".[chgnet]"   # CHGNet extraction (shares an env with either below)
+pip install -e ".[mace]"     # MACE extraction
+pip install -e ".[uma]"      # UMA extraction
+```
+
+Run from a clone of the UQ-MLIP repo (requires Python 3.11–3.13). The base
+package pulls in `ase`, `numpy`, `pandas`, and `xgboost`; the extra adds the
+selected potential. **MACE and UMA depend on incompatible `e3nn` versions** — if
+you need both, use separate environments; CHGNet has no `e3nn` dependency and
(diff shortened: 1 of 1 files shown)


sarah-allec (approved):
This is a really clean skill — the frontmatter, availability gating (requires: [trajectory]), the section vocabulary, and the JSON contract all line up with the sibling simulation_analysis skills, and the install / backend-matching notes (the e3nn incompatibility, the macOS libomp and FAIRCHEM_CACHE_DIR caveats) are genuinely helpful.

One thing I wanted to raise as a question rather than a change, since you know the UQ-MLIP method far better than I do. The flagged-fraction metric computes thr = np.percentile(flat, 95) from the trajectory's own uncertainties and then flags atoms above it — which, as I read it, comes out to ~5% essentially by construction, since 5% of any distribution sits above its own 95th percentile, regardless of how in- or out-of-distribution the run actually is. Should the threshold instead come from the validation set's uncertainties (the in-distribution reference), so the flagged fraction reflects how far the trajectory departs from what the potential was validated on? The recipe comment ("a common choice is the validation upper tail as a threshold") reads like that was the intent, and the Interpretation note about a "> 10% flagged" warning would then actually be reachable (with the current computation it's pinned near 5%). Entirely possible I'm misreading how the GBM brackets are meant to be applied here, so I'll defer to you.

Approving — this is a good addition and the question above doesn't need to block it.

Issue #429, opened 2026-08-05
Is `machine_learning_potentials` a scale, or a potential source?


sarah-allec:
While building routing benchmarks against `SimulationRouter`, we hit a problem that looks like a layering question rather than a benchmark artifact.

## Observation

`discover_scale_agents()` presents four scales: `periodic_dft`, `molecular_qc`, `molecular_dynamics`, `machine_learning_potentials`. The first three name a physical regime. The fourth names how forces are evaluated.

The code seems to agree:

- `md_simulation_agent.py:317` — the MD agent can "*run* a simulation with any potential — classical FF or MLIP"
- `md_simulation_agent.py:441` — "the MD agent doesn't care whether the potential is an MLIP"
- `mlip_agent.py:1030` — returns `"potential": DeployedPotential,   # hand to the MD agent`

So structurally, `MDSimulationAgent` runs the dynamics, and `ForceFieldAgent` and `MLIPAgent` are both potential providers feeding it.

## Why it bites in practice

The MLIP routing criterion isn't a property of the system's physics. `DEFAULT_SCALE_DESCRIPTIONS["machine_learning_potentials"]` (`simulation_router.py:67`) says "best for systems ... without good classical force fields" — that's a fact about force-field coverage for a chemistry, which isn't known at routing time.

Concretely: for the other three scales, a prompt can be phrased as system + observable and the scale inferred ("compute the work function of Cu(111)" → periodic DFT). For MLIP we could not write a single prompt that didn't name the method, because the same physics question is answerable by classical MD whenever a force field exists. All three of our MLIP routing prompts had to say "using a pretrained machine-learning interatomic potential", which makes them string matches rather than routing tests. We've removed them from the routing benchmark for that reason.

## Question

Is `machine_learning_potentials` intended as a scale, or would it sit better as a potential-source decision downstream of `molecular_dynamics`, chosen against `ForceFieldAgent` once the chemistry is known?

Not proposing a change — routing to MLIP does select which agent runs first, so it may well be deliberate. But if the layering is as above, the router commits to a potential type before anything about force-field coverage is established.

Happy to send a PR in whichever direction you prefer, or to leave it as a documentation note if the current design is intended.

ziatdinovmax:
My vote is for “potential source.”

sarah-allec:
That's what I was leaning toward too - I'll get to work on it!

sarah-allec:
[...]

PR #432, opened 2026-08-07
feat: MLIP is a potential source, not a simulation scale (#429)

sarah-allec:
Follow-up to #429 — writing down what "potential source" implies, so the design can be argued about before any code is written.

**Docs only.** No behaviour change in this PR.

## Why a proposal rather than a patch

The change touches `simulation_router`, `simulation_pipeline`, `structure_planner`, two orchestrator prompt sites, and `available_software`. More importantly, the central question — *who owns the classical-FF-vs-MLIP decision?* — is a design call I don't think an outside contributor should make unilaterally. `docs/proposals/` seemed like the right venue given the five existing docs there.

## The evidence

A 26-query routing sweep (claude-opus-4-8, 2026-08-06) scored 22/26, and all four misses went the same way:

| query | expected | routed to |
|---|---|---|
| thermal conductivity of **amorphous** Si | `molecular_dynamics` | `machine_learning_potentials` |
| melting point of Cu | `molecular_dynamics` | `machine_learning_potentials` |
| water density profile at a TiO₂ surface | `molecular_dynamics` | `machine_learning_potentials` |
| Li diffusion coefficient in LiCoO₂ | `molecular_dynamics` | `machine_learning_potentials` |

The behaviour is coherent rather than noisy: MD-sampling tasks on **molecular liquids and biomolecules** (water density/self-diffusion, aqueous NaCl conductivity, LiPF₆/EC viscosity, solvated glycine, peptide unfolding, benzene hydration free energy) all routed to `molecular_dynamics` and all scored correct. The same kind of task on **inorganic solids, metals and interfaces** routed to `machine_learning_potentials`.

That is the router faithfully applying `DEFAULT_SCALE_DESCRIPTIONS["machine_learning_potentials"]` — *"best for systems without good classical force fields"*. **I want to be clear that this is not a routing bug.** That criterion is a fact about force-field coverage for a chemistry, not about the system's physics, and it isn't knowable at routing time.

## What the doc says

- Most of the mechanism already exists: `DeployedPotential` is the engine-neutral seam, `MDSimulationAgent.run()` already accepts one (and its comments at `md_simulation_agent.py:317,441` already state the intended layering), and `MLIPAgent` already returns one annotated `# hand to the MD agent`.
- What's missing is a **potential-selection step** — plus the observation that `simulation_pipeline`'s MLIP branch calls `MLIPAgent.deploy_pretrained(runner="ase")` directly and never passes `potential=` to `MDSimulationAgent`, so it bypasses that seam rather than feeding it.
- Blast radius, and four open questions left deliberately unanswered — ownership of potential selection, what decides FF-vs-MLIP (a checkable coverage rule would be more testable than a judgment call), back-compat for callers passing the scale explicitly, and whether `structure_planner`'s parallel four-way axis moves too.

## What this does not claim

The router isn't broken; `MLIPAgent`'s capability doesn't change; and I take no position on the selection policy itself. Worth noting on that last point: `ForceFieldAgent` currently has no benchmark coverage, so "was this the right potential for this chemistry?" is unmeasured either way — which is a gap on our side, not yours.

Happy to implement whichever direction you prefer once the ownership question is settled, or to close this if you'd rather keep the design discussion in #429.

### DIFF
--- a/docs/proposals/mlip-as-potential-source.md
+++ b/docs/proposals/mlip-as-potential-source.md
@@ -1,0 +1,163 @@
+# Proposal: MLIP is a potential source, not a simulation scale
+
+Follow-up to [#429](https://github.com/ziatdinovmax/SciLink/issues/429), where the
+answer was "potential source". This writes down what that implies, what already
+exists, and what is missing — so the change can be argued about before it is
+written.
+
+> **Status: implemented in this PR.** `machine_learning_potentials` is removed
+> from the scale axis; a `potential_selection` skill + LLM step (in
+> `potential_selection.py`) chooses classical-FF vs MLIP for a
+> `molecular_dynamics` task once the structure exists, and `simulation_pipeline`
+> dispatches through it. The explicit scale is kept as a deprecation shim. See
+> **Design decisions (resolved)** below. The routing benchmark already labels the
+> four inorganic-dynamics cases `molecular_dynamics` (issue #429), so re-running
+> `test_router` on this branch turns the four documented misses into hits.
+
+## Context — the four scales are not the same kind of thing
+
+`discover_scale_agents()` presents four scales:
+
+| scale | what it names |
+|---|---|
+| `periodic_dft` | a physical regime — extended systems, electronic structure |
+| `molecular_qc` | a physical regime — finite systems, electronic structure |
+| `molecular_dynamics` | a physical regime — classical ensemble sampling |
+| `machine_learning_potentials` | **how forces are evaluated** |
+
+The first three answer "what physics is this?". The fourth answers "where do the
+forces come from?" — a question that only arises *once* you have decided the
+regime is dynamics.
+
+The router's own scale description concedes this. `DEFAULT_SCALE_DESCRIPTIONS
+["machine_learning_potentials"]` reads "best for … systems **without good
+classical force fields**". That is a statement about force-field coverage for a
+chemistry, not about the system's physics — and it is not knowable at routing
+time, before anything has looked at what parameters exist for the species
+involved.
+
+## The evidence — the router applies that criterion, and it is load-bearing
+
+A 26-query routing sweep (`SciLink-Benchmarks/benchmark/test_router.py`,
+claude-opus-4-8, 2026-08-06) scored 22/26. Every one of the four misses went the
+same way:
+
+| query | expected | routed to |
+|---|---|---|
+| thermal conductivity of **amorphous** Si | `molecular_dynamics` | `machine_learning_potentials` |
+| melting point of Cu | `molecular_dynamics` | `machine_learning_potentials` |
+| water density profile at a TiO₂ surface | `molecular_dynamics` | `machine_learning_potentials` |
+| Li diffusion coefficient in LiCoO₂ | `molecular_dynamics` | `machine_learning_potentials` |
+
+The behaviour is coherent, not noisy. MD-sampling tasks on **molecular liquids
+and biomolecules** (water density/self-diffusion, aqueous NaCl conductivity,
+LiPF₆/EC viscosity, solvated glycine, peptide unfolding, benzene hydration free
+energy) all routed to `molecular_dynamics` and all scored correct. MD-sampling
+tasks on **inorganic solids, metals, and interfaces** all routed to
+`machine_learning_potentials`.
+
+That is exactly the documented criterion, applied consistently. **This is not a
+routing bug.** The router is doing what its scale description tells it to. The
+defect is that the decision is being asked at the wrong layer — and because it
+is a scale, answering it commits the whole downstream pipeline.
+
+There is a second, quieter cost. Splitting one regime across two scale labels
+means benchmark ground truth has to pick a side for every dynamics query, and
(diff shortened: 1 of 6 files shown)


sarah-allec:
**Updated from proposal → implementation.** #429 resolved to "potential source"; this implements it.

Design decisions (written up in the doc under *Design decisions (resolved)*):

- a general `potential_selection` skill the LLM reasons over — the skill carries the classical-FF-vs-MLIP knowledge and the model makes the judgment, rather than a hardcoded coverage rule (keeping with the skill-injects-knowledge pattern);
- `simulation_pipeline`'s `molecular_dynamics` branch now selects the potential family once the structure exists and dispatches to the classical-FF or MLIP path (both pre-existing, refactored into helpers);
- `scale="machine_learning_potentials"` is kept as a deprecation shim (→ `molecular_dynamics` + a forced MLIP potential + a warning), so no caller breaks;
- `structure_planner`'s four-way axis and the orchestrator tool-prompt text are deferred — they still work via the shim, and realigning the planner is a separate, non-urgent change.

Verified against this branch that MD-sampling tasks on inorganic solids and interfaces now route to `molecular_dynamics` (their actual regime) and that the downstream FF-vs-MLIP selection behaves as intended.

@ziatdinovmax — ready for your review. Happy to split the deferred `structure_planner` realignment into its own follow-up PR if you'd prefer to keep this one focused.

ziatdinovmax (changes requested):
Thanks for the follow-up. My claude has identified one blocking issue that needs to be addressed before I can merge ->

## Blocking: the FF-vs-MLIP decision moved below the availability gate without carrying availability with it

Pre-PR, software availability constrained this choice at routing time: `candidate_engines()` (`simulation_router.py:193`) intersects each scale's engines with `AvailableSoftware`, so the MLIP path was only ever taken when an MLIP backend was importable, and classical MD only when an MD engine was present. Post-PR, `select_potential_family()` sees only the structure, the goal, and the skill guidance — nothing downstream of routing consults `AvailableSoftware` at all (grep `mlip_agent.py`, `md_simulation_agent.py`, `simulation_pipeline.py`: zero references). That regresses two real environments, and note that `scilink[sim]` installs neither LAMMPS nor an MLIP package, so both are common:

1. **LAMMPS-only, no MLIP stack** (e.g. conda-forge LAMMPS, no `mace-torch`). "Melting point of Cu" — one of the four queries motivating this PR — now routes to `molecular_dynamics` (correct), but the skill's guidance sends metals to MLIP, and MLIP is also the hard fallback when the selection call fails (`_FALLBACK_FAMILY`, `potential_selection.py`), so the pipeline lands in `MLIPAgent.deploy_pretrained` with no MLIP backend installed. Pre-PR this environment ran classical EAM via LAMMPS and worked; the flagship fix converts a working path into a failure here.

2. **MLIP-only, no LAMMPS** (pip `mace-torch` — the ASE-runner audience). With the MLIP scale gone from `discover_scale_agents()`, `molecular_dynamics` only becomes a routing candidate if a classical MD engine is available, and the `molecular_dynamics` skill domain contains only `lammps`. So every dynamics task becomes unroutable for this user — even though the MLIP/ASE path they installed needs no MD engine at all.

The fix is contained and stays within your design:

- In `discover_scale_agents()` / `candidate_engines()`, let `molecular_dynamics` candidacy count available MLIP backends as well as classical MD engines, since the scale can now be served by either family.
- Pass the available families (derived from `AvailableSoftware`) into `select_potential_family()` and constrain both the LLM's choice and the fallback to installed families. When only one family is installed, you can skip the LLM call entirely — the decision is forced.

sarah-allec:
Thanks — that was a real regression, and your fix is exactly right. Addressed in the latest commit; both environments now behave:

- **`candidate_engines()`**: `molecular_dynamics` stays routable when only an MLIP backend is installed, exposing the ASE runner as its engine — so the pip `mace-torch` / ASE-only user can route a dynamics task with no classical MD engine present.
- **`select_potential_family()`** now takes `available_software` (defaults to `AvailableSoftware.auto()`): when only one family is installed the decision is forced with **no LLM call**, and when both are installed the model's pick *and* the fallback (`_FALLBACK_FAMILY`) are clamped to installed families. So a metal MD task in a LAMMPS-only environment selects the classical force field and never reaches `deploy_pretrained` without a backend, and the MLIP-only environment forces the MLIP family.

I verified the three scenarios (both / LAMMPS-only / MLIP-only) resolve as expected and that routing behavior is unchanged. Ready for another look.

ziatdinovmax (approved):
The availability gating resolves the regression — approving. I verified the revision offline (mocked `AvailableSoftware`, no LLM): LAMMPS-only forces classical with no model call, MLIP-only forces MLIP, both-installed falls back to an installed family when the model is unreachable, supplied force-field files still short-circuit, and `candidate_engines()` keeps `molecular_dynamics` routable via `ase` in the MLIP-only environment without leaking `ase` into mixed environments. All seven scenarios behave as your comment describes.

The two non-blocking asks (offline tests for the selection/gating paths, and the EAM nuance in the skill's `planning` section) can land here or as a follow-up — your call. Thanks for the quick and clean turnaround on this.

PR #451, opened 2026-08-10
Fix two LAMMPS validate_script false positives (#409)

### LINKED ISSUE #409: LAMMPS validate_script: two false positives on correct sequential-ensemble decks

sarah-allec:
Closes #409.

`validate_script` reported two spurious errors on physically correct decks. The gate is fail-open so runs still proceed, but the simulate pipeline treats them as validation failures and burns refinement cycles "fixing" a deck that was already right — surfaced live on the UC2 electrolyte runs, where a combined minimize→NPT→unfix→NVT deck tripped both on every member.

### 1. nvt/npt overlap ignored `unfix`

The check accumulated every `fix … npt/nvt` group across the whole file and flagged any group appearing under both, with nothing consuming `unfix`. So a legitimate sequential deck — NPT equilibrate, `unfix`, then NVT produce on the same group — read as if both integrators ran simultaneously.

Now the parse loop tracks the fix lifecycle: an `active_integrators` map (fix-ID → group, family) updated on `fix`/`unfix`, and a conflict is recorded only when an nvt and an npt/nph fix are active on the **same group at the same time**.

### 2. Unresolved-template scan hit comments and valid variables

The regex ran over the entire file, comments included, and rejected `${name}` even when a `variable name …` defined it — so a comment merely *mentioning* `${scale}`, next to a defined `variable scale` used as `v_scale`, was flagged.

Now it scans **code only** (strips `#` comments), exempts a `${name}` that has a `variable` definition in the deck, and still flags a bare `{name}` (not LAMMPS syntax) as an unrendered placeholder.

### Verification
- Both true positives still fire: nvt+npt **simultaneously** active on `all`; `${temperature}` with no definition.
- The real UC2 decks (combined `run.lammps` + the staged `run_equilibration`/`run_production`) now validate with **zero errors**.
- Adds a sequential-ensemble regression fixture + two tests. **100 LAMMPS-skill tests pass.**

### DIFF
--- a/scilink/skills/molecular_dynamics/lammps/lammps.py
+++ b/scilink/skills/molecular_dynamics/lammps/lammps.py
@@ -868,9 +868,13 @@ def validate_script(
     commands_seen: set = set()
     command_order: List[str] = []
 
-    # Track thermostat/barostat per group
-    npt_groups: set = set()
-    nvt_groups: set = set()
+    # Track integrator fixes over their lifecycle (fix … / unfix …) so a
+    # legitimate sequential deck — NPT equilibrate, unfix, then NVT produce on
+    # the same group — is not flagged. `active_integrators` maps fix-ID ->
+    # (group, family); a conflict is recorded only when an nvt and an npt/nph
+    # fix are active on the same group at the same time.
+    active_integrators: Dict[str, tuple] = {}
+    integrator_conflict: set = set()
 
     for line in lines:
         stripped = line.strip()
@@ -899,14 +903,21 @@ def validate_script(
         elif keyword == "run":
             result["has_run"] = True
         elif keyword == "fix" and len(parts) >= 4:
+            fix_id = parts[1]
             group = parts[2]
             style = parts[3].lower()
             if "shake" in style or "rattle" in style:
                 result["has_shake"] = True
-            if style in ("npt", "nph"):
-                npt_groups.add(group)
-            elif style == "nvt":
-                nvt_groups.add(group)
+            if style in ("npt", "nph", "nvt"):
+                family = "nvt" if style == "nvt" else "npt"
+                # Conflict only if a DIFFERENT-family integrator is already
+                # active on this same group (i.e. never unfixed).
+                for a_group, a_family in active_integrators.values():
+                    if a_group == group and a_family != family:
+                        integrator_conflict.add(group)
+                active_integrators[fix_id] = (group, family)
+        elif keyword == "unfix" and len(parts) >= 2:
+            active_integrators.pop(parts[1], None)
 
     errors = result["errors"]
     warnings = result["warnings"]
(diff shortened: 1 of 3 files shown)


ziatdinovmax:
No major blockers identified - merging as is.

ziatdinovmax (approved):

Issue #439, opened 2026-08-08
Bring the surgical-edit and proven-script machinery to the simulation agents


ziatdinovmax:
## How this machinery has worked so far

Over the last week the surgical-edit family shipped across the planning and experimental agents, and the live evidence has been consistently strong:

- **Planning mode — `edit_file` / `rename_file` / PDF-twin invariant (#427, merged).** Mechanical document changes stopped costing whole-document LLM re-emission: a figure swap is one exact-snippet edit, byte-identical elsewhere; forwarded PDFs can no longer silently serve stale content; renames are byte-exact instead of hand-rebuilt through chunked saves (observed live to drop content). 48 live checks across 10 scenarios, including a replay of the meta-session failure that motivated it (an 18-call edit chain reduced to one batched call).
- **Experimental agents — `script_edits` on locked reuse (#434/#435, merged).** "Same analysis, one knob changed" now reruns the prior saved script byte-identical except the requested edit, with execution as the verification and the exact old/new pairs recorded as provenance. Live: an orchestrator conversation where the model read the saved script itself and produced a one-line diff between consecutive campaign points; 36/36 across nine scenarios including meta-mode delegation chains and realtime frames.
- **Experimental agents — bank minimal-edit adaptation (#436/#437/#438).** Proven scripts in the bank now adapt by *edit list* instead of full re-emission, keeping the proven logic byte-recognizable and accumulating cross-session success (proven-N) on the same record. Live highlights: one SAM segmentation script generalized **AFM → OM → TEM** with 1–2 edits each (proven-N 3); an X-ray K-edge script adapted **cross-element Au → Bi** (10 edits: edge energy, element, density) and survived the strict verifier; and on hard cases (AuSputter beam-hardening, TaS₂ zone-to-zone STS) the unchanged verification loop **correctly rejected** transfers the adapter believed were fine — the fall-through ladder means the mode can only add one attempt, never subtract. A degenerate LLM edit reply (`old_text == new_text`) was refused pre-execution by the atomic validator.

The pattern that made all of this safe: **exact-snippet edits applied mechanically (atomic, capped), execution/verification as the unchanged arbiter, loud fall-throughs, and provenance only on clean acceptance.**

## Proposal: three tiers for the simulation side

[...]

Issue #397, opened 2026-07-28
Unify the file-reading layer across orchestrator modes (extraction is shared, windowing is not)


ziatdinovmax:
Every orchestrator mode reads files, each does it differently, and the differences are accidents rather than decisions. Text **extraction** is already shared (`scilink/parsers/extract.py`); **windowing** — how a file too large to return whole is navigated, and how that is reported — is not. So each mode is missing the half another mode has.

**Full plan, written to be actionable from a fresh session:** [`unified_file_reading_plan.md`](../blob/fix-planning-kb-scope-and-tool-args/unified_file_reading_plan.md) on `fix-planning-kb-scope-and-tool-args`. It carries file:line anchors, a reproducer, the measured blast radius and per-phase verification. This issue is the summary.

## Two observed failures

**Planning cannot open a PDF.** A delegation was handed a 29-page funded proposal to fold into a portfolio:

```
read_file     -> '%PDF-1.7\n%<bytes>...1082 0 obj'
extract_text  -> 70,987 chars, 29 pages, clean
```

The agent diagnosed the garbling and proceeded from the caller's summary instead — the run degraded quietly, grounded on a paraphrase rather than the document. Reproducer: `meta_session_20260727_195954/uploads/Narrative_PNNL_Zhang_87627.pdf`.

**Meta and analysis cannot reach the end of one.** Both truncate from the head with no `tail`, no `search`, and a bare `truncated: true` — the same shape that made an agent call planning's `read_file` five times on a 35 KB white paper before giving up (fixed there in #394, not elsewhere).

## Current state

| mode | tool | extraction | windowing |
|---|---|---|---|
| planning | `read_file` (`planning_agents/orchestrator_tools.py:4656`) | text/json/csv only | head + tail + search (lines) |
| meta | `view_document` (`meta_agent/meta_orchestrator_tools.py:1502`) | `extract_text` | head only, 200k chars **per document** |
| analysis | `read_document` (`exp_agents/analysis_orchestrator_tools.py:5500`) | `extract_text` | head only, 200k chars **combined** |
| simulation | — | — | — |

## Proposal

Not one tool — inspecting a session artifact and ingesting source material are different jobs, and collapsing them makes the tool description harder to route on. Share the middle instead:

```
        extract_text()      <- already shared
              |
          window()          <- the missing piece
        /     |      \
read_file  view_document  read_document
```

## Phases

[...]

PR #452, opened 2026-08-10
Simulation orchestrator: edit_file / rename_file (#439 Tier 1)

sarah-allec:
Tier 1 of #439 — brings the surgical-edit machinery to the simulation side.

`edit_file` / `rename_file` wrappers on the simulation orchestrator over the shared `scilink/utils/file_edit.py` core, same shape as the planning adoption (#427). The canonical case is *"change ENCUT to 520 in this INCAR"* without an LLM-mediated regeneration round — `apply_input_adjustments` today is overkill for a single value and can perturb other settings while it's in there.

### What's added
- **`edit_file`** — atomic exact-snippet edits (single `old_text`/`new_text` or a batched `edits` list), 2000-char per-snippet cap, sandboxed to the session directory, pre-edit backup kept. The editable-suffix set extends the shared text defaults with `""` (extensionless VASP `INCAR`/`POSCAR`/`KPOINTS`/`POTCAR`) and the LAMMPS/QE deck extensions (`.lammps`, `.data`, `.in`, `.param`, `.mod`, `.inc`, `.sbatch`, `.pwi`, `.lmp`). The cap-error routing hint points at `apply_input_adjustments`, per the issue.
- **`rename_file`** — byte-exact rename/copy within the file's own directory; content never passes through the model.

All the guards (sandbox, size cap, suffix allowlist, backup) live in the shared core — these are thin wrappers owning path resolution + the sim routing hint.

### Tests
`edit_file`/`rename_file` exercised through `execute_tool`: an extensionless `INCAR` ENCUT edit (+ backup created), a non-text artifact refused, an out-of-sandbox path refused, and a rename. Also completes the stale `EXPECTED_TOOLS` set — `run_simulation` / `run_mlip_simulation` / `run_exafs_workflow` were registered but unlisted, which was already failing `test_1_orchestrator_constructs` on main. **9/9 targeted tests pass.**

Tiers 2 (SimulationAnalysisAgent → shared QC engine, after #405 + a seam review) and 3 (deck bank) are tracked separately in #439.

### DIFF
--- a/scilink/agents/sim_agents/simulation_orchestrator_tools.py
+++ b/scilink/agents/sim_agents/simulation_orchestrator_tools.py
@@ -1416,6 +1416,156 @@ class SimulationOrchestratorTools:
         )
 
         # =====================================================================
+        # 10b. EDIT / RENAME FILE — surgical, byte-exact, no LLM regeneration
+        # =====================================================================
+        # Editable input/deck formats: "" covers extensionless VASP inputs
+        # (INCAR/POSCAR/KPOINTS/POTCAR); the rest are LAMMPS/QE decks and job
+        # scripts. On top of the shared text defaults.
+        _SIM_EDIT_SUFFIXES = {
+            "", ".lammps", ".data", ".in", ".param", ".mod", ".inc",
+            ".sbatch", ".pwi", ".lmp",
+        }
+
+        def edit_file(path: str, old_text: str = None, new_text: str = None,
+                      replace_all: bool = False, edits: list = None) -> str:
+            """Mechanical in-place edit of a generated input/deck file:
+            replace exact text snippets, one old/new pair or a batched `edits`
+            list applied atomically. For retuning several coupled parameters,
+            use apply_input_adjustments instead."""
+            print(f"  ⚡ Tool: Editing file '{path}'...")
+            try:
+                from ...utils.file_edit import (apply_surgical_edits,
+                                                 DEFAULT_EDITABLE_SUFFIXES)
+                if not edits:
+                    if old_text is None or new_text is None:
+                        return json.dumps({"status": "error", "message": (
+                            "Provide old_text+new_text or a non-empty edits "
+                            "list.")})
+                    edits = [{"old_text": old_text, "new_text": new_text,
+                              "replace_all": replace_all}]
+                root = Path(self.orch.base_dir).resolve()
+                rp = Path(path)
+                if not rp.is_absolute():
+                    rp = root / rp
+                rp = rp.resolve()
+                out = apply_surgical_edits(
+                    rp, edits, root=root, backup_dir=rp.parent,
+                    allowed_suffixes=DEFAULT_EDITABLE_SUFFIXES | _SIM_EDIT_SUFFIXES,
+                    too_large_message=(
+                        "Edit too large for edit_file. For a broader input "
+                        "change — retuning several coupled parameters, or "
+                        "regenerating a deck — use apply_input_adjustments. For "
+                        "a verbatim insertion, split the text at a unique "
+                        "boundary into consecutive smaller edit_file calls."),
+                )
+                if out["status"] == "success":
+                    n = out.get("n_edits", 1)
+                    print(f"    ✏️  Edited in place"
+                          f"{f' ({n} edits)' if n > 1 else ''}: {rp.name}")
+                return json.dumps(out)
+            except Exception as e:
+                logging.error(f"edit_file failed: {e}", exc_info=True)
+                return json.dumps({"status": "error", "message": str(e)})
+
+        self._register_tool(
+            func=edit_file,
+            name="edit_file",
+            description=(
+                "Surgically edit a generated input or deck file IN PLACE by "
+                "replacing exact text snippets — one INCAR tag, a timestep, a "
+                "thermostat damping, a pair_style cutoff. The canonical use is "
+                "'change ENCUT to 520 in this INCAR' without regenerating the "
+                "whole input. Copy old_text VERBATIM from the file (read it "
+                "first); each snippet is capped at 2000 characters. Several "
+                "changes to ONE file go in a single call as the `edits` list "
+                "(atomic, in order). Retuning several coupled parameters at "
+                "once, or a change that needs the input rebuilt, is "
+                "apply_input_adjustments' job, not this. A pre-edit backup is "
+                "kept automatically."
+            ),
+            parameters={
+                "path": {"type": "string", "description": (
+                    "Path of the file to edit — absolute, or relative to the "
+                    "session directory. Must be inside the session.")},
+                "old_text": {"type": "string", "description": (
+                    "Single-edit form: the exact snippet to replace, copied "
+                    "VERBATIM including whitespace. Must match exactly one "
+                    "place unless replace_all. Omit when passing `edits`.")},
+                "new_text": {"type": "string", "description": (
+                    "Single-edit form: the replacement text. Omit with "
+                    "`edits`.")},
+                "replace_all": {"type": "boolean", "description": (
+                    "Single-edit form: replace every occurrence instead of "
+                    "requiring a unique match. Default false.")},
+                "edits": {
+                    "type": "array",
+                    "items": {"type": "object", "properties": {
+                        "old_text": {"type": "string"},
+                        "new_text": {"type": "string"},
+                        "replace_all": {"type": "boolean"}},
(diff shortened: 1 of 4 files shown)


ziatdinovmax (changes requested):
This is great - thanks for following up on this so quickly! There's one blocker that needs addressing before merging - see Claude's comments below.

Also, in my experience, `edit_file` works best when paired with `read_file`. But since #397 already outlines a general reading layer across all three modes, I suggest we wait for that to land and use it for sim agents, rather than introducing a simulate-specific implementation now that we'd have to consolidate later.

## BLOCKER — `""` in the suffix allowlist lets `edit_file` corrupt binary VASP files

`simulation_orchestrator_tools.py:1424`

The empty suffix was added so extensionless VASP decks (`INCAR`, `POSCAR`, `KPOINTS`) are editable. Downloaded run outputs are extensionless too. The shared core gates only on suffix, then does `read_text(errors="replace")` + `write_text`.

Confirmed live on a session file `structures/mp/outputs/WAVECAR` seeded with `b'\x00\x01\x02ENCUT = 400\xff\xfe binary tail'`: the tool returns `status: success` and the file becomes `...\xef\xbf\xbd\xef\xbf\xbd...` — every non-UTF-8 byte replaced by U+FFFD. Worse, the pre-edit backup is written with `bak.write_text(current)` from that same lossy read, so the original bytes are unrecoverable.

`WAVECAR`, `CHGCAR`, `CHG`, `WAVEDER` are all extensionless and land inside the sandbox via the download path. There is no size guard either — a multi-GB `WAVECAR` is read whole into memory.

**Fix.** A binary sniff (`b"\x00"` in the first few KB) and/or a byte-size cap before the read. The new test's "non-text artifact is refused" case uses `.png`, which the suffix gate already catches — it does not exercise the guard this PR weakens.

## Two smaller asks, while the reader is deferred

Both are about not instructing the model to use a tool simulate doesn't have.

1. **`edit_file` description** (`simulation_orchestrator_tools.py`) — drop "(read it first)" from *"Copy old_text VERBATIM from the file (read it first)"*. There is no reader on this orchestrator, and `generate_dft_inputs` returns `{input_files: <paths>, summary}`, so the instruction is unsatisfiable as written.

2. **Parameterize the not-found hint in the shared core.** `apply_snippet_edits` hardcodes `"old_text not found — read_file the document and copy the snippet verbatim"` at `file_edit.py:93`, while the sibling `too_large_message` is a parameter threaded from `apply_surgical_edits` (`:45`, `:136`, `:151-152`) precisely so each mode names its own alternative. Same asymmetry — add e.g. `not_found_message` with today's string as the default so planning stays byte-identical, and have the simulate wrapper pass one that names a surface simulate has: for `periodic_dft`, `apply_input_adjustments` is the LLM-mediated path that retunes tags *without* needing an exact snippet match, which is the real recovery route when a verbatim guess misses.

sarah-allec:
Thanks — all four addressed in `8046ae6e`.

**The blocker.** Guards go in the shared core ahead of the read, since that's where the others live:

- **byte cap** (16 MB, via `stat()`) so an oversized file never enters memory. This is also the *only* guard that stops `CHGCAR` — it's ASCII, so it passes a binary sniff.
- **NUL byte** in the first 8 KB.
- **strict UTF-8 decode**, replacing `errors="replace"`. This catches the no-NUL-but-not-text case (the `\xff\xfe` tail in your repro).

Two more the blocker text implicated directly: the backup is now `shutil.copyfile` rather than `write_text(current)` — it's the only route back, so it has to be the file's bytes, not a re-encoding of what we decoded from them — and the write pins `encoding="utf-8"` to match the strict decode.

Your repro now returns `status: error`, leaves the bytes untouched, and writes no backup.

**`read_file`** — agreed, nothing simulate-specific added; deferred to #397.

**The description** — `(read it first)` dropped.

**`not_found_message`** — added alongside `too_large_message` with today's string as the default, so planning stays byte-identical. Simulate passes one naming `apply_input_adjustments`, per your point that it retunes tags without needing an exact match.

**Tests.** The `.png` case stays but is relabelled — it pins the suffix gate, not the content guard, which was a fair hit. New coverage: your repro at both the core and `execute_tool` levels, undecodable-but-no-NUL, the byte cap in both directions, backup byte-exactness, and the new parameter. 15/15 core, 17/17 planning `edit_file` (unchanged behavior), 21/21 rename/copy, 9/9 targeted sim orchestrator.

One unrelated thing worth knowing: `test_rename_copy_delegation_integrity.py` and `test_rename_copy_across_delegations.py` only collect under `--import-mode=importlib` — they do `from tests.test_edit_file_tool import make_tools` and `tests/` has no `__init__.py`. Pre-existing, not from this PR, but it'll bite anything running plain `pytest tests/`.

sarah-allec:
Working on resolving the conflicts - having some OneDrive issues so it's just taking some time 😅

sarah-allec:
Merged `upstream/main` in (`48e1c1cf`) to clear the conflict — worth a note, since the resolution looks like it drops one of the 188 fixes from the UTF-8 sweep (70f070e2).

Both sides touched the same two writes in `utils/file_edit.py`:

```
rp.write_text(res["text"], encoding="utf-8")   both sides identical — no real conflict
bak.write_text(current, encoding="utf-8")      resolved to shutil.copyfile
```

The second one **supersedes** the sweep's fix rather than reverting it. The sweep made the pre-edit backup's write encoding-explicit; this branch replaced that write with `shutil.copyfile`, because the backup is the only route back from a bad edit and so has to be the file's bytes — a backup written through any decode/encode round-trip can't restore bytes that didn't survive the decode, which was the second half of the blocker. Copying never decodes, so there's no encoding left to state.

I've left a comment at the call site saying exactly that, so the next encoding audit doesn't see an unstated-encoding backup path and helpfully restore the write.

`tests/test_utf8_process_mode.py` passes — the one remaining `write_text` in the file states utf-8, and `copyfile` isn't a text write. `simulation_orchestrator_tools.py` auto-merged despite both sides touching it.

Full run after the merge: 42 passed (core + planning `edit_file` + utf8), 21 rename/copy, 9/9 targeted simulation orchestrator.

ziatdinovmax:
I pushed a small follow-on (047d5c76) directly to your branch. 

The strict decode in the binary guard dropped the newline normalization read_text() used to do. So \n-authored snippets could never match CRLF (Windows-written) files and a successful edit on Windows would corrupt them to \r\r\n. The commit normalizes CRLF for matching, writes the file's own newline convention back untranslated, and pins it with tests.

ziatdinovmax (approved):
Approving — the binary-corruption blocker and both smaller asks are all addressed, and the CRLF follow-on commit rounds out the guard work. A deeper second pass over the branch turned up the items below. None of them block this PR: they're rough edges on the new surface, recorded here so they don't get lost. We'll spin them into issues as we get to them.

**Session state after file operations** (the tools work, but the orchestrator's records don't follow them)
- `rename_file` doesn't update the generated-structures records, so renaming a registered POSCAR/INCAR orphans the stored path — HPC submit then reports "No local structure file to upload" and validate/refine lose the structure's provenance.
- Editing a structure's `script_*.py` doesn't invalidate the record's cached `script_content` (refine can silently revert the edit), and the `.before_edit.py` backup lands in the `script_*.py` glob's candidate set.
- The one-per-file `*.before_edit*` backup goes stale across the regenerate-in-place loop (edit INCAR → regenerate → edit again reports the old generation's deck as the rollback point).

**Guard gaps**
- A small ASCII run output (few-atom CHGCAR/LOCPOT, DOSCAR, OUTCAR — extensionless, no NULs, under 16 MB) still passes every guard; a filename blocklist for known VASP outputs or an `outputs/` exclusion would close it.
- The NUL sniff only checks the first 8 KB; since the whole file is already in memory, `b"\x00" in raw` over the full buffer is free and catches text-header/binary-tail files.

**Routing and ergonomics**
- `_SIM_EDIT_SUFFIXES` omits `.sh`, but the orchestrator itself writes `job.sh` — "bump the walltime" gets refused and pushes the model toward regeneration.
- The byte-cap message hard-codes sim wording ("a file this size is a run output") in the shared core; worth parameterizing like `too_large_message` / `not_found_message` so planning isn't told sim things.
- The UTF-8 refusal could route through `utils/text_io.py`'s decode ladder (on the branch since the merge) or at least name a next step, instead of dead-ending on legacy cp1252 files.
- `rename_file` silently strips directory components from `new_name` (`outputs/INCAR.bak` becomes an in-place rename); an error or a note when `Path(new_name).name != new_name` would keep the model's bookkeeping honest.

PR #456, opened 2026-08-11
Move ion parameter guidance into the Planning section, and cover trivalent

sarah-allec:
The AMBER skill's ion guidance never reaches the decision it is meant to inform.

`_select_optimal_force_field()` injects `_get_skill_context(section="planning")`, and the ion bullets live under `## Overview`. Loading the skill shows it directly:

```
overview   1487 chars | Joung=True  Merz=True  Aqvist=True
planning   3608 chars | Joung=False Merz=False Aqvist=False
```

`_get_skill_context` returns the requested section or `""` — there's no fallback to overview. Ion selection is also the only *selection* decision filed under Overview: water-model and charge-method selection are already in Planning, each with a `Decision rule:`, and the agent handles both reliably (it picks SPC/E correctly every time).

## Measurement

Aqueous Ca²⁺, Zn²⁺ and Al³⁺, coordination-number goal, 3 runs each, `claude-opus-4-8`. Counting whether the answer names a real ion parameter set:

| arm | names a set | recommends 12-6-4 |
|---|---|---|
| no skill loaded | 1/9 | — |
| skill loaded, guidance in Overview | 4/9 | **0/9** |
| skill loaded, guidance in Planning (this PR) | **8/9** | **4/9** |

Loading the skill helps somewhat; moving the guidance into the injected section is what actually lands it, and is the only arm that produces a 12-6-4 recommendation at all.

Without it, answers are generic for systems where coordination number is the requested observable — "non-bonded ion model", "12-6 point-charge", "monovalent/divalent ion LJ parameters".

## Changes

Docs only, one file, 23 insertions.

- Ion guidance moves to Planning as its own subsection, in the same table + `Decision rule:` style as its neighbours. Overview keeps a one-line pointer.
- **Trivalent ions added.** The old text covered "Mg2+, Ca2+, Zn2+" only. Al³⁺ went 0/3 → 3/3, all three naming 12-6-4.
- The rule distinguishes when 12-6-4 matters — coordination number, hydration structure, binding free energies, where a 12-6 set can be fitted to the hydration free energy *or* the ion–oxygen distance but not both — from when plain 12-6 is fine (spectator counter-ions).

n=9 per arm is small, but the effect is large and the mechanism was verified directly rather than inferred.

## Separate, not addressed here

`_auto_select_skill()` runs *after* `_select_optimal_force_field()`, and 4 of 6 construction sites — including `simulation_pipeline.py:484` — never pass `skill=`. So on those paths no skill is loaded at all when the force field is chosen. That's a code change and a different discussion; happy to open it separately if useful.

Found while building force-field coverage for the benchmark suite, which had none.

### DIFF
--- a/scilink/skills/force_field/amber/amber.md
+++ b/scilink/skills/force_field/amber/amber.md
@@ -78,14 +78,23 @@ water model, and the right set depends on the ion's CHARGE.
 | Trivalent (Al3+, Fe3+, lanthanides) | Li-Merz **12-6-4** | 12-6 is worst here; charge-induced polarisation dominates |
 
 Decision rule: For monovalent ions use Joung-Cheatham matched to the water
-model. For divalent and trivalent cations use the Li-Merz **12-6-4** set
-whenever the research goal involves coordination number, hydration structure,
-ion-water distances, or binding free energies — the r^-4 term represents
-charge-induced dipole and is what lets one parameter set reproduce hydration
-free energy and coordination geometry simultaneously. A plain 12-6 set is
-acceptable only when the ion is a spectator (bulk neutralising counter-ions).
-Never leave a multivalent cation on a generic point-charge Lennard-Jones
-model when its coordination is what is being measured.
+model. For divalent and trivalent cations the Li-Merz **12-6-4** set is the
+physically right answer whenever the research goal involves coordination
+number, hydration structure, ion-water distances, or binding free energies —
+the r^-4 term represents charge-induced dipole and is what lets one parameter
+set reproduce hydration free energy and coordination geometry simultaneously.
+A plain 12-6 set is fine when the ion is a spectator (bulk neutralising
+counter-ions).
+
+**12-6-4 needs a C4 term this pipeline does not currently carry.** The AMBER
+to LAMMPS route here is `lj/charmm/coul/long` / `lj/cut/coul/long`, both 12-6
+forms, and the ParmEd conversion does not transport `LENNARD_JONES_CCOEF`.
+Selecting 12-6-4 where it cannot be implemented is worse than not selecting
+it: the run proceeds on plain 12-6 while carrying parameters fitted on the
+assumption that C4 is present. So when the workflow cannot implement C4,
+choose the 12-6 set fitted to the ION-OXYGEN DISTANCE rather than the
+HFE-fitted one — the goals above are structural — and record both the
+substitution and its consequence for the target observable in `cautions`.
 
 ### Charge method selection for small molecules
 
--- a/scilink/skills/force_field/amber/amber.md
+++ b/scilink/skills/force_field/amber/amber.md
@@ -19,8 +19,8 @@ convert to a LAMMPS data file that embeds all coefficients and charges.
 - Carbohydrates / glycans: GLYCAM06j-1
 - Small organic molecules, ligands, solvents: GAFF2 (preferred) or GAFF
 - Solvated systems with explicit water: TIP3P, SPC/E, OPC, TIP4P-Ew
-- Monovalent ions (Na+, K+, Cl-): Joung-Cheatham (auto-loaded with water model)
-- Divalent ions (Mg2+, Ca2+, Zn2+): 12-6 LJ sets from Aqvist or Li-Merz
+- Ions, monovalent through trivalent — see "Ion parameter selection" under
+  Planning
 
 ### When NOT to use AMBER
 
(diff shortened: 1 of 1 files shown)


ziatdinovmax (changes requested):
Thanks — the diagnosis checks out end to end. I verified the injection path (`_select_optimal_force_field()` injects only the `planning` section, `force_field_agent.py:1578` — the overview fallback at 1579–1580 only fires when planning is empty, which amber's isn't), and the `_auto_select_skill()` ordering note is accurate as well; please do open that one separately.

One change requested before merge: the new decision rule mandates 12-6-4 in absolute terms, but both the agent's baseline prompt and the downstream pipeline currently point the other way.

The baseline selection prompt says, verbatim (`force_field_agent.py:1610-1612`):

> For metal ions, consider: simple point charge models (Joung-Cheatham, Aqvist) which work with standard LJ, rather than 12-6-4 models requiring special implementation

while the new skill text says:

> Never leave a multivalent cation on a generic point-charge Lennard-Jones model when its coordination is what is being measured.

Both land in the same prompt, in direct conflict — which is likely why even the improved arm produces a 12-6-4 recommendation only 4/9 times. The deeper problem is that nothing downstream implements 12-6-4: the skill's own Implementation section documents only `lj/charmm/coul/long` / `lj/cut/coul/long`, and the ParmEd conversion path doesn't carry the C4 (`LENNARD_JONES_CCOEF`) term. So when the agent does select 12-6-4, the pipeline either fails at parameterization or silently runs plain 12-6 with 12-6-4-fitted LJ parameters — arguably worse than a properly chosen 12-6 set, since those parameters assume the C4 term is present.

The science in the rule is right (a 12-6 set fits the hydration free energy or the ion–oxygen distance, not both) — the ask is not to weaken it but to make it implementation-aware. Concretely, in the `Decision rule:` paragraph:

- Keep the 12-6-4 recommendation for coordination-number / hydration-structure / binding goals, but add something like: "12-6-4 requires a C4 term the standard `lj/charmm/coul/long` pipeline does not carry; if the workflow cannot implement it, select the 12-6 set fitted to the ion–oxygen distance (not the HFE-fitted set) and state the limitation in `cautions`."
- Soften "Never leave…" so it routes the concern through `cautions` rather than mandating a selection the pipeline can't execute.

Removing the baseline prompt's anti-12-6-4 clause and adding real C4 support to the conversion path is the better long-term fix, but that's a code change and belongs with the `_auto_select_skill` ordering issue, not this docs PR.

sarah-allec:
Revised in `3267a716`, and re-measured.

**The rule is now implementation-aware.** The science is unchanged — 12-6-4 is still the right answer for coordination / hydration-structure / binding goals, and the HFE-vs-IOD tradeoff still stands. What's added is the gap and a defined fallback:

> **12-6-4 needs a C4 term this pipeline does not currently carry.** [...] Selecting 12-6-4 where it cannot be implemented is worse than not selecting it: the run proceeds on plain 12-6 while carrying parameters fitted on the assumption that C4 is present. So when the workflow cannot implement C4, choose the 12-6 set fitted to the ION-OXYGEN DISTANCE rather than the HFE-fitted one — the goals above are structural — and record both the substitution and its consequence for the target observable in `cautions`.

"Never leave a multivalent cation on a generic point-charge model" is dropped, per your note that it mandates a selection the pipeline cannot execute.

I verified your two claims rather than taking them on trust: grepping the package for `C4` / `CCOEF` / `12-6-4` finds only prompt and skill *prose* — no implementation anywhere — and `amber.md` documents only `lj/charmm/coul/long` / `lj/cut/coul/long`, both 12-6 forms.

### Re-measurement — same protocol, 3 ions × 3 trials, `claude-opus-4-8-project`

| | before | after |
|---|---|---|
| names a real parameter set | 8/9 | **9/9** |
| selects 12-6-4 | 4/9 | **0/9** |
| selects the IOD-fitted 12-6 fallback | — | **9/9** |
| states the C4 limitation in `cautions` | — | **9/9** |

The 12-6-4 number going to zero is the intended effect, not a regression: the agent now recognises it cannot be implemented and substitutes deliberately. Representative `ion_parameters`: *"Li-Merz 12-6 set fitted to the ion-oxygen distance (IOD) for Ca2+, matched to SPC/E"*, with `cautions` reading *"CRITICAL: The optimal parameter set for a Ca2+ hydration-structure study is Li-Merz 12-6-4, but this pipeline's AMBER->LAMMPS conversion cannot carry the C4 term. We substitute the 12-6 Li-Merz set fitted to the ion-oxygen distance (IOD) rather than the HFE-fitted set..."* `lammps_pair_style` is a 12-6 form in 9/9, consistent with the substitution.

**Two caveats on the numbers, since this PR lives or dies on its table.**

First, my initial scoring pass reported *"recommends 12-6-4: 9/9"* — wrong. The regex matched `12-6-4` anywhere in the response, so it was counting the cautions text explaining why 12-6-4 *can't* be used as though it were a selection of it. A check that cannot fail, exactly the failure mode we keep hitting. Corrected to score the `ion_parameters` slot; the honest number is 0/9.

Second, the original probe script was never committed, so this is a reconstruction from the protocol in `ab_ions_after.log` (same ions, trial count and goal shape; system held constant across ions). `named` is scored on the `ion_parameters` field specifically rather than the force-field name string the old log printed, which is stricter — so the 8/9 → 9/9 improvement is real but measured slightly differently at each end. n=9 remains small.

Skill still loads with all six sections and empty `extras`; ion guidance sits in `planning` at 5471 chars.

The `_auto_select_skill()` ordering issue is next on my list — I'll open it separately as you asked.

ziatdinovmax (approved):

PR #459, opened 2026-08-13
Added MACE-POLAR-1 info to MACE skill

jenna1701:
I updated the MACE skill to add the MACE-POLAR-1 (PolarMACE) family.

  - Frontmatter — description now mentions the polar family; detect.guidance documents the extra runtime dependency (mace-torch>=0.3.16 plus graph_electrostatics providing graph_longrange), with the git install line and an import check.
  - Overview — added a MACE-POLAR-1 row to the model table and a dedicated MACE-POLAR-1 subsection covering the architecture (local backbone + non-self-consistent polarizable-field update, Fukui equilibration, arbitrary charge/spin, external-field response), the OMol25/ωB97M-V training, the polar-1-s/m/l checkpoints with receptive-field sizes, and the reported accuracy gains.
  - Planning — new model-selection rules (charged/ionic, redox, applied field, long-range electrostatics, or per-atom charge/dipole targets → polar-1-m/polar-1-l), the polar-1-l vs polar-1-m tradeoff, and a fine-tuning pointer.
  - Implementation — the mace_polar ASE loader path (setting charge/spin/external_field in atoms.info), reading dipole / density_coefficients / spin_charge_density, and the not-uniquely-defined/isolated-fragment caveat.
  - Interpretation — polar-specific output caveats (PBC dipole is ill-defined, partial charges are interpretive, applied-field sanity checks).

Verified it loads through the skill loader with all sections intact.

### DIFF
--- a/scilink/skills/machine_learning_potentials/mace/mace.md
+++ b/scilink/skills/machine_learning_potentials/mace/mace.md
@@ -1,5 +1,5 @@
 ---
-description: MACE — equivariant message-passing neural network potential with production-ready foundation models for inorganic and organic systems; supports both ASE and LAMMPS pair_style.
+description: MACE — equivariant message-passing neural network potential with production-ready foundation models for inorganic and organic systems; supports both ASE and LAMMPS pair_style. Includes the MACE-POLAR-1 family with explicit long-range electrostatics for charged/polar systems and external fields.
 detect:
   binaries: []
   env_vars: []
@@ -10,6 +10,13 @@ detect:
     runs, also check for the `lammps-mace` plugin (`pip install lammps-mace`)
     or a LAMMPS build with the ML-IAP package linked against libtorch
     (`lmp -help | grep mace`).
+
+    The MACE-POLAR-1 (PolarMACE) electrostatic models need `mace-torch>=0.3.16`
+    PLUS the `graph_electrostatics` package, which provides the
+    `graph_longrange` module required at runtime:
+    `pip install git+https://github.com/WillBaldwin0/graph_electrostatics.git@v0.4.0`.
+    A successful `from mace.calculators import mace_polar` plus `import
+    graph_longrange` confirms the polar backend is usable.
 ---
 # MACE (Multi-ACE)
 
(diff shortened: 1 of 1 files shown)


sarah-allec:
Thanks for this, @jenna1701! This is a well-structured addition and it loads cleanly through the skill loader (all six sections intact, nothing dropped to `extras`). The physics caveats (PBC dipole, interpretive partial charges, isolated-fragment rule) are exactly right.

One integration boundary worth stating in the skill before it merges, because otherwise it advertises something the orchestrated pipeline can't yet deliver silently. polar-1's capability lives in the per-frame `atoms.info["charge"]` / `["spin"]` / `["external_field"]` your Implementation snippet sets. That direct ASE / single-point path is fine. But when a model is chosen through the agent (`MLIPAgent → DeployedPotential → MDSimulationAgent`), the generic ASE runner reads coordinates and attaches the calculator and never sets `atoms.info` — so through the MD seam polar-1 would run at neutral / no-field defaults, and for a charged or redox system that's a silent wrong result rather than an error. I've opened #463 to add a `system_state` carrier to `DeployedPotential` so the seam can actually pass charge/spin/field; until that lands, a note keeps the skill honest.

Suggested paragraph — drop it into the Implementation section right after the `mace_polar` block:

> **Scope through SciLink's MD pipeline.** The `atoms.info` charge/spin/field
> inputs above apply to the direct ASE path (single-point evaluation, or an
> ASE script you drive yourself). The orchestrated MLIP→MD path
> (`MLIPAgent`/`MDSimulationAgent`) does not yet propagate per-frame system
> state, so a polar-1 model selected through it runs at **neutral charge,
> default spin, no field** — i.e. as a standard MACE run. Until SciLink#463
> adds a system-state carrier to `DeployedPotential`, use polar-1 for
> single-point charge/dipole analysis rather than charged/field/redox MD
> through the agent.

And optionally a half-line in the Planning selection rule — something like "(single-point/ASE path today; MD-through-the-agent for charged/field systems is pending SciLink#463)" next to the polar-1 selection bullet — so the model-selection step itself carries the caveat, not just the implementation notes.

Everything else looks good to me.

jenna1701:
Added the two caveats. Will SciLink#463 also remove these caveats once resolved?

sarah-allec (approved):
Yes, #463 is exactly that fix: the ASE runner currently attaches the calculator but never sets `atoms.info[...]`, so the seam drops the charge/spin/field state. Once #463 lets `DeployedPotential` carry per-frame state to the runner, these caveats stop being true and come out - they're written as temporary against #463 on purpose. Worth noting #463 gets more load-bearing after #432 (MLIP-as-potential-source): charged/polar systems, lacking good classical FFs, will route to an MLIP more often, so more real cases hit this gap. 

Anyway, approving — thanks for the clean caveats!

PR #460, opened 2026-08-14
Human-in-the-loop as first-class state: chokepoint, channels, durable log, MCP answers

ziatdinovmax:
## Summary

Today, whenever a SciLink agent wants your opinion mid-run — approve a fitting plan, pick a best-of-N candidate, review extracted metrics — it stops and waits for you to type into the terminal. That works at the keyboard, but the question itself is invisible everywhere else: parallel fan-out branches must silence their questions entirely, nothing records what was asked or answered, and an external agent driving SciLink over MCP can never see or answer a question at all.

This PR makes the question itself a first-class object. Every human-feedback prompt now flows through one chokepoint that knows what kind of decision is being asked, who is asking, and what "accept as-is" means. From there:

- **The console and UI behave exactly as before** — same prompts, same buttons.
- **Parallel branches can now pause for your approval** (opt-in): their questions queue up and are presented one at a time, labeled by branch, instead of being suppressed.
- **Every session keeps a record** (`feedback_log.jsonl`) of what was asked, what you answered, and how long you took — and if a run dies while waiting, a sidecar file says what it was waiting for.
- **Remote agents can answer questions over MCP**: a background job pauses with `awaiting_input`, the client sees the structured question, answers via `scilink_respond`, and the run resumes. This is the LangGraph-style interrupt/resume experience, without adopting a graph runtime — and it fixes three existing MCP bugs along the way (a second question silently overwriting the first, free-text feedback discarded as a cancel, and the autonomy switch not reaching the live agents).
- **You can watch SciLink work in real time from any client**: while a background job runs, `scilink_job_status` now returns a `log_tail` of the orchestrator's recent narration (tool calls, fit progress, verifier verdicts) — so a remote agent or dashboard sees live progress through the protocol itself, not just at completion.

Each phase was live-validated end-to-end (co-pilot curve fit; per-branch fan-out approvals; durable log; full co-pilot-over-MCP loop with a real `scilink serve` child), and all defaults are behavior-preserving: console byte-identical, fan-out branches autonomous unless opted in, autonomous MCP serving unchanged.

---

<details>

### Commit 1 — chokepoint (`scilink/hitl.py`)
`request_human_feedback(prompt, kind, options, default, context, origin)` builds a `FeedbackRequest` and delegates to the active channel (thread-local override, else process default). The default `ConsoleChannel` calls `builtins.input` **at call time**, so console behavior is byte-identical and front-ends that monkeypatch `builtins.input` kept working throughout the migration. All 22 gated `input()` sites converted (curve/image controllers, base controller, tier-2 escalation, planning `user_interface`, scalarizer, feedback mixins, fan-out confirm). CLI bootstrap/REPL prompts, destructive-op confirms, and the TTY-gated sandbox consent intentionally stay raw. Dead `ask_user_proceed_or_refine` removed. A tokenizer-based test gate asserts no raw `input()` outside the allowed files (skips f-string text on py3.12).

### Commit 2 — channels
- `_UIChannel` (`ui/app.py`): installed thread-locally for the chat thread; publishes structured requests to `task.feedback_request` (`ui/state.py` gains additive `kind`/`options`/`origin` fields). Widget dispatch gets kind-based fast paths with the existing prompt/context sniffing retained as fallback; the `builtins.input` monkeypatch stays as a raw-input safety net for third-party code and worker threads, delegating to the same channel. Metadata auto-reply cache keys on `origin['filename']` with the legacy regex fallback.
- `QueueChannel` (`hitl.py`): workers park requests; the coordinator thread serves them serially through its own channel, prompts prefixed `[branch: <label>]`. Fan-out wires it behind `orch.fanout_branch_hitl` (**default off**): when enabled with a human attached, branches keep the meta's autonomy instead of hardcoded `AUTONOMOUS` (`fanout.py:587/:746` paths), and the coordinator's wait loop drains the queue each poll.

### Commit 3 — durable record
The chokepoint appends to `<session>/feedback_log.jsonl` at ask time and at answer/interrupt time, and maintains a `pending_question.json` sidecar while blocked. Binding is thread-local: each orchestrator binds its session file at `chat()` entry; `run_task` restores the caller's binding in `finally`, so a delegating meta logs to its own session while children (and fan-out branches) log to theirs. Recording is best-effort and can never break a prompt; unbound threads are a no-op.

### Commit 4 — MCP
- `state["pending"]` is a dict keyed by request id, holding both tool approvals and parked agent questions — queue, not slot.
- `_MCPChannel` is installed around every tool / orchestrate execution; questions block the worker with a mandatory timeout (`--hitl-timeout`, default 1800 s) falling back to the question's declared default.
- `scilink_respond` gains optional `request_id` (not needed when exactly one request is pending). Agent questions receive the response **verbatim** ('' = accept); the approve-keyword gate applies only to tool approvals, and a cancel echoes the feedback with an explicit note.
- `scilink_job_status` reports `awaiting_input` with the structured question(s) for a blocked job.
- `scilink_set_autonomy` now also calls the live orchestrators' mode setters (enum lookup by *name*, since `AutonomyLevel` and `AnalysisMode` disagree on the co-pilot value string).
- `orchestrate_*` routes through `run_task` (per-call autonomy, structured `status`/`files_produced`/`key_findings` contract) with the legacy chat text preserved in a `response` field — the one client-visible shape change.

### Commit 5 — real-time job narration (`log_tail`)
Each background job carries a bounded narration buffer fed by a thread-scoped `logging` handler (keyed via `log_context.effective_thread`, so concurrent jobs and their worker threads stay separated) plus a tee on the captured stdout. `scilink_job_status` returns the last N lines as `log_tail` (`tail_lines` param, default 25, max 200, `0` to omit). Additive — jobs without a buffer and existing clients are unaffected. Live-validated: 97 unique narration lines streamed over 26 in-flight polls of a running co-pilot job.

### Testing
- Offline: 36 tests across `test_hitl_chokepoint.py`, `test_hitl_channels.py`, `test_hitl_durability.py`, `test_mcp_hitl.py`; plus scalarizer-passthrough 14/14, analyze_file shapes 9/9, candidate-pool 4/4, fan-out/fusion suites. `test_fusion_incremental.py` fails 25/26 identically on the base commit (pre-existing).
- Live (Gemini): co-pilot curve fit through the chokepoint (R²=0.9915); real UI chat path driven headlessly with structured requests; 2-branch fan-out with per-branch approvals served serially; durable log + sidecar verified in a real session; full co-pilot-over-MCP loop against a real stdio `scilink serve` child (`awaiting_input` → `scilink_respond` → completed structured result).

- Stress campaign (all passed): mixed pending types + free-text corrections that changed the fit (2→3 Gaussians as instructed); unattended-timeout fallback (exactly at the configured deadline); live `set_autonomy` switch; Stop during a pending prompt (clean interrupt, logged); console revision loop; UI fan-out with labeled per-branch approvals; and every orchestrator — meta sequential delegation (per-session logs correct), planning in-agent prompts (Bedrock), simulation no-regression, and a meta-mediated analysis→BO hand-off ending in one connected report.
- Real deepagents/LangChain stack: the autonomous closed-loop BO campaign reproduces the pre-change protocol exactly (11 tool calls, zero `needs_input`, 61.1→87.5%), and in autopilot the deep agent handled the `needs_input` → `scilink_respond` approval cycles unprompted.
- Known pre-existing limitations surfaced while testing (not introduced here): the planning orchestrator's *chat* tool schemas are rejected by Gemini's function-declaration validator (drive planning with Claude models, as the demos do), and the sim stack's Gemini path expects Vertex credentials.

### Not in this PR
Restart durability (plan Phase 4 — coupled to mid-fan-out resume) and the MCP elicitation fast path (client adapters lack support; the polling contract is the baseline). Plan doc: `hitl_durable_feedback_plan.md`.

</details>

### DIFF
--- a/scilink/agents/exp_agents/analysis_orchestrator.py
+++ b/scilink/agents/exp_agents/analysis_orchestrator.py
@@ -1349,6 +1349,12 @@ class AnalysisOrchestratorAgent:
 
     def chat(self, user_input: str) -> str:
         """Main chat interface with robust function calling support."""
+        # Bind this turn's human-feedback prompts to the session's
+        # durable feedback log (nested child chats rebind to their own;
+        # run_task restores the caller's binding on exit).
+        from ...hitl import set_thread_feedback_log as _set_fblog
+        from pathlib import Path as _P
+        _set_fblog(str(_P(self.base_dir) / "feedback_log.jsonl"))
         self.message_count += 1
         self._last_chat_hit_iter_cap = False
 
@@ -1436,6 +1442,8 @@ class AnalysisOrchestratorAgent:
         run_mode = autonomy if autonomy is not None else AnalysisMode.AUTONOMOUS
         original_mode = self.analysis_mode
         original_max_iter = self.max_iterations
+        from ...hitl import get_thread_feedback_log, set_thread_feedback_log
+        _prev_fblog = get_thread_feedback_log()
         try:
             self.set_analysis_mode(run_mode)
             if max_iterations is not None:
@@ -1462,6 +1470,9 @@ class AnalysisOrchestratorAgent:
             # Always restore the original mode, even if chat() raised.
             self.set_analysis_mode(original_mode)
             self.max_iterations = original_max_iter
+            # chat() bound the feedback log to THIS session; restore the
+            # caller's binding (a delegating meta keeps logging to its own).
+            set_thread_feedback_log(_prev_fblog)
 
         # Derive the structured summary from the session-state delta.
         new_analyses = self.analysis_results[n_before:]
--- a/scilink/agents/exp_agents/controllers/base_controllers.py
+++ b/scilink/agents/exp_agents/controllers/base_controllers.py
@@ -4,6 +4,8 @@ import textwrap
 from typing import Any, Callable, Optional
 import json
 
+from ....hitl import request_human_feedback
+
 
 # Generic Tier-A synthesis re-entry instructions (issue #322). Deliberately
 # modality-neutral: the payload critique + surfaced features carry the
@@ -301,7 +303,11 @@ class IterativeFeedbackController:
         print("-" * 80)
         
         try:
-            user_feedback = input("\n🤔 Your feedback to adjust the targets/plan (or press Enter to accept): ").strip()
+            user_feedback = request_human_feedback(
+                "\n🤔 Your feedback to adjust the targets/plan (or press Enter to accept): ",
+                kind="review_plan",
+                origin={"stage": "preprocess_plan"},
+            ).strip()
         except KeyboardInterrupt:
             self.logger.warning("User interrupted feedback. Accepting original decision.")
             return state
--- a/scilink/agents/exp_agents/controllers/curve_fitting_controllers.py
+++ b/scilink/agents/exp_agents/controllers/curve_fitting_controllers.py
@@ -40,6 +40,7 @@ from .._locked_exec import (
 from .._qc_engine import CodegenQCEngine, QCEngineSpec, QCItemContext
 from ....utils.codegen_parse import parse_codegen_response
 from ....utils.synthesis_parse import salvage_synthesis_from_response
+from ....hitl import request_human_feedback
 
 # Canonical fitted-curve output the fit script saves alongside visualization.png.
 # Best-effort: when present it powers controller-side residual diagnostics; when
@@ -1919,7 +1920,11 @@ class CurveFittingPlanningController:
 
     def _get_human_feedback(self, state: dict) -> dict:
         self._display_plan(state)
-        feedback = input("\n🤔 Your feedback (or Enter to accept): ").strip()
+        feedback = request_human_feedback(
+            "\n🤔 Your feedback (or Enter to accept): ",
+            kind="review_plan",
+            origin={"stage": "fitting_plan"},
+        ).strip()
         
         if feedback == "":
             print("✅ Plan accepted.")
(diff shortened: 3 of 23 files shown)


sarah-allec (approved):
Approving — this is a careful, behavior-preserving piece of work and the design is sound. The channel abstraction is clean, the console fallback is genuinely byte-identical (resolving `builtins.input` at call time so monkeypatching front-ends keep working is a nice touch), thread-local overrides are the right mechanism, and the durability layer is best-effort in a way that can never break a prompt. The test coverage and the honest disclosure of the pre-existing `test_fusion_incremental` failure are appreciated.

Two non-blocking observations from reading `hitl.py`, `QueueChannel`, and the MCP `_MCPChannel`/`scilink_respond` seam (I did not deep-read the UI channel, the `input()`-site migration, or the `log_tail` buffer, so this isn't a full audit):

**1. `QueueChannel` timeout can leave a ghost prompt.** In `ask()` the `(req, holder, event)` tuple is enqueued once; on timeout the worker returns `req.default` but the tuple stays on the queue. A later `serve_pending()` then retrieves it and prompts the human for a question whose worker already gave up — and the answer is written to a `holder` nobody reads. The MCP `_MCPChannel` sidesteps exactly this by popping its pending entry in `finally`. The severity depends entirely on the timeout `QueueChannel` runs with in the fan-out path: if it's `None` (wait-for-coordinator), this can't fire and there's nothing to do. If fan-out ever sets a finite timeout, a cancelled-flag check in `serve_pending` (or draining on timeout) would close it. Mostly flagging to confirm the intended fan-out timeout — you have the context I don't here.

**2. Benign boundary race (both channels).** An answer arriving in the microseconds after the timeout fires is discarded, since `event.wait` already returned. Fine at a 1800s boundary — noting it only in case a comment there would save a future reader the double-take.

Neither blocks. Nice work.

PR #464, opened 2026-08-18
edit_file/rename_file: route oversize edits to one batched call; copy dead-end to relative path

ziatdinovmax:
## Summary

Two tool-message fixes that stop the planning agent from burning its 20-iteration turn budget on file housekeeping. Observed live: a meta delegation asked to add a section and a diagram to an existing white paper hit "Maximum tool iterations reached" — the section and figure had landed, but the figure never got embedded in the document.

**What went wrong**
- `edit_file`'s "too large" error told the agent to split an oversize verbatim insertion into *consecutive smaller edit_file calls* — contradicting the tool spec, which asks for one batched `edits` list. The agent obeyed the error: 11 calls (2 rejected, then one paragraph per call) for a single six-paragraph section.
- `rename_file(copy=true)` on a figure already in the current delegation folder returned "already this file's name — nothing to do". A copy can only land in the current folder by design, so when the *document* lives in an earlier delegation this was a dead end, and the agent spent its last iterations trying to work around it.

**Fix**
- Too-large message (shared default in `utils/file_edit.py`, planning, sim) now says: split into snippets under the cap and pass them TOGETHER as one `edits` list in a single call.
- The copy-into-own-folder case now routes to a relative-path reference from the other document's folder (`../<folder>/<file>`), keeping the deliberate "destination is never agent-chosen" design.

## Technical details

- `scilink/utils/file_edit.py` — `DEFAULT_TOO_LARGE_MESSAGE` reworded.
- `scilink/agents/planning_agents/orchestrator_tools.py` — planning `edit_file` too-large message reworded; `rename_file` gains a `copy`-specific branch on `rp == dest` with the relative-path routing.
- `scilink/agents/sim_agents/simulation_orchestrator_tools.py` — sim `edit_file` too-large message reworded.
- `tests/test_rename_copy_across_delegations.py` — same-name-copy test now asserts the routing message.
- `tests/test_edit_file_live.py` — new p11 (6 KB verbatim insertion) and p12 (figure here, document in an earlier delegation).

**Verification**
- Offline: 49/49 (`test_rename_copy_across_delegations`, `test_edit_file_tool`, `test_file_edit_utility`).
- Live on Bedrock Opus 4.8, 12/12: p11 lands the section in **one** `edit_file` call carrying a 6-item `edits` list (was 11 calls), 0 rejections; p12 hits the copy message, follows it, and embeds `../02_…/ai_loop.png` in the delegation-01 document in 4 tool calls, PDF twin refreshed.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/scilink/agents/planning_agents/orchestrator_tools.py
+++ b/scilink/agents/planning_agents/orchestrator_tools.py
@@ -5153,10 +5153,13 @@ class OrchestratorTools:
                         "revision use write_technical_document with "
                         "revise_path (whole-document rewrite under its "
                         "length guard). For a VERBATIM insertion, split "
-                        "the text at a unique boundary into consecutive "
-                        "smaller edit_file calls — do not rewrite the "
-                        "whole file with save_file, which keeps no "
-                        "backup and has no truncation guard."),
+                        "the text at unique boundaries into snippets under "
+                        "the cap and pass them TOGETHER as one `edits` "
+                        "list in a single call (each snippet inserted "
+                        "after the previous one) — not a chain of one-edit "
+                        "calls, which burns the turn's tool budget. Do not "
+                        "rewrite the whole file with save_file, which "
+                        "keeps no backup and has no truncation guard."),
                 )
                 if out["status"] != "success":
                     return json.dumps(out)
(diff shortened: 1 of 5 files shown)


sarah-allec (approved):
Approving — both fixes are correctly diagnosed and cleanly implemented, and they build on the #452 infrastructure without disturbing it (the sim message still routes oversize edits to `apply_input_adjustments`). The `rename_file(copy=true)` branch is placed right — inside the `rp == dest` block, so the copy case gets the relative-path embed while the move case keeps its "nothing to do" message — and it respects the "destination is never agent-chosen" design by advising a reference rather than inventing a copy target. Offline 49/49 and the live before/after (11 calls → 1; the copy dead-end → embedded in 4) are convincing.

Two non-blocking notes:

**1. The three too-large messages have drifted slightly out of sync.** The shared `DEFAULT_TOO_LARGE_MESSAGE` and the planning message both warn against "a chain of one-edit calls"; the sim message (`simulation_orchestrator_tools.py`) keeps "pass them TOGETHER as one `edits` list in a single call" but drops that explicit anti-chain clause. Since the whole point of this PR is to stop the one-edit-per-call pattern, the sim wrapper is the one place most likely to see it (deck edits), so it's the message you'd least want to be the terse one. Cheap to align the three — worth doing while they're all open in this diff.

**2. The relative-path hint assumes a flat delegation layout.** `../{dest.parent.name}/{safe_name}` is exactly right for sibling folders under `delegations/` (`../02_.../fig.png` from a doc in `01_...`), which is the layout today. It only breaks if delegation folders ever nest, or if the referencing document isn't one level up from the figure's folder — at which point `../` is wrong. Not worth generalizing now (the flat layout is a real invariant), but worth a one-line comment at the branch noting the assumption, so a future nesting change trips over the note rather than the bug.

Neither blocks merge.

PR #476, opened 2026-08-21
feat: deterministic geometry-consistency check for MolecularQC decks (#400)

### LINKED ISSUE #400: MolecularQC: deterministic geometry-consistency check after deck generation
## Summary

Add a deterministic geometry-consistency check after molecular QC deck generation: verify the deck's inline `geometry` block matches the source structure file on **atom count** and **element list** (composition), and fail generation loudly on a mismatch.

## Why

In `MolecularQCAgent` (NWChem), the deck's `geometry` block is transcribed from the structure by the LLM. A silent transcription error — a dropped atom, a wrong element, a miscounted multiplicity of some species — produces a deck that runs to completion on the *wrong* system. Nothing downstream catches it: the run succeeds, the output parses, and the result is simply for a different molecule than intended. Every other failure mode has a guard; this one doesn't.

## Proposed check

After deck generation, parse the inline `geometry` block and compare against the structure file (the same one the deck was generated from):
- atom count matches;
- element list / per-element counts match (composition identical).

Deterministic (no LLM), cheap, and fail-loud — a mismatch should surface as a generation error, not a warning, since a wrong-system deck is unrecoverable downstream.

## Context

Raised by @ziatdinovmax in review of #286 (MolecularQCAgent). Non-blocking for that PR; filed as the tracked follow-up.

sarah-allec:
Closes #400.

The NWChem deck's inline `geometry` block is LLM-transcribed from the source structure (`molecular_qc_agent.py` injects the structure as raw text and the model copies it into `geometry ... end`). A silent transcription error — a dropped atom, a wrong element — produces a deck that runs to completion on the **wrong** system: the run succeeds, the output parses, the answer is for a different molecule. Every other failure mode has a guard; this one didn't.

## What this adds
A deterministic, no-LLM, **fail-loud** check after deck generation.

- **New skill-bundle tool** `skills/molecular_qc/nwchem/nwchem_geometry.py` — `check_geometry_consistency(input_files, structure_file)` parses the deck's first `geometry ... end` block (element = first token per atom line; skips in-block directives like `units`/`symmetry`/`zcoord`) and compares **atom count** and **per-element composition** against the ASE-read structure. Returns `ok` / `mismatch` / `skipped`.
- **`MolecularQCAgent.generate_inputs`** resolves it through the registry (`active_skills=[software]`) after the deck is assembled and, on `mismatch`, returns an error status dict — generation fails loudly. `ok`/`skipped` proceed (a `skipped` covers no-geometry-block or unreadable-structure, so it never blocks an inapplicable case); a QC engine without the tool no-ops, matching the existing advisory syntax hook.
- **Engine-neutral:** the agent names no parser; a future QC engine adds its own bundle tool of the same name.

## Testing
`tests/test_molecular_qc_geometry_check.py` (8 cases, all passing): matching deck → `ok`; dropped atom / wrong element → `mismatch`; no geometry block / unreadable structure → `skipped`; reordered atoms → `ok`; registry resolution; and an agent-level test asserting `generate_inputs` returns `status="error"` on a mismatched deck.

## Scope note
Parses the first geometry block (QC decks are single-system); multi-fragment or nested-directive decks fall to `skipped` rather than a false failure.

### DIFF
--- a/scilink/agents/sim_agents/molecular_qc_agent.py
+++ b/scilink/agents/sim_agents/molecular_qc_agent.py
@@ -349,6 +349,39 @@ class MolecularQCAgent:
         if "entry_file" not in result and len(result["input_files"]) == 1:
             result["entry_file"] = next(iter(result["input_files"]))
 
+        # ── deterministic geometry-consistency check (fail loud) ──────
+        # The deck's inline geometry is LLM-transcribed from the structure; a
+        # dropped atom or wrong element would run to completion on the wrong
+        # system. Compare composition against the source and fail generation on
+        # a mismatch. Engine-neutral: resolved from the active skill bundle, so
+        # a QC engine without this tool simply skips it.
+        try:
+            from ...skills._shared._registry import get_tool_function
+            try:
+                geom_check = get_tool_function(
+                    "check_geometry_consistency", active_skills=[software]
+                )
+            except LookupError:
+                geom_check = None
+            if geom_check is not None:
+                check = geom_check(
+                    input_files=result["input_files"],
+                    structure_file=structure_file,
+                )
+                result["geometry_check"] = check
+                if check.get("status") == "mismatch":
+                    return {
+                        "status": "error",
+                        "message": (
+                            "Generated deck geometry is inconsistent with the "
+                            "source structure: " + check.get("reason", "")
+                        ),
+                        "geometry_check": check,
+                        "raw_result": result,
+                    }
+        except Exception as exc:
+            self.logger.warning("geometry consistency check failed: %s", exc)
+
         result["status"] = "success"
         result["software"] = software
         return result
--- a/scilink/skills/molecular_qc/nwchem/nwchem_geometry.py
+++ b/scilink/skills/molecular_qc/nwchem/nwchem_geometry.py
@@ -18,44 +18,68 @@ from typing import Any, Dict, List, Optional
 
 from ..._shared._spec import ToolSpec
 
-# Non-atom lines that can appear inside a `geometry` block.
+# Plain non-atom directive lines that can appear inside a `geometry` block.
 _GEOMETRY_DIRECTIVES = {
-    "units", "symmetry", "zcoord", "zmatrix", "load", "system", "adjust",
-    "autosym", "noautosym", "autoz", "noautoz", "center", "nocenter", "print",
+    "units", "autosym", "noautosym", "autoz", "noautoz", "center", "nocenter",
+    "print", "system", "adjust",
 }
+# Sub-directives whose presence means the block is NOT a plain Cartesian list we
+# can count atom-for-atom (internal coords, external/loaded coords, nested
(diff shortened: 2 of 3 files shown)


ziatdinovmax (changes requested):
One change request before merge, because it's a false-positive path in a fail-loud guard.

## Unrecognized atom lines silently vanish, then fail loud

`_ELEMENT_RE` is `^([A-Z][a-z]?)` — case-sensitive — but NWChem tags are case-insensitive, so `o 0.0 0.0 0.0` is a legal atom line. The parser silently skips any token it can't match, which converts "couldn't parse this line" into "this atom doesn't exist". Confirmed on the branch:

```
geometry units angstrom
  o 0.0 0.0 0.0
  c 0.0 0.0 1.16
  o 0.0 0.0 2.32
end
```

→ `{'status': 'mismatch', 'reason': 'atom count 0 (deck) vs 3 (source); ...'}`

That's a correct deck rejected loudly — the exact failure the module's own docstring promises to avoid ("we skip the check rather than risk a false fail") — and it kills the whole workflow as `failed_input_generation`. The same silent-skip applies to any legitimately-spelled atom token the regex doesn't anticipate, so this is a class, not one case. (Note the inconsistency that makes it easy to miss: the ghost/dummy check right below already compares case-insensitively via `el.lower()`, but the element extraction above it doesn't.)

Two-part fix, both small:

1. Match elements case-insensitively and normalize with `.capitalize()` before appending, so lowercase decks count correctly.
2. Treat any ≥4-token line inside a Cartesian block that *still* doesn't parse as an atom as `reliable: False` → `skipped`, instead of dropping the line. That closes the whole class: genuine transcription errors (dropped atom, wrong element — real symbols, wrong ones) still fail loud, while "the parser didn't understand this line" degrades to skip, consistent with the rest of the design.

Please add a lowercase-deck test pinning the behavior (and ideally one with an exotic-but-≥4-token line asserting `skipped`).

Everything else is merge-ready - happy to merge as soon as this lands.

sarah-allec:
Good catch — that's exactly the false-fail the module was supposed to avoid, and you're right that the case-insensitive ghost/dummy check right below made it easy to miss. Fixed both parts in the latest commit:

1. Element extraction is now case-insensitive: the leading 1–2 letters are captured, normalized with `.capitalize()`, and checked against the real element symbols (`ase.data.chemical_symbols`), so `o` / `CL` / `cl` all count.
2. A ≥4-token line inside a Cartesian block whose leading token is **not** a real element is now `reliable: False` → `skipped`, rather than dropping the line — so "the parser didn't understand this line" degrades to skip, while a genuine transcription error (dropped atom, or a real-but-wrong element) still fails loud.

Added tests: a lowercase deck → `ok`, a mixed-case deck → `ok`, and an unrecognized-token line → `skipped`. Full suite is 14/14. Ready for another look.

ziatdinovmax (approved):

Issue #407, opened 2026-08-01
Forward-model skills on SimulationAnalysisAgent (EXAFS, STM/STEM image simulation)


sarah-allec:
### Idea (from @ziatdinovmax on #405)

Serve **forward models** with the same `SimulationAnalysisAgent`: a forward model takes a finished DFT/MD run and computes an observable from it — an STM LDOS map from a DFT charge density, an EXAFS χ(k) spectrum, a STEM multislice image. That is exactly the agent's shape (data → property from a finished run), so it fits the availability-gated, codegen-based design without a new agent.

### Proposed structure

A sibling skill domain served by the same agent:

```
scilink/skills/
├── simulation_analysis/       # scalar properties (viscosity, band_gap, diffusion, …)
└── forward_models/            # forward-simulated observables
    ├── stem_multislice/       # requires: dft_output|trajectory → image
    ├── stm_ldos/              # requires: dft_output → image
    └── exafs_feff/            # requires: dft_output → curve
```

Each skill declares `computes` / `requires` exactly as the analysis skills do today; the availability gate routes them by which run output is present. This composes with the cross-scale analysis demo already validated (HOMO-LUMO / band gap / diffusion recovered from real NWChem / VASP / LAMMPS output).

### The one real extension

`BaseAnalysisAgent.compute_property` currently returns a **scalar** and the verification gate judges a single number. Forward models return a **curve / image / datacube**, so this needs:

- a generalized `compute_property` output type (scalar | curve | image | datacube),
- a matching verification/plausibility gate per output type,
- artifact handling (save + reference the produced curve/image, not just a JSON value).

### Scope

Out of scope for #405 (the analysis agent + scalar skills). Track here so the output-type generalization is designed deliberately. See `docs/proposals/simulation-analysis-agent-design.md`.

sarah-allec:
**Scope update after #477 (merged).**

The *"one real extension"* this issue called out is done: `BaseAnalysisAgent.compute_property` now supports `scalar | curve | image | datacube` output types, with a per-output-type verification gate and artifact save/reference handling, and `SimulationAnalysisAgent` now serves a `forward_models/` skill domain alongside `simulation_analysis/` — availability-gated exactly like the scalar skills.

What remains here is the enumerated forward-model **skills** themselves. `forward_models/` currently ships one reference skill (`structure_factor`, S(q) → curve); still to add:

- [ ] `stem_multislice` (dft_output | trajectory → image)
[...]

PR #477, opened 2026-08-21
feat: generalize SimulationAnalysis to non-scalar outputs + serve forward_models (#407)

sarah-allec:
Addresses the core of #407 — "**the one real extension**": generalizing the analysis agent's output contract from **scalar** to **scalar | curve | image | datacube**, so forward models (EXAFS χ(k), STM LDOS, STEM images) fit the existing availability-gated, codegen-based agent. Everything is additive; **no existing scalar skill changes**.

## What changed
- **`BaseAnalysisAgent.compute_property`** takes `output_type` (default `scalar`). A non-scalar script writes its artifact into `OUTPUT_DIR` and returns an `artifact` reference (`{path, format, shape}`) plus `summary` statistics. The agent **resolves + validates** the artifact (absolutizes the path; a success that wrote no artifact becomes an error), relaxes the verify trigger to fire on a value **or** an artifact, and `_verify_result` **branches per output type** — the scalar path is byte-for-byte unchanged; curve/image/datacube are judged on their summary + shape.
- The codegen prompt now states the per-output-type contract.
- **`SimulationAnalysisAgent`** serves a sibling **`forward_models`** domain (`EXTRA_DOMAINS`), discovered and availability-gated identically to the scalar skills (the router was already data-driven), and threads each skill's `output:` frontmatter to `compute_property`.
- **Reference skill** `forward_models/structure_factor` — static structure factor S(q), a genuine scattering/diffraction forward observable, **dependency-free** (numpy + ase) — exercises the curve path end-to-end.

## Testing (27 pass)
Curve artifact collection; missing-artifact fail-loud; verify-runs-on-curve; **scalar regression**; `forward_models` domain served + gated on `trajectory`; and `output:` frontmatter threading into `compute_property`.

## Deliberate follow-ups (separate PRs)
- **EXAFS** — reuse `exafs_simulation/feff_tools` (already returns χ(k)); needs the skill's `TOOL_SPEC`s wired into the codegen agent + the external FEFF binary.
- **STM/LDOS** — needs an LDOS library + physics.
- **STEM/multislice** — needs abTEM added + physics.
- Image/datacube gates are contract-admitted with a summary-based check, extensible to figure-vision.

### DIFF
--- a/scilink/agents/sim_agents/base_analysis_agent.py
+++ b/scilink/agents/sim_agents/base_analysis_agent.py
@@ -170,6 +170,7 @@ class BaseAnalysisAgent(ABC):
         recipe: str = "",
         packages: Optional[List[str]] = None,
         verify: bool = True,
+        output_type: str = "scalar",
     ) -> Dict[str, Any]:
         """Compute one property from ``data_files`` via verified codegen.
 
@@ -198,7 +199,7 @@ class BaseAnalysisAgent(ABC):
             f"DATA_FILES = {json.dumps(data_files)}\n"
             f"OUTPUT_DIR = {json.dumps(str(self.output_dir))}\n\n"
         )
-        code = self._generate_code(task, data_files, recipe, pkgs)
+        code = self._generate_code(task, data_files, recipe, pkgs, output_type)
         error_info: Dict[str, Any] = {}
         for attempt in range(self.max_refinement_attempts + 1):
             if attempt > 0:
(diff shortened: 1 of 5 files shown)


ziatdinovmax (changes requested):
This is awesome - thanks for extending/generalizing the SimulationAnalysis as we discussed. One quick item to address before the merger ->

## Change request: the retry path never states the non-scalar contract

`_generate_code` gained the per-output-type contract block, but `_refine_code` still tells the model "same contract: … print one JSON object on the last stdout line" — it was not touched by this PR. We've been bitten by exactly this before: a codegen contract change has to reach **all** prompt variants, refinement included, with a retry-path test to force it. And it bites hardest exactly where this PR depends on it: the new "success without artifact" error deliberately routes into `_refine_code`, so the retry that exists to fix a contract violation is the one prompt that doesn't state the contract. The reference skill masks this by restating the JSON contract in its recipe, but the base agent's contract can't depend on skill authors doing that.

Fix: thread `output_type` into `_refine_code`, include the same `output_contract` block there, and add a retry-path test (first attempt violates the artifact contract → refined attempt is prompted with it).

## Verification gaps (confirmed on the branch)

1. **The gate judges claims, not the artifact.** Nothing ever reads the file back — I had a script write a plain-text file, claim it was `npy` with shape `[2, 100]` and a clean summary, and it returned `status: success`. A cheap deterministic readback before the LLM gate — `np.load` shape + NaN count for `npy`, row count for `csv`, image size for `png` — turns the judged statistics from self-reported into measured, and catches the shape/format lies for free. This is the same silent-fallback class we've hardened elsewhere.
2. **No containment check on the artifact path.** `_resolve_artifact` only requires existence: my probe returned the *input trajectory itself* as the artifact and it was accepted. One line — require the resolved path to live under `OUTPUT_DIR`.

---

Also, two non-blocking minor notes ->

- The S(q) recipe writes the Debye equation as `sinc(q r_ij)` meaning sin(x)/x, but `np.sinc` is the normalized sin(πx)/(πx) — a codegen model transcribing literally gets wrong peak positions that the summary gate cannot see. Worth a one-line warning in the skill's `implementation` section ("write `np.sinc(q*r/np.pi)` or `sin(qr)/(qr)` explicitly").
- `output_type` isn't validated against `{scalar, curve, image, datacube}`; a typo'd frontmatter value flows into the prompts silently. A whitelist-with-warning would give skill authors feedback, consistent with how the loader warns on off-vocabulary sections.

sarah-allec:
Thanks — all three blocking items and both minor notes addressed.

**1. Refine path now states the non-scalar contract.** Pulled the output-contract clause into a shared `_output_contract(output_type)` used by *both* `_generate_code` and `_refine_code`, and threaded `output_type` into `_refine_code`. So the "success without artifact" retry — which deliberately routes into the refine loop — is now prompted with the artifact contract it exists to enforce, instead of the scalar default. Added `test_refine_prompt_states_the_artifact_contract` (first attempt writes no artifact → assert the *refine* prompt carries "curve observable" / "WRITE the computed …").

**2a. The gate now judges the artifact, not the claim.** `_resolve_artifact` reads the file back deterministically via `_measure_artifact` — `np.load` shape + NaN count for npy, row/col + NaN for csv, a header-parsed size for png (dependency-free) — and hands those *measured* facts to `_verify_result`, which now prefers them over the self-reported summary. A structured file that can't be read as its declared format is rejected → retry. Your probe (plain-text file claimed as npy `[2,100]`) is now caught: `test_readback_catches_format_lie`, plus `test_measured_facts_override_reported_shape` (script claims `[9,9]`, real file is `(2,100)` → resolved artifact reports the measured shape).

**2b. Containment.** `_resolve_artifact` requires the resolved *real* path (realpath both sides, so a symlink can't escape) to live under `OUTPUT_DIR`. Returning the input trajectory as the artifact is now rejected: `test_artifact_must_be_contained_in_output_dir`.

**Minors.** S(q) skill now warns that `numpy.sinc` is the normalized `sin(πx)/(πx)` and to write `np.sinc(q*r/np.pi)` or `sin(qr)/(qr)` explicitly. `output_type` is validated against `{scalar, curve, image, datacube}` with a warning on an off-vocabulary value (`test_offvocab_output_type_warns`).

Suite: `test_sim_base_analysis_agent.py` 20/20; broader analysis suite green.

ziatdinovmax (approved):
Looks great! Approving...

PR #478, opened 2026-08-21
fix: derive structure_class from scale in the one-shot pipeline (#474)

### LINKED ISSUE #474: Meta orchestrator builds solvated/aqueous systems under the crystal structure skill - structure_class isn't auto-selected
## Summary
  
  The one-shot `run_simulation` tool builds structures with a hardcoded
  `structure_class="crystal"` regardless of `scale`. For a classical-MD run of a
  condensed-phase system (`scale="molecular_dynamics"`, explicit water + ions), the
  solvated box is therefore built **and validated** under the `crystal` skill and rubric.
  There is no way to get `condensed` through this path — `run_simulation` exposes no
  `structure_class` argument, and nothing derives the class from `scale`.

  ## Root cause (v0.0.62, commit `19822ab0`)

  `run_simulation` → `run_complete_workflow` → `_run_workflow_once`:

  - `simulation_orchestrator_tools.py:862` — `run_simulation(description, run_command, scale, software, …)` has **no `structure_class` parameter**, and its
  call to `run_complete_workflow` (`:910–921`) never passes one.
  - `simulation_pipeline.py:707` — `run_complete_workflow` is a thin `**kwargs` passthrough to `_run_workflow_once`.
  - `simulation_pipeline.py:348` — `_run_workflow_once(..., structure_class="crystal", ...)` hardcodes the default and forwards it verbatim to
  `StructurePipeline.generate_and_validate` (`:461`). **`scale` is never consulted** to pick the class.

  Net: `scale="molecular_dynamics"` (inherently liquid/condensed) still builds under
  `crystal`. The two-step `plan_structure` → `generate_structure(structure_class=...)`
  path *can* set the class, but the one-shot `run_simulation` path (the natural choice for
  "build + run in one call") cannot.

  ## Reproduction
  
  Delegate a classical-MD task for a solvated ionic system and tell the specialist to
  `run_simulation(scale="molecular_dynamics")`. Logs show, for every build:

  ```
  Loaded structure_generation skill: crystal
  Loaded validation rubric for structure_class: crystal
  ```

  ## Impact

  Solvated/liquid systems are validated against crystal-appropriate criteria — mismatched
  validation. Structures still build (the LLM follows the request text), but the
  condensed-phase packing guidance and rubric are never applied.

  ## Suggested fix

  Derive the class from `scale` when not explicitly supplied — e.g. in `_run_workflow_once`,
  default `structure_class` from a `scale → class` map (`molecular_dynamics → condensed`)
  via a `None` sentinel so explicit callers still win — and/or expose `structure_class` on
  `run_simulation` / thread it through `run_complete_workflow`. Optionally run
  `StructurePlanner` inside the one-shot path when the class is unspecified.

sarah-allec:
Closes #474.

The one-shot `run_simulation` path hardcoded `structure_class="crystal"` in `_run_workflow_once` regardless of `scale`, so a classical-MD run of a solvated / condensed-phase system (explicit water + ions) was built **and validated** under the `crystal` skill and rubric — the condensed-phase packing guidance and validation criteria were never applied. The two-step `plan_structure` → `generate_structure` path could set the class; the one-shot path couldn't.

## Fix
- **`_run_workflow_once`**: `structure_class` defaults to `None` and is derived from `scale` via `_SCALE_STRUCTURE_CLASS` (`molecular_dynamics → condensed`, `periodic_dft → crystal`, `molecular_qc → molecular`; unknown → `crystal`). An **explicit `structure_class` still wins**, so a biomolecular MD run can pass `"biomolecular"`. `run_complete_workflow` already threads it via `**kwargs`.
- **`run_simulation` tool**: exposes an optional `structure_class` override (signature + tool spec) and passes it through, so a caller can override the derived default.

## Tests
`tests/test_structure_class_from_scale.py` (6): scale→class derivation for each scale, explicit override wins, unknown-scale fallback to crystal, and the public `run_complete_workflow` passthrough — all by stubbing `StructurePipeline` to capture the class it receives.

## Scope note
The derivation is a deterministic scale→class default (a strict improvement over the crystal hardcode). Running `StructurePlanner` inside the one-shot path to pick the class from the description (e.g. condensed vs biomolecular MD) is a possible follow-up; explicit callers and the two-step path already cover that today.

### DIFF
--- a/scilink/agents/sim_agents/simulation_orchestrator_tools.py
+++ b/scilink/agents/sim_agents/simulation_orchestrator_tools.py
@@ -1003,10 +1003,16 @@ class SimulationOrchestratorTools:
                     "type": "string",
                     "description": (
                         "Optional structure-class override ('crystal' | "
-                        "'molecular' | 'condensed' | 'biomolecular'). Omit to "
-                        "derive it from the scale (molecular_dynamics -> "
-                        "condensed, periodic_dft -> crystal); set it explicitly "
-                        "for a biomolecular MD system."
+                        "'molecular' | 'condensed' | 'biomolecular'). Omit and it "
+                        "is derived from the scale, where the molecular_dynamics "
+                        "default 'condensed' means a liquid / solution / solvated "
+                        "box. MD also covers crystalline and biomolecular systems, "
+                        "and the derived default is wrong for those — so SET IT "
+                        "EXPLICITLY from the task: 'crystal' for a crystalline "
+                        "solid, a melt-from-crystal, or a slab interface (e.g. "
+                        "melting Cu, Li diffusion in LiCoO2, water on a TiO2 slab); "
+                        "'biomolecular' for a protein. periodic_dft derives "
+                        "'crystal', molecular_qc derives 'molecular'."
                     ),
                 },
                 "max_run_cycles": {
--- a/scilink/agents/sim_agents/simulation_orchestrator_tools.py
+++ b/scilink/agents/sim_agents/simulation_orchestrator_tools.py
@@ -861,6 +861,7 @@ class SimulationOrchestratorTools:
         # =====================================================================
         def run_simulation(description: str, run_command: str = None,
                            scale: str = None, software: str = None,
+                           structure_class: str = None,
                            max_refinement_cycles: int = 4,
                            max_run_cycles: int = 3,
                            run_timeout: int = 3600) -> str:
@@ -910,6 +911,7 @@ class SimulationOrchestratorTools:
                 result = run_complete_workflow(
                     description,
                     scale=scale, software=software,
+                    structure_class=structure_class,
                     output_dir=str(workdir),
                     api_key=self.orch.api_key, base_url=self.orch.base_url,
                     model_name=self.orch.model_name,
@@ -996,6 +998,16 @@ class SimulationOrchestratorTools:
                     "type": "string",
                     "description": ("Optional engine override "
                                     "('vasp' | 'lammps'); routed if omitted."),
+                },
+                "structure_class": {
+                    "type": "string",
+                    "description": (
+                        "Optional structure-class override ('crystal' | "
+                        "'molecular' | 'condensed' | 'biomolecular'). Omit to "
+                        "derive it from the scale (molecular_dynamics -> "
+                        "condensed, periodic_dft -> crystal); set it explicitly "
+                        "for a biomolecular MD system."
+                    ),
                 },
                 "max_run_cycles": {
                     "type": "integer",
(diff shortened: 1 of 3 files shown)


ziatdinovmax (changes requested):
One revision request before merge:

## `molecular_dynamics → condensed` collides with #432

Since #432, the MD scale deliberately absorbs inorganic solids, metals, and interfaces — its motivating queries (melting Cu, Li diffusion in LiCoO₂, water on a TiO₂ slab) start from crystalline or crystal-slab structures. Under this map, a one-shot `run_simulation(scale="molecular_dynamics")` for those now builds and validates the starting structure under the *condensed* rubric — the same mismatched-rubric failure #474 describes, in the opposite direction, for tasks we just made first-class. So this isn't strictly an improvement over the crystal hardcode: it fixes solvated boxes and flips crystalline-solid MD.

The tool spec compounds it: the description says "set it explicitly for a biomolecular MD system" but never tells the specialist LLM — the actual caller, which has the task description in hand — to pass `crystal` for crystalline-solid MD.

**Asks:**

1. Reword the `structure_class` parameter description (and the `_run_workflow_once` docstring note): frame the derived `condensed` as the liquids/solutions default, and instruct explicitly — `crystal` for crystalline solids, melts-from-crystal, and slab interfaces; `biomolecular` for proteins. That fixes the live path, because the caller is an LLM reading that description.
2. Make the deprecated `machine_learning_potentials` shim's mapping deliberate — it's absent from the map so it falls to `crystal`, while its `molecular_dynamics` equivalent gets `condensed`. An entry or a comment, either way on purpose.
3. File the description-aware derivation (running `StructurePlanner` when the class is unset) as a tracked follow-up issue rather than a scope-note sentence — the scale→class map is now structurally unable to be faithful for MD, so the follow-up is where this actually gets resolved.

Happy to merge once (1) lands; (2) and (3) are small enough to bundle here.

sarah-allec:
Thanks — good catch that the map flips crystalline-solid MD. Addressed all three:

**1. Wording (the live path).** Reworded the `run_simulation` `structure_class` tool-spec description and the `_run_workflow_once` docstring + derivation comment: `condensed` is now framed as the MD **liquids/solutions default**, with an explicit instruction to the caller to pass `crystal` for a crystalline solid, a melt-from-crystal, or a slab interface (I named the melting-Cu / Li-in-LiCoO₂ / water-on-TiO₂-slab cases directly in the tool spec so the specialist LLM sees them), and `biomolecular` for a protein.

**2. MLP shim mapping is now deliberate.** Added `machine_learning_potentials: "condensed"` to the map with a comment — it's treated as an MD task, so it shares MD's default on purpose rather than falling through to `crystal`. New test locks this in.

**3. Follow-up filed as #482** — replace the scale→class default with a description-aware `StructurePlanner` derivation when the class is unset, which is where the MD-can't-be-one-default problem actually gets resolved.

Tests: added `test_explicit_crystal_wins_for_crystalline_md` and `test_mlip_shim_derives_condensed`; suite 8/8 green.

ziatdinovmax (approved):
Claude is happy with the revision - merging it.

PR #480, opened 2026-08-21
Added generic file/document read, write, append tools to the simulation orchestrator

jenna1701:
The simulation orchestrator (simulate mode) exposed only edit_file and rename_file for file work — no way to read files or persist arbitrary text, unlike the analysis and planning agents. I ported the generic file tools those agents already implement onto the simulation orchestrator.

  Tools added to SimulationOrchestratorTools (scilink/agents/sim_agents/simulation_orchestrator_tools.py)
  - read_file — the full-featured reader from the planning orchestrator: text/JSON/CSV/log + PDF/DOCX text-extraction, with offset/tail/search/max_lines, section-outline hints on truncation, and whole-file reads for report-stemmed files. Reworded for sim artifacts (INCAR, POSCAR, OUTCAR, LAMMPS decks, job scripts, run logs).
  - read_document — the analysis orchestrator's document reader: combines PDF/DOCX/MD/TXT into extracted text (tables preserved, scanned pages OCR'd via the session model).
  - save_file / append_file — write/append text into the session directory, with the chunked-write pattern (save first chunk, append the rest).

  All four resolve relative paths against the session base_dir (matching the existing edit_file/rename_file sandbox convention), and I added a module-level _extract_document_text helper + import pandas as pd.

  Tests (tests/test_simulation_orchestrator.py)
  - Added the four names to EXPECTED_TOOLS (the registry-equality assertion in test_1_orchestrator_constructs — which passes).
  - Added test_2c_generic_file_io, a zero-LLM targeted test exercising save→append→read round-trips, regex search on an extensionless INCAR, tail on a long log, JSON pretty-print, the  missing-file error path, and document extraction. It passes.

  Verification
  - Module imports cleanly; all 26 tool schemas serialize.
  - Ran the targeted suite: my new test and the tool-registry test pass.
  - One suite failure — "Tool error paths" — is pre-existing and unrelated: I confirmed it fails identically on the unmodified main code. It's a sandbox-environment quirk where Path('/nonexistent/POSCAR').exists() raises PermissionError instead of returning False, inside validate_structure (code I didn't touch).

I also fixed a minor typo in CLAUDE.md (Ramand fixed to Raman on line 33).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -30,7 +30,7 @@ a new subagent class is justified only when its execution structure
 itself cannot be expressed within an existing agent. This applies
 across all three modes. For example, adding an XRD or Raman skill for 
 existing CurveFittingAgent is strongly preferred over creating two new agents
-for Ramand and XRD.
+for Raman and XRD.
 
 ## Foundation agents
 
--- a/scilink/agents/sim_agents/simulation_orchestrator_tools.py
+++ b/scilink/agents/sim_agents/simulation_orchestrator_tools.py
@@ -22,6 +22,34 @@ from datetime import datetime
 from pathlib import Path
 from typing import Any, Callable, Dict, Optional
 
+import pandas as pd
+
+# read_document character cap — longer documents are truncated (~50k tokens).
+_READ_DOC_MAX_CHARS = 200_000
+
+
+def _extract_document_text(path: Path, ocr_model: Any = None) -> Dict[str, Any]:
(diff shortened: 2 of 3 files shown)


ziatdinovmax (approved):
Thanks! Approving as is. Also, a thought: read/edit/append/rename file are pretty general utilities not specific to any of three modes. In the (near) future, it would make sense to have them moved out of agentic modules into the unified file-I/O utility. I've opened an issue for that -> #481.

PR #483, opened 2026-08-21
Fix condensed-box charge neutrality + reuse a built structure in run_simulation

sarah-allec:
Two autonomous classical-MD stability fixes surfaced while running end-to-end aqueous MD over a batch of ionic conditions. Only a few conditions completed; each burned many retry cycles. Root-caused to two issues (of four total; the other two — unresolved MD-deck template variables and a 300 s structure-build timeout — are follow-ups).

> **Stacked on #478** (`fix/structure-class-from-scale`). Until #478 merges, the diff here also shows its commits; the changes unique to this PR are the condensed skill + `run_simulation` structure-reuse below.

## 1 — Charge-imbalanced composition (the retry-storm root cause)

The condensed structure skill said the box must be "net-neutral when ions are involved" but gave **no charge-balance arithmetic**. For a multivalent salt the LLM under-counted the shared counter-ion — e.g. a box with 4 Na⁺ and 1 Mg²⁺ (**+6**) got only 2 OH⁻ + 2 Cl⁻ (**−4**), leaving it at **+2**. The build script's neutrality guard then raised, the script exited **rc 1**, and the follow-on `run_simulation` failed **"file or path not found"** (no structure was written) and retried — the dominant time sink.

**Fix (skill knowledge, per the usual skill+LLM split):** add explicit guidance to `condensed.md` — balance on *total charge* weighting each ion by valence (Σ nᵢzᵢ over cations = Σ nⱼ|zⱼ| over anions), a worked multivalent example (Mg²⁺ needs 2 OH⁻; a shared counter-ion neutralizes every cation at once), and an instruction to **compute the signed charge sum in-script and correct the counts before packing** rather than emitting an imbalanced box and hard-failing.

## 4 — Structure built twice

`run_simulation` always rebuilt the structure, even when `generate_structure` had just produced one for the same system — duplicating the (expensive) build and producing divergent workdir slugs for one condition.

**Fix:** add a `structure_file` argument to `run_simulation` (accepts a path *or* a session slug from `generate_structure`), threaded into `run_complete_workflow`, which already skips generation when handed a structure. A missing file returns a structured error. The tool-spec description and an orchestrator convention tell the model to pass the prior `structure_path` instead of building a second copy.

## Tests
- `condensed.md`: knowledge change (no unit test; exercised by structure-gen runs).
- `run_simulation` reuse: new `test_2c_run_simulation_reuses_prebuilt_structure` — a path and a slug are both forwarded as `structure_file`, omitting it forwards `None`, and a bogus path is a structured error. Targeted orchestrator suite 10/10; `import scilink` clean.

### DIFF
--- a/scilink/agents/sim_agents/simulation_orchestrator.py
+++ b/scilink/agents/sim_agents/simulation_orchestrator.py
@@ -170,6 +170,11 @@ inside `generate_structure` when MP_API_KEY is available.
   regenerates the inputs and can perturb parameters you did not mean to change);
   `edit_file`'s cap-exceeded hint points you there when a change is too big for a
   surgical edit. Use `rename_file` to change a filename byte-exactly.
+- `run_simulation` builds the structure itself when none is given. If you have
+  ALREADY built the structure this run (a prior `generate_structure` call),
+  pass that structure's `structure_path` to `run_simulation`'s `structure_file`
+  argument so it reuses it instead of rebuilding — don't run `generate_structure`
+  and then let `run_simulation` build a second copy of the same system.
 """
 
 
--- a/scilink/agents/sim_agents/simulation_orchestrator_tools.py
+++ b/scilink/agents/sim_agents/simulation_orchestrator_tools.py
@@ -890,10 +890,37 @@ class SimulationOrchestratorTools:
         def run_simulation(description: str, run_command: str = None,
                            scale: str = None, software: str = None,
                            structure_class: str = None,
+                           structure_file: str = None,
                            max_refinement_cycles: int = 4,
                            max_run_cycles: int = 3,
                            run_timeout: int = 3600) -> str:
             from .simulation_pipeline import run_complete_workflow
+
+            # Reuse an already-built structure instead of rebuilding it. If a
+            # prior generate_structure produced a structure for this condition,
+            # pass its `structure_path` here (the caller may give a path or the
+            # slug of a session-generated structure) — the workflow then skips
+            # structure generation entirely. This avoids the duplicated build
+            # (and the divergent workdir slugs) when the orchestrator runs
+            # generate_structure before run_simulation.
+            if structure_file and not Path(structure_file).is_file():
+                match = next(
+                    (s for s in (self.orch.generated_structures or [])
+                     if s.get("slug") == structure_file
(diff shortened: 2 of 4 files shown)


ziatdinovmax (changes requested):
One blocking issue to address before the merge ->

## The neutrality auto-correction must be loud, and must not override an explicit request

The new Implementation bullet says: compute `Q`, and if `|Q| > 0` "correct the counts — add/adjust the most abundant counter-ion … rather than raising", failing only "if neutrality cannot be reached". Two problems:

- **The fail branch is dead.** With counter-ion adjustment allowed, neutrality is *always* reachable, so every imbalance — whatever its origin — gets corrected. The guard can never fire.
- **The correction is silent, and it mutates the science.** Adjusting an ion count changes the composition the box models — and in your own motivating batch (a scan over ionic conditions), the species being adjusted in the worked example is OH⁻, i.e. the pH-determining ion that may be the very variable the batch is sweeping. A silently "corrected" condition produces a successful build of the *wrong point in the scan*, which is worse than the retry storm it replaces: the storm was loud.

Can you address them?

sarah-allec:
Good catch — you're right that "silently correct" is worse than the retry storm it replaces (the storm was loud), and that the fail branch was dead. Reworked the guidance:

- **Free vs pinned.** The charge-balance derivation now applies *only* to a **free counter-ion** — a species added solely to neutralize, whose count carries no independent scientific meaning. It explicitly must **not** change the count of anything the request pins: a fixed concentration, a stoichiometric salt (NaCl → equal Na⁺/Cl⁻), or a controlled/swept variable like the OH⁻/H⁺ that sets pH.
- **Fail loud, for real.** The in-script `Q = Σ nᵢzᵢ` check stays as verification, but if the requested composition is itself charged and no free counter-ion can absorb it without altering a pinned quantity, the build **raises and exits non-zero**, naming `Q` and the offending ions — no nudging a requested ion to force `Q == 0`. So the failure branch now actually fires (on a genuinely charged/over-specified request), and a wrong scan point can never be built silently.

The worked example is framed with hydroxide as the *free* neutralizer; the caveat covers the case where OH⁻ is instead the pinned/swept variable — which was exactly your concern.

ziatdinovmax (approved):
Thanks for addressing it quickly - approving

PR #485, opened 2026-08-21
run_simulation: auto-reuse the session's built structure (fixes #4 double-build loop)

sarah-allec:
Follow-up to #483. That PR made `run_simulation` *accept* a `structure_file`, but in practice neither the sim orchestrator nor the meta orchestrator threads `generate_structure`'s returned path through — so `run_simulation` kept **rebuilding** the structure on every call, each into a fresh `description`-slugged workdir.

Observed end-to-end (an autonomous MD-BO run): `generate_structure` builds a complete, runnable setup in one dir (`structure.extxyz` + `system.data` + `run.lammps`); the orchestrator then calls `run_simulation` several times, each creating a **new, differently-slugged, empty** dir; the rebuild there diverges/fails, no `log.lammps` is produced, and the caller re-words the request and retries — an infinite loop in which the engine never runs, even though the first dir was ready to go.

## Fix
When `run_simulation` is called with **no `structure_file`**, reuse the most-recent structure this session already built and **run in its dir** instead of rebuilding:

- The sim child is shared across meta delegations (`meta_orchestrator.py` caches `self._children["simulation"]`), so `generate_structure`'s session record persists and is visible here.
- Reuse is **announced** in the result (`reused_structure`), never silent.
- An explicit `structure_file` (path or session slug) still forces a specific structure — for a genuine multi-system run where the latest isn't the target.
- A fresh build happens only when the session has **no** structure yet.
- Session records are appended only when a structure was actually written, so a prior empty/failed run is never picked up for reuse.

This makes the fix robust across **both** orchestrators without depending on either LLM obeying a prompt convention to thread the path.

Also updates the `structure_file` tool-spec description and the sim-orchestrator convention to state that reuse is automatic (no manual threading), and reworks `test_2d` to cover: empty-session fresh build, explicit path, session-slug resolution, **auto-reuse with run-in-reused-dir**, and the missing-file error.

## Verification
Targeted orchestrator suite **11/11**; `import scilink` clean; `test_structure_class_from_scale` 8/8.

### DIFF
--- a/scilink/agents/sim_agents/simulation_orchestrator.py
+++ b/scilink/agents/sim_agents/simulation_orchestrator.py
@@ -171,10 +171,13 @@ inside `generate_structure` when MP_API_KEY is available.
   `edit_file`'s cap-exceeded hint points you there when a change is too big for a
   surgical edit. Use `rename_file` to change a filename byte-exactly.
 - `run_simulation` automatically reuses a structure this session already built
-  (a prior `generate_structure` or `run_simulation`) and runs in its dir, so a
+  (a prior `generate_structure` or `run_simulation`) and runs it, so a
   `generate_structure` → `run_simulation` sequence does NOT rebuild — you need
-  not thread the path yourself. Pass `structure_file` only to force a SPECIFIC
-  structure (e.g. a multi-system run where the most-recent one isn't the target).
+  not thread the path yourself. This assumes the run targets the session's
+  CURRENT system. To simulate a DIFFERENT system, do not merely reword the goal:
+  build it first with `generate_structure`, or pass `structure_file="new"` to
+  force a fresh build — otherwise the request runs against the previous system's
+  structure. Pass a specific path/slug to reuse a particular earlier structure.
 """
 
 
--- a/scilink/agents/sim_agents/simulation_orchestrator.py
+++ b/scilink/agents/sim_agents/simulation_orchestrator.py
@@ -170,11 +170,11 @@ inside `generate_structure` when MP_API_KEY is available.
   regenerates the inputs and can perturb parameters you did not mean to change);
   `edit_file`'s cap-exceeded hint points you there when a change is too big for a
   surgical edit. Use `rename_file` to change a filename byte-exactly.
-- `run_simulation` builds the structure itself when none is given. If you have
-  ALREADY built the structure this run (a prior `generate_structure` call),
-  pass that structure's `structure_path` to `run_simulation`'s `structure_file`
-  argument so it reuses it instead of rebuilding — don't run `generate_structure`
-  and then let `run_simulation` build a second copy of the same system.
+- `run_simulation` automatically reuses a structure this session already built
+  (a prior `generate_structure` or `run_simulation`) and runs in its dir, so a
+  `generate_structure` → `run_simulation` sequence does NOT rebuild — you need
+  not thread the path yourself. Pass `structure_file` only to force a SPECIFIC
+  structure (e.g. a multi-system run where the most-recent one isn't the target).
 """
 
 
--- a/scilink/agents/sim_agents/simulation_orchestrator_tools.py
+++ b/scilink/agents/sim_agents/simulation_orchestrator_tools.py
@@ -922,6 +922,29 @@ class SimulationOrchestratorTools:
                     ),
                 })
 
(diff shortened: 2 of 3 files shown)


ziatdinovmax (changes requested):
Two changes before merge ->

## 1. There is no way to get a fresh build once the session has any structure

The new default is "reuse the most recent structure regardless of what `description` says," and the only override — `structure_file` — can only force a *different existing* structure. So after the first build, a bare `run_simulation("MD of water")` in a session whose last structure is a Cu slab runs the water request against the slab, and nothing checks the description against the reused record. The caller learns from `reused_structure` only after the engine has already run — in an autonomous loop, plausible-but-wrong output may be consumed before anyone reads the note. Cross-workflow makes it wider: `run_complete_dft_workflow` also appends session records, so an MD call can silently pick up a DFT-built structure.

Two-part fix, both small:

- **An explicit opt-out** — `structure_file="new"` (or a `reuse=False` flag) that forces a fresh build for this description. Without it the model's only route to a second system is knowing to call `generate_structure` first, which nothing tells it.
- **Say so in both prompt surfaces** — the convention bullet and the tool-spec description should state plainly: automatic reuse assumes the run targets the session's current system; for a DIFFERENT system, build it first (`generate_structure`) or pass `structure_file="new"`. Extend `test_2d` with the opt-out case.

(A description-vs-record relevance check is NOT the answer — the motivating loop reworded the same request, so text matching would defeat the fix. The escape hatch + prompt guidance is the right level.)

## 2. Repeated runs now clobber outputs in the shared structure dir

Running in the reused dir is what makes the fix work, but every subsequent `run_simulation` on the same structure regenerates inputs and reruns in that one dir — a temperature/pressure sweep over one structure (the common same-structure-many-runs case) overwrites each prior run's `log.lammps`/outputs. Land each run in a per-call subdir of the structure dir (`runs/<NN>_<slug>/`, structure referenced by path) — or, if you'd rather keep the flat layout for now, say so explicitly in the PR and we track it; but it shouldn't be an accident.

Nice touches worth keeping exactly as-is: the notes list replacing the single `note` field, and appending records only when a structure was actually written so failed runs never become reuse targets.

sarah-allec:
Added the execution fix to this PR. With reuse working, the ready deck still wasn't being *run*: the executor gate only built a `LocalExecutor` when no `hpc_connection` was attached. In an **in-allocation** MD run with a remote connection present, that left `run.lammps` + `system.data` sitting unexecuted (no `log.lammps`) and pushed the orchestrator toward `submit_simulation_job` (SSH) — wrong when the job *is* us.

Fix: also run locally when `SLURM_JOB_ID`/`SLURM_JOBID` is set — inside an allocation the compute node is here, so execute the deck directly (`LocalExecutor` → `lmp -in …`) rather than deferring to a remote submit. Out-of-allocation remote submission (`hpc_connection`, no `SLURM_JOB_ID`) still generates-only for the `submit_simulation_job` path. New `test_2e` covers both branches; targeted suite 12/12.

sarah-allec:
Both addressed.

**1. Fresh-build opt-out + prompt guidance.** Added `structure_file="new"` — it forces a fresh build for the current description even when the session already holds a structure, so a different system can be started without the model having to know to call `generate_structure` first. Both prompt surfaces now say it plainly: the `run_simulation` tool-spec description and the orchestrator convention state that automatic reuse assumes the run targets the session's *current* system, and that a **different** system must be built first (`generate_structure`) or requested with `structure_file="new"` — *not* by rewording the goal. Per your note, no description-vs-record text matching (the motivating loop reworded the same request, which would defeat it) — just the escape hatch + guidance.

**2. Per-run subdirs.** A reused structure now lands each run in `runs/<NN>_<slug>/` under the structure dir, structure referenced by path — so a same-structure sweep (temperature/pressure) no longer clobbers prior `log.lammps`/outputs. A fresh build still gets its own slug dir.

`test_2d` now covers the `01_`/`02_` per-run subdirs and the `structure_file="new"` opt-out (asserts a fresh build, `reused_structure` null). Kept the notes-list and write-then-record touches as-is. Targeted suite 12/12.

(Also in this PR from the in-allocation report: `run_simulation` now executes locally when `SLURM_JOB_ID` is set even with an `hpc_connection` attached — `test_2e`.)

ziatdinovmax (approved):

PR #486, opened 2026-08-21
Fixed import error in route_simulation when fairchem not installed and added deterministic mlip inventory

jenna1701:
Import error was showing when route_simulation was called and the fairchem module was not installed

  Root cause: route_simulation → SimulationRouter.__init__ → AvailableSoftware.auto() probes each engine skill's declared Python modules. The uma skill declares fairchem.core.
  importlib.util.find_spec('fairchem.core') must import the parent package fairchem to locate the submodule — and when fairchem isn't installed, that raises ModuleNotFoundError instead of
  returning None. The exception propagated up and aborted the whole tool.

  Fix (scilink/utils/available_software.py, _probe_python_modules): wrap find_spec in a try/except (ImportError, ValueError), treating any probe failure as "not importable here" (spec =
  None) — the same outcome as a genuinely-absent module. ImportError covers the ModuleNotFoundError from a missing parent package; ValueError guards the __spec__-is-None edge case
  find_spec can raise.

  Verified:
  - The exact failing probe (fairchem.core) now returns None.
  - AvailableSoftware.auto() and SimulationRouter(model=None) (the line in your traceback) both complete without error.
  - Positive detection is preserved (ase → available), and other missing modules return None cleanly.

  route_simulation now works whether or not fairchem is installed — it simply reports the uma backend as unavailable.

---- 

"List available MLIPs" returned an error, as the simulation orchestrator had no deterministic listing tool. The LLM's only nearby option was route_simulation, which is a decision tool. When the model is (mis)used as an inventory lookup, that JSON parse is fragile — surfacing to the user as a "transient JSON-parsing error each time."

Fix: Added a new list_available_software tool to SimulationOrchestratorTools — a pure inventory lookup over AvailableSoftware.auto() + skill-bundle discovery, with no LLM call, so it
  cannot hit a parsing error. It reports:
  - mlip_backends: installed (e.g. chgnet), known (the full set the code can drive — mace, chgnet, deepmd, uma, orb, excluding the general strategy bundle), and not_installed;
  - installed_by_domain: every installed engine per scale;
  - a hint on installing a missing backend.

  Its description explicitly steers the LLM here for "which MLIPs/engines can I use" and reserves route_simulation for choosing an engine for a goal — so inventory questions no longer
  misroute.


sarah-allec (approved):
Approving — both changes are correct and the targeted suite is green (12/12 on this branch).

**`_probe_python_modules` fix:** a real crash fix. `find_spec('fairchem.core')` imports the parent `fairchem` to locate the submodule, so an absent parent raises `ModuleNotFoundError` — which was aborting all of `route_simulation`. Wrapping in `try/except (ImportError, ValueError)` → `spec=None` is precise and idiomatic (`ImportError` covers the missing parent; `ValueError` guards `find_spec`'s `__spec__`-is-None edge), and positive detection is preserved.

**`list_available_software`:** verified `AvailableSoftware.domains()` / `.list_available(domain=…)` behave exactly as assumed (installed `[]`, known `['chgnet','deepmd','mace','orb','uma']` with `general` correctly excluded). Deterministic, no LLM, clean error handling — and the `_NoLLM` poison object in the test nicely proves no LLM is touched. Good call steering inventory questions here and reserving `route_simulation` for choosing an engine.

Non-blocking nits (fine to leave or fold in):
- The new test is `test_2d_list_available_software`, next to the existing `test_2d_run_simulation_reuses_prebuilt_structure` — two `test_2d_*`; a unique number (and a blank line between the two functions, E302) would tidy it.
- The no-domain path iterates every domain including experimental ones (`curve_fitting`, `image_analysis`); harmless since they return `[]`, but a sim-only scope would be tighter.

Heads-up on merge order: this and #485 both touch `EXPECTED_TOOLS`, `TARGETED_TESTS`, and the same test-file region, so whichever lands second will need a small rebase.

PR #484, opened 2026-08-21
Remove legacy engine-specific sim agents (engine-neutral cleanup)

sarah-allec:
The live simulation path — `simulation_orchestrator` → `simulation_orchestrator_tools` → `simulation_pipeline` / `simulation_router` → foundation agents (`PeriodicDFTAgent`, `MDSimulationAgent`, `MolecularQCAgent`, `MLIPAgent`, `ForceFieldAgent`) + skill bundles — is fully engine-neutral. The old engine-specific agents are dead from it: the dependency direction is always legacy → foundation (`VaspInputAgent(PeriodicDFTAgent)`, `LAMMPSSimulationAgent(MDSimulationAgent)`), never the reverse, and no live-path module imports any of them. This removes them ahead of the engine-agnostic story in the paper.

## Removed (10 modules — no live-path importer)
- **LAMMPS:** `lammps_orchestrator.py`, `lammps_agent.py`, `lammps_utils.py`, `lammps_analysis.py`, `lammps_analysis_updater.py`, `lammps_updater.py`
- **VASP:** `vasp_agent.py`, `vasp_updater.py`, `vasp_quality.py`, `post_run_analysis.py` (a support module only `vasp_quality` used)

## Ripple
- `sim_agents/__init__.py` — dropped the 7 legacy re-exports (`VaspInputAgent`, `VaspQualityAgent`, `LAMMPSSimulationAgent`, `LAMMPSUpdater`, `LAMMPSAnalysisAgent`, `LAMMPSAnalysisUpdater`, `LAMMPSOrchestrator`).
- **Tests** — deleted `test_vasp_quality.py`, `test_vasp_updater_fixes.py`; dropped the `LAMMPSSimulationAgent` back-compat half of `test_md_agent_smoke.py`; dropped the `lammps_orchestrator` import-smoke line in `test_skill_live.py`.
- **`examples/vasp/`** — removed (all three scripts demo the deleted `VaspUpdater`/`VaspQualityAgent` path).
- **Docs** — rewrote the stale Simulation-Agents section of `README.md` (it also referenced never-existing `StructureOrchestrator`/`DFTOrchestrator`) and updated the post-run / debugger references in `CLAUDE.md` to the engine-neutral surface (`analyze_output`, `refinement.py`).

## Kept (verified live despite the name)
[...]

PR #490, opened 2026-08-22
Make base_agent._generate_json robust to an empty/garbled completion

sarah-allec:
Follow-up to #485 — this fix was pushed to the #485 branch *after* it merged, so it didn't make it into `main`. Re-submitting it cleanly on top of current `main` (cherry-pick of the stranded commit; the `test_md_agent_smoke` reconciliation drops the `LAMMPSSimulationAgent` backward-compat test that #484 already removed).

## Why
End-to-end testing of #485 (in-allocation MD) showed `run_simulation` still never executing — the failure was **upstream of the executor**. `MDSimulationAgent`'s deck-generation LLM call occasionally returns an **empty completion** (`JSON parse failed: Expecting value: line 1 column 1 (char 0)` = empty string — a proxy/content-filter hiccup, not the `response_mime_type` config, which `_map_gen_config` silently drops). `base_agent._generate_json` did `json.loads(response.text)` and **raised on the first bad response**, with only a regex fallback that can't rescue an empty string. That abort propagated up, the specialist burned its iteration budget re-issuing `run_simulation` (proliferating workdirs), and it never reached the `LocalExecutor` path.

## Fix
`_generate_json` now retries a few times (default 3), nudging explicitly for JSON on the retries and salvaging an embedded object before giving up, instead of aborting on one bad completion. Since `_run_workflow_once` returns `failed_input_generation` (does not raise), a run that now succeeds records its structure, so `run_simulation`'s reuse also stops the dir proliferation.

## Verification
`test_md_agent_smoke.py` gains a `_generate_json` regression test: rides through empty completions (returns on the 3rd call), salvages a prose-wrapped object, and raises only after the retry budget. Suite passes; `import scilink` clean.

### DIFF
--- a/scilink/agents/sim_agents/base_agent.py
+++ b/scilink/agents/sim_agents/base_agent.py
@@ -192,23 +192,45 @@ class SimulationAgent(ABC):
     # LLM HELPERS
     # ================================================================
 
-    def _generate_json(self, prompt: str) -> Dict[str, Any]:
-        try:
+    def _generate_json(self, prompt: str, _attempts: int = 3) -> Dict[str, Any]:
+        """Generate a JSON object from the model, robust to a bad completion.
+
+        An occasional empty or non-JSON response — a proxy hiccup, a
+        content-filter stop, a stray prose wrapper — otherwise raises on the
+        first try and aborts the whole generation, which is expensive mid-workflow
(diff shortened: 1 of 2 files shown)


ziatdinovmax (approved):
All look good - approving

PR #491, opened 2026-08-22
Expose simulation agents over MCP (scilink serve --mode simulate)

sarah-allec:
`scilink serve` exposes the **analysis** and **planning** agents over MCP but not **simulation** — so an MCP client (e.g. a LangChain deepagents deep agent) can drive analysis/planning but not DFT/MD/MLIP. This adds a standalone **`--mode simulate`** that serves the `SimulationOrchestratorTools` surface as clean `scilink_*` tools, replicating the exact pattern the analysis/planning modes use.

## Design
- **Standalone mode, not folded into `both`.** Several sim tool names collide with the analysis/planning generic tools (`read_file`, `session_status`, `view_structure`, `save_file`, …); a standalone mode keeps clean names (`scilink_run_simulation`, `scilink_generate_structure`, …) with **no `scilink_sim_` prefix**. Cross-mode workflows already have a home in `--mode meta`, which routes to simulation via `delegate_to_simulation`. `both` stays analyze+plan, unchanged.
- **Mirrors the planning block** everywhere in `mcp_server.py`, including a **guarded construction** — the `[sim]` extra is optional, so a checkout without it just won't offer the mode rather than breaking `scilink serve`.

## Wiring
- `cli/serve.py`: `simulate` added to `--mode` choices (`--print-mcp-json` forwards it unchanged).
- `_init_orchestrators`: guarded `SimulationOrchestratorAgent` construction for `mode == "simulate"`; `state["simulation_orch"]` slot.
- `_ensure_initialized` / `list_tools`: sim `tool_map` + listing loops (plain `scilink_` prefix).
- `scilink_orchestrate_simulation`: background-job tool → `SimulationOrchestratorAgent.run_task`, via `_handle_orchestrate` + a `SimulationMode` branch in `_execute_run_task_captured`; a not-available error points at `--mode simulate`.
- Long tools (`run_simulation`, `run_complete_dft_workflow`, `run_mlip_simulation`, `run_exafs_workflow`, `analyze_output`, `submit_simulation_job`, `generate_final_report`, `orchestrate_simulation`) marked **background-capable**; the run tools gate for approval in co-pilot/autopilot like `run_analysis`/`run_optimization`.
- `set_autonomy` propagates to the sim orchestrator.

## Verification
[...]

PR #492, opened 2026-08-22
docs: list the --mode tool sets in the client guide (incl. simulate)

sarah-allec:
The unified client guide (`docs/connecting_agent_clients.md`) used `--mode both` in its examples but never enumerated the choices. Adds a one-paragraph note listing `analyze` / `plan` / `both` / `simulate` / `meta`, so the new `--mode simulate` surface (simulation agents over MCP, #491) is discoverable from the guide. Docs-only.


ziatdinovmax (approved):
I was just about to update the docs... thanks for doing it!

PR #493, opened 2026-08-22
lammps analyze_structure: classify aqueous electrolytes correctly

sarah-allec:
The `analyze_structure` system-category heuristic in the LAMMPS skill labeled **any** bonded, carbon-bearing system `biomolecular` — so an aqueous electrolyte (water + salt + organic co-solvent) came back `biomolecular`. That label feeds the MD planning prompt, so a Zn(OTf)₂ / water / sulfone electrolyte was planned as if it were a protein/nucleic-acid system.

Fix: **require nitrogen for the `biomolecular` category** (protein/nucleic-acid backbones carry N; organic electrolyte species — triflate, sulfone, carbonates — are carbon-bearing but N-free), and add `electrolyte` / `solution` / `molecular_liquid` categories for the bonded aqueous and organic-liquid cases.

Verified on the real UC2 Zn(OTf)₂ / water / sulfone box (no N) → `electrolyte`; the alanine-dipeptide fixture (has N) stays `biomolecular`. Adds an electrolyte fixture + test (`test_lammps_skill` 12/12).

Surfaced while bringing the UC2 electrolyte-validation use case onto current `main` — the driver ran end-to-end, and this was the one substantive mislabel worth fixing upstream.

### DIFF
--- a/scilink/skills/molecular_dynamics/lammps/lammps.py
+++ b/scilink/skills/molecular_dynamics/lammps/lammps.py
@@ -486,20 +486,33 @@ def parse_data_file(data_file: str) -> Dict[str, Any]:
         and ec.get("H", 0) >= 2 * ec.get("O", 0)
         and info["has_bonds"]
     )
+    # Monatomic cations common in electrolytes (alkali, alkaline-earth, and a few
+    # multivalent cations incl. Zn2+) plus halide anions. NOTE: this does not see
+    # polyatomic anions (triflate, sulfate, nitrate) — a halide-free box with such
+    # an anion falls to "solution" rather than "electrolyte" (both non-biomolecular,
+    # so nothing downstream breaks). Teaching the ion check about polyatomic anions
+    # is tracked in #494.
     info["has_ions"] = bool(elems & _HALIDES) or bool(
-        elems & {"Na", "K", "Ca", "Mg", "Li", "Rb", "Cs"}
+        elems & {"Na", "K", "Ca", "Mg", "Li", "Rb", "Cs", "Zn", "Al", "Sr", "Ba"}
     )
     info["has_organic"] = "C" in ec and info["has_bonds"]
     info["has_metal"] = bool(elems & _METALS) and not info["has_bonds"]
     info["has_semiconductor"] = bool(elems & _SEMICONDUCTORS) and not info["has_bonds"]
 
     # Category (bonded molecular systems first, then non-bonded solids).
-    # "biomolecular" requires nitrogen, not just any bonded carbon: proteins and
-    # nucleic acids carry N in their amide/amine/base backbones, whereas most
-    # organic electrolyte and solvent species (triflate, sulfone, carbonates,
-    # glymes) are carbon-bearing but N-free. Keying on N (rather than "C or N")
-    # stops those liquids from being swept up as biomolecular, while a solvated
-    # small molecule with N is rare enough to accept the occasional miss.
+    # KNOWN LIMITATION (#494): "biomolecular" keys on nitrogen (proteins and
+    # nucleic acids carry N in their amide/amine/base backbones; N-free organic
+    # electrolyte/solvent species — triflate, sulfone, carbonates, glymes — do
+    # not). N is a PROXY, not a biomolecule signature: it mislabels N-bearing
+    # SMALL-molecule electrolytes and solvents — acetonitrile, imidazolium /
+    # ammonium / pyridinium ionic liquids, amide solvents (DMF/DMAc/NMP),
+    # organic-nitrate systems — as biomolecular, because this branch is checked
+    # before the electrolyte/solution ones. Those are common chemistries, not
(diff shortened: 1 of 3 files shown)


ziatdinovmax (changes requested):
The added change may be "overfitting" to UC2's chemistry, introducing the mirror image of the bug it fixes. Specifically, 

### The N-based discriminator overfits to UC2's N-free anion

Requiring N for `biomolecular` fixes the triflate/sulfone case **because UC2's electrolyte happens to be nitrogen-free** — not because N is a biomolecule signature. Presence of N is a proxy that holds for this one system's chemistry. The symmetric failure it creates is large and common: because `biomolecular` is checked *first* inside the `has_bonds` block, any **N-bearing** small-molecule electrolyte or solvent now classifies as biomolecular before ever reaching the `electrolyte`/`solution` branch:

- **acetonitrile (CH₃CN)** — the canonical aprotic electrolyte solvent
- **ionic liquids** — imidazolium, pyridinium, ammonium, pyrrolidinium cations are all N+C (an entire electrolyte class)
- **amide solvents** — DMF, DMAc, NMP
- **nitrate electrolytes** with any organic co-solvent

An acetonitrile/NaCl/water box → `biomolecular` → planned as a protein system: the exact mislabel this PR fixes for triflate, displaced onto very common chemistries. So the code comment's "a solvated small molecule with N is rare enough to accept the occasional miss" **materially understates it** — ionic liquids and MeCN are not occasional. That sentence, left as-is, will convince the next reader the case is negligible.

The honest discriminator between a biomolecule and a small N-bearing electrolyte molecule is **size/topology**, not element identity: proteins/nucleic acids are large bonded polymers; MeCN, imidazolium, triflate are small. `atom_count` and `bond_count` are already parsed, so a size/bond-density floor is feasible — though I acknowledge it's genuinely hard to set cleanly, since the test fixture (alanine *dipeptide*, ~22 atoms) and imidazolium (~19 atoms) overlap. That difficulty is the point: there may be no clean structural separator from this parse, which is worth saying out loud rather than papering over with an N proxy.

**Asks:**
1. Rewrite the comment to name the real failure class (N-bearing electrolytes / ionic liquids / amide solvents → mislabeled biomolecular), not "occasional miss."
2. Add a test pinning the current N-bearing-electrolyte behavior (e.g. a MeCN + NaCl + water fixture) so the limitation is visible and tracked, not silently wrong.
3. File the size/topology-based discriminator as the real follow-up.

### Secondary (non-blocking): the `electrolyte` category rides on halide presence

`has_ions` only counts halides + alkali/alkaline-earth — it does **not** recognize transition-metal electrolyte cations (Zn²⁺) or polyatomic anions (triflate, sulfate, nitrate). The real UC2 Zn(OTf)₂ box classifies as `electrolyte` only because **triflate contains F** (F ∈ `_HALIDES`); Zn²⁺ itself is invisible to the ion check. Swap the anion for nitrate or sulfate (no halide) and the same electrolyte falls to `solution` instead. Both labels are non-biomolecular so nothing breaks, but the `electrolyte`/`solution` boundary is more accidental than it looks — worth a note, and ideally teaching `has_ions` about Zn²⁺ / common polyatomic ions.

sarah-allec:
Thanks — this is the right critique, and I've reframed the change around it. Pushed a3d0f73.

**1. Honest comment about the failure class.** You're right that "accept the occasional miss" undersold it. The N-check is a *proxy*, and the systems it mislabels aren't rare — I rewrote the comment to say so plainly: N-bearing **small-molecule** electrolytes and solvents (acetonitrile, imidazolium/ammonium/pyridinium ionic liquids, amide solvents like DMF/DMAc/NMP, organic-nitrate systems) get swept up as `biomolecular` because that branch is checked before the electrolyte/solution ones. These are common chemistries.

**2. Test pinning the current (wrong) behavior.** Added a `MeCN + NaCl + water` fixture and a test that asserts it currently classifies as `biomolecular`, with a `FIXME(#494): -> electrolyte`. It's deliberately a *characterization* test — it locks in the limitation so it stays visible and the assertion flips the day the real fix lands, rather than the behavior silently drifting.

**3. Follow-up issue for the real discriminator.** Filed #494. The honest fix is size/topology (biopolymers are large; MeCN/imidazolium/triflate are small), but atom/bond counts alone don't separate them cleanly — alanine dipeptide (~22 atoms) overlaps imidazolium (~19) — so it needs more than a threshold. That's scoped there; this PR is the interim heuristic plus an honest account of what it gets wrong.

**Plus (non-blocking, from your note):** extended `has_ions` to recognize Zn/Al/Sr/Ba cations. Left a note that polyatomic anions (triflate/sulfate/nitrate) are still unseen — a halide-free box with such an anion falls to `solution`, not `electrolyte` (both non-biomolecular, so nothing downstream breaks); folded into #494.

Full lammps-skill suite green (130 passed).

ziatdinovmax (approved):

PR #499, opened 2026-08-24
Add GROMACS as a molecular_dynamics engine skill

sarah-allec:
Makes **GROMACS** a first-class MD engine alongside LAMMPS. The `MDSimulationAgent` was already engine-neutral except in one spot — it wrote the generated run-control file under a hardcoded `run.lammps` entry name. This parameterizes that by engine and registers a GROMACS tools module; skill discovery (`list_skills`), the extension map, and the `validate_script` hook were already generic.

### What's in the bundle
- **`gromacs.md`** — engine knowledge with a `detect:` block on `gmx`/`gmx_mpi`/`mdrun`. The primary generated artifact is the **`.mdp`** run-control file; the skill covers integrator/`dt`/`nsteps`, v-rescale / nose-hoover thermostats, c-rescale / parrinello-rahman barostats for NVT vs NPT, PME + Verlet cutoffs, constraints, and the staged min→NVT→NPT→production workflow (`gmx grompp` → `gmx mdrun`).
- **`gromacs.py`** — deterministic `parse_mdp` (dash/underscore-normalized, comment-stripped) and `validate_script(script_path, system_info)` matching the LAMMPS validator contract (`{valid, errors, warnings, ...}`): integrator/timestep sanity, a thermostat on dynamics runs, a well-formed NVT-vs-NPT pressure-coupling block, `ref-t`/`tc-grps` field-count agreement, and cutoff/electrostatics presence.

### Agent integration
- Register the `gromacs` tools module in `_TOOL_REGISTRY`.
- Add an `_ENTRY_FILE` map + `_entry_name()` helper (`lammps`→`run.lammps`, `gromacs`→`grompp.mdp`, `openmm`→`run_openmm.py`) and use it at the three entry-file sites (single, fan-out campaign, staged). **LAMMPS behavior is unchanged** — still `run.lammps` / `run_<stage>.lammps`.

### Tests
`tests/test_gromacs_skill.py` (12): parser normalization + last-assignment-wins; validator on good NPT/NVT, a minimizer, a `ref-t`/`tc-grps` mismatch, missing barostat params, a large-timestep warning, and an unreadable file; plus agent integration (discovery, registry membership, engine-parameterized entry name, LAMMPS unchanged). The LAMMPS-skill suite (130) and MD-agent smoke/FF tests stay green.

### Why
This makes the "one agent per method, engines as skills" MD side extensible in practice, not just in principle — and backs the architecture figure's GROMACS chip with real code.

### Scope
GROMACS is first-class for **single-shot** and **staged** (min→NVT→NPT→production) generation. **Fan-out parameter sweeps are not yet supported** — the bundle ships no `expand_parameter_sweep`, so a sweep request degrades to a single run with a loud warning (`_sweep_spec`). A sweep expander is the follow-up.

### DIFF
--- a/scilink/agents/sim_agents/md_simulation_agent.py
+++ b/scilink/agents/sim_agents/md_simulation_agent.py
@@ -817,6 +817,16 @@ class MDSimulationAgent(SimulationAgent):
             return None
         if not (self.tools_module
                 and hasattr(self.tools_module, "expand_parameter_sweep")):
+            # A sweep WAS requested (multiple sims + >=2 values) but the active
+            # engine has no expander — degrade to a single run LOUDLY, never
+            # silently. GROMACS is first-class for single + staged generation;
+            # fan-out sweeps are not in the bundle yet.
+            self.logger.warning(
+                "engine %r has no parameter-sweep expander; producing a single "
+                "run instead of the requested %d-point sweep over %r — express "
+                "it as a staged campaign for now",
+                self.skill_name, len(values),
+                plan.get("variable_parameter") or "parameter")
             return None
         return {"var_name": plan.get("variable_parameter") or "parameter",
                 "values": values}
--- a/scilink/agents/sim_agents/md_simulation_agent.py
+++ b/scilink/agents/sim_agents/md_simulation_agent.py
@@ -19,6 +19,11 @@ try:
     _TOOL_REGISTRY["lammps"] = lammps_tools
 except ImportError:
     pass
+try:
+    from ...skills.molecular_dynamics.gromacs import gromacs as gromacs_tools
+    _TOOL_REGISTRY["gromacs"] = gromacs_tools
+except ImportError:
+    pass
 
 
 def _assemble_fanout_stage(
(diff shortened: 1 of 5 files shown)


ziatdinovmax (changes requested):
One blocking issue and one minor issue to address ->

## Change request: parameter sweeps silently degrade to a single run

`_sweep_spec` returns `None` when the engine lacks `expand_parameter_sweep` (`md_simulation_agent.py:794`), and GROMACS's tools module doesn't provide it. Because the sweep-prompt block is also gated on `SWEEP_PLACEHOLDER` (also absent), a GROMACS request like *"run a temperature sweep at 300/350/400 K"* produces **one** simulation — no fan-out, no error, no warning. The user (or an MCP client, or the meta agent) gets a single run and no signal that the requested sweep was dropped. That's exactly the quiet wrong-result the codebase avoids elsewhere, and it undercuts the PR's "first-class MD engine alongside LAMMPS" claim: GROMACS is first-class for single + staged generation, not for fan-out.

Two small asks:
1. **Make the degradation loud.** In `_sweep_spec`, when a sweep *is* requested (`requires_multiple_simulations` + >1 sims + ≥2 values) but the engine can't expand it, log a warning ("engine 'gromacs' has no parameter-sweep expander; producing a single run — request the values as a staged campaign instead") before returning `None`. Two lines, and the drop stops being silent.
2. **Scope the claim honestly** — in the PR description and ideally a line in `gromacs.md`: GROMACS covers single-shot and staged (min→NVT→NPT→production) generation; fan-out parameter sweeps are not yet supported (no `expand_parameter_sweep`/`SWEEP_PLACEHOLDER` in the bundle). That sets the right expectation and points the follow-up.

## Minor (non-blocking)

`split_shared_types` (`:729`) is the one **un**guarded `tools_module` call. It's reached only in the analysis-realizability conflict-resolution path, which is a LAMMPS `.data` shared-atom-types concept and shouldn't arise for GROMACS — so it's very likely unreachable here. Worth a defensive `hasattr` guard anyway (matching the other five sites) so a future realizability rule can't AttributeError a GROMACS run. And the `_DEFAULT_ENTRY = ("run", "lammps")` fallback for an unknown engine is a slightly odd default, but only reachable if `skill_name` is off-registry.

sarah-allec:
Both addressed — thanks, the silent sweep drop was a real gap. Pushed.

**Blocking — loud degradation.** `_sweep_spec` now warns before returning `None` when a sweep *was* requested (`requires_multiple_simulations` + ≥2 values) but the active engine has no expander, naming the engine and the dropped point count:

> engine 'gromacs' has no parameter-sweep expander; producing a single run instead of the requested 3-point sweep over 'temperature' — express it as a staged campaign for now

**Scope the claim honestly.** Added a **Scope** note to `gromacs.md` and to the PR description: GROMACS is first-class for single-shot and staged (min→NVT→NPT→production) generation; fan-out parameter sweeps are not yet supported (no `expand_parameter_sweep` in the bundle) and degrade to a single run *with the warning above*. A sweep expander is the follow-up.

Tests: the bundle exposes no `expand_parameter_sweep`, and a sweep request on GROMACS emits the warning (caplog).

**Minor — `split_shared_types`.** I believe this one is already guarded: `_enforce_analysis_realizability` returns early at its entry guard (`if self.tools_module is None or not hasattr(self.tools_module, "split_shared_types"): return script, info`, `md_simulation_agent.py:726-729`) — so a GROMACS run (whose tools module has no `split_shared_types`) never reaches the call at :753. I audited the other five `tools_module` calls too; all are guarded (inline `hasattr`, or transitively via `_sweep_spec` for `expand_parameter_sweep`). Happy to add a redundant call-site guard if you'd still prefer one. The `_DEFAULT_ENTRY` fallback I left as-is (only reachable for an off-registry `skill_name`).

ziatdinovmax (approved):

PR #501, opened 2026-08-24
Fix divide-by-None when a plan sets timestep=None (MD minimization)

sarah-allec:
Follow-up from the cross-engine parity sweep, and a **re-land**: this fix was on the #499 branch but landed one commit after #499 merged, so it didn't make it into main.

A minimization plan has no timestep — GROMACS uses `integrator = steep` — so the LLM plan sets `timestep: null`. `plan.get("timestep", 2.0)` returns `None` (the key *exists*, so the default is never used), and the equilibration/production step computation then does `float / None` → `TypeError`, crashing generation. In the parity sweep this hard-failed the GROMACS `md_minimize` case (LAMMPS minimize kept a timestep, so it never hit this).

**Fix:** use `plan.get(k) or <default>` for `timestep` and the equilibration/production times, so an explicit `None` falls back to the default. Regression test: `_generate_md_input` on a minimize plan with `timestep=None` returns a script instead of raising.

### DIFF
--- a/scilink/agents/sim_agents/md_simulation_agent.py
+++ b/scilink/agents/sim_agents/md_simulation_agent.py
@@ -576,12 +576,15 @@ class MDSimulationAgent(SimulationAgent):
                 f"{rows}\n"
             )
 
-        ts = plan.get("timestep", 2.0)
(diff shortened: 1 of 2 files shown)


ziatdinovmax (approved):
No issues found - approving

PR #505, opened 2026-08-26
fix: refine_structure no longer assumes POSCAR output format

jenna1701:
# fix: refine_structure no longer assumes POSCAR output format

## Problem

When refining a generated structure, the simulation agent reported that *the user had explicitly requested POSCAR format* and flagged a discrepancy between the generation script writing an **extxyz** file and the "requested" **POSCAR** file — **even when the user never asked for POSCAR**.

## Root cause

`refine_structure` (in `simulation_orchestrator_tools.py`) unconditionally appended `". Save the structure in POSCAR format."` to the refinement request whenever the request didn't already contain the word `"poscar"`:

```python
request = original_request
if "poscar" not in request.lower():
    request = request + ". Save the structure in POSCAR format."
```

This is stale code from when POSCAR was the default output format. Today `generate_structure` writes **engine-neutral extxyz** for crystals (the crystal skill declares `output_format: extxyz`; the engine-native POSCAR is written downstream by the engine step). So on every refinement:

- the request told the validator *"the user wants POSCAR"*, while
- the prior generation script actually wrote **extxyz**,

and the validator dutifully flagged a phantom "requested POSCAR vs wrote extxyz" discrepancy that nobody had asked for.

## Fix

Don't assume any default output format. On refinement:

1. If the user **explicitly named** a format in their request, honor it.
2. Otherwise, **round-trip whatever format the structure was originally generated in** (inferred from its filename), so the refinement request matches the prior script.

New helpers on `SimulationOrchestratorTools`:

- `_requested_format(request)` — the format the user explicitly named (poscar/vasp, xyz, extxyz, pdb, cif, lammps), or `None` (the common case). Never guesses.
- `_existing_structure_format(path)` — the format inferred from the already-generated file's name.

Also corrects a stale comment in `generate_structure` that claimed the orchestrator appends a POSCAR-format instruction — it appends the class-derived format (extxyz for crystals).

## Behavior after fix

| Scenario | Before | After |
|---|---|---|
| Crystal (extxyz) + plain request | forces POSCAR → phantom discrepancy | round-trips **extxyz**, no false flag |
| User explicitly asks for POSCAR | POSCAR | POSCAR (honored) |
| Molecule (`xyz`) / biomolecule (`pdb`) | forces POSCAR | round-trips own format |

## Testing

Added `test_2e_refine_preserves_structure_format` (pure, no LLM) covering both helpers and the exact request-building logic — asserts a crystal refine never injects POSCAR and that an explicit POSCAR request is still honored.

The full targeted suite passes except the pre-existing, environment-specific `Tool error paths` failure (a sandbox `PermissionError` on `/nonexistent`, unrelated to this change — it fails identically on `main`).

### DIFF
--- a/scilink/agents/sim_agents/simulation_orchestrator_tools.py
+++ b/scilink/agents/sim_agents/simulation_orchestrator_tools.py
@@ -501,9 +501,11 @@ class SimulationOrchestratorTools:
             # so crystal is the sensible default *class* (it supplies the
             # class-specific validation rubric). A user-supplied `skill` (rendered
             # into skill_content above) overrides the crystal *generation* skill;
-            # the crystal validation rubric still applies. The orchestrator
-            # appends the POSCAR-format instruction. (When the StructurePlanner
-            # lands it will set structure_class per request.)
+            # the crystal validation rubric still applies. The pipeline appends
+            # the output-format instruction derived from the structure_class
+            # skill's `output_format` frontmatter (extxyz for crystals), NOT a
+            # hardcoded POSCAR. (When the StructurePlanner lands it will set
+            # structure_class per request.)
             result = so.generate_and_validate(
                 description,
                 structure_class=structure_class,
@@ -1292,9 +1294,12 @@ class SimulationOrchestratorTools:
                     "message": f"Failed to construct StructureGenerator: {e}",
                 })
 
-            request = original_request
-            if "poscar" not in request.lower():
-                request = request + ". Save the structure in POSCAR format."
+            # Do NOT assume POSCAR. Honor a user-named format, else round-trip
+            # the format the structure was ORIGINALLY generated in (crystals
+            # default to extxyz). Shared with the regression test via
+            # _refined_request so the real code path is the one under test.
+            request = self._refined_request(
+                original_request, record["structure_path"])
 
             # Re-apply the same skill (if any) the original generation used,
             # so the refinement prompt has the same library guidance available.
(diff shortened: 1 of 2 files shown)


sarah-allec (changes requested):
Nice fix — the root-cause diagnosis (stale POSCAR default vs. the crystal skill's engine-neutral `extxyz`) is spot on, and honor-explicit-else-round-trip is the right principle. Clean helpers, good docstrings, and a pure regression test. One thing I'd like tightened before merge, plus a couple of minor notes.

## Requesting: `_requested_format` substring matching mis-fires on engine/source mentions

The keyword check is a bare substring scan over the whole request, so it fires on words that aren't format requests at all:

```
POSCAR       <- 'build rutile TiO2 for a VASP calculation'
lammps-data  <- 'relax the Si slab then run a LAMMPS MD'
pdb          <- 'a protein from the PDB database'
cif          <- 'the CIF from Materials Project'
xyz          <- 'optimize the xyz positions of the adatom'
```

The first two are the ones I'd worry about: users routinely name the **target engine** in a request. When `_requested_format` returns non-`None` from an incidental "VASP"/"LAMMPS" mention, the code skips the round-trip append — so the beneficial `extxyz` round-trip is lost, and the request carries an engine name the LLM validator may itself read as "wants POSCAR." That re-invites the exact phantom discrepancy this PR removes, just triggered by "vasp" instead of the old hardcoded string.

Suggestion: only treat it as an explicit format request when there's a real cue — e.g. `"in <fmt> format"`, `"save as <fmt>"`, `"write … as <fmt>"`, or an unambiguous token (`poscar`/`contcar`/`extxyz`/`lammps-data`) — and drop the bare ambiguous tokens (`vasp`, `lammps`, `xyz`, `pdb`, `cif`) unless they appear in that context. Word-boundary (`\b`) matching would also help.

## Minor

- The new test exercises a *copy* of the request-building logic (the inline `refined()` helper) rather than `refine_structure` itself, so the helpers are covered but the real code path isn't — small drift risk. Could extract the request-building into a tiny pure method (e.g. `_refined_request(request, path)`) that both `refine_structure` and the test call.
- Nit: `xyz` ⊂ `extxyz` is currently handled only by dict insertion order (`extxyz` listed before `xyz`). Worth a one-line comment so a future reorder doesn't silently break it — or the phrase/word-boundary approach above removes the hazard entirely.

The common case (plain request) is genuinely fixed; the engine-mention case is the one I'd like tightened since it's a frequent real input. Thanks!

sarah-allec (approved):

PR #506, opened 2026-08-26
fix: robust staged MD generation + LAMMPS transport/staging skill guidance

sarah-allec:
# fix: staged MD generation validates the split stages (catch a bad split before it runs)

## Problem

Staged MD campaigns (`staged=True`: minimize → NPT-equilibrate → NVT-produce) intermittently failed generation with `[…] full -> refinement_failed`, on two recurring LAMMPS deck errors the refinement loop could not repair at run time:

1. `Unresolved template variables: ['${visc}']` — the model computed shear viscosity *in* the LAMMPS input (an undefined `variable visc`).
2. `fix nvt and fix npt both on group(s): {'all'}` — a produce/equilibrate stage kept both integrators.

Surfaced running an experiment-validated electrolyte use case end to end.

## Root cause (corrected per review — thanks @ziatdinovmax)

`generate_staged_simulation` generates a monolithic deck *only as the source for the model to split*, then splits it into per-phase stages and returns those. Before this PR **the split stages were not validated** — a bad split (a stage that kept both integrators, or an unresolved `${visc}`) went straight to the engine, where the refinement loop's dry-run gate + fixer could not repair the categorical conflict → `refinement_failed`.

The monolithic deck's own validation does **not** gate anything: `_generate_classical_md_inputs` setdefaults `status="success"` and the pre-run `InputValidator` sees only stage 0. (An earlier version of this description wrongly said it did — corrected.)

## Fix

**Engine-neutral, in the MD agent.** `generate_staged_simulation` now:
1. **Validates each produced stage** via the engine-neutral `_validate` and, on any error, **re-splits with the concrete errors as feedback** (one integrator per stage) — up to three attempts. This per-stage structural check + retry is what the PR adds; it did not exist before.
2. **Last-resort per-stage repair:** if a stage is still invalid, runs the same engine-neutral `_attempt_fix` used on the monolithic deck and re-validates, before failing — preserving the recovery path the refinement loop's dry-run gate would otherwise have taken.
3. Only if a stage is *still* invalid does generation hard-fail (`status="error"`), rather than shipping a broken stage to the engine.
4. Unlinks the prior attempt's stage files on each retry (a retry may rename stages, e.g. `minimize` → `equilibration`).

**Skill (`lammps.md`):** a "Transport coefficients" entry stating the deck emits the raw stress log for the `viscosity_greenkubo` analysis skill (never an in-deck `${visc}`), and a positive `unfix`-between-stages template. The `fix ave/time` snippet now defines `variable pxy equal pxy` (etc.), since `fix ave/time` averages variables/computes, not bare thermo keywords.

## Behavior after fix

| Scenario | Before | After |
|---|---|---|
| Clean split | shipped unvalidated | validated, runs |
| Split keeps two integrators / `${visc}` | reached engine → `refinement_failed` | re-split with feedback, then `_attempt_fix`; runs if recoverable |
| Unrecoverable after retries + fix | reached engine → `refinement_failed` | `status="error"` at generation |

## Testing

`tests/test_md_staged_split.py` (no LLM/engine; real `lammps.validate_script`): clean split; conflicted → self-corrects on re-split; persistent-bad → repaired by `_attempt_fix`; unrepairable → `status="error"`.

**End-to-end:** the electrolyte use case now runs — all four members clear deck validation into the staged run, where before every member hit `full -> refinement_failed`.

### DIFF
--- a/scilink/agents/sim_agents/md_simulation_agent.py
+++ b/scilink/agents/sim_agents/md_simulation_agent.py
@@ -937,32 +937,55 @@ class MDSimulationAgent(SimulationAgent):
         plan = result["simulation_parameters"]
         impl = self._get_skill_context(section="implementation")
 
-        prompt = (
+        info = result.get("system_info", {})
+        base_prompt = (
             "Split this simulation into 2-4 checkpointed stages.\n\n"
             "SCRIPT:\n"
             f"{full_script}\n\n"
             f"{impl}\n\n"
             "Each stage: complete, standalone, runnable. First reads data file,\n"
             "later stages read restart. All include force field commands.\n"
+            "Each stage uses exactly ONE time integrator (e.g. NPT to equilibrate,\n"
+            "NVT to produce) — never two on a group in the same stage.\n"
             "Use literal values. Write restart at end of each stage.\n\n"
             'Return JSON: {"equilibration": "script...", "production": "script..."}'
         )
-        try:
-            stages = self._generate_json(prompt)
-        except Exception:
-            stages = {"production": full_script}
 
-        stage_scripts = {}
-        steps = []
-        for name, content in stages.items():
-            if not isinstance(content, str):
-                continue
-            content = self._clean_and_fix(content, plan)
-            entry = self._entry_name(name)
-            path = self.working_dir / entry
-            path.write_text(content, encoding="utf-8")
-            stage_scripts[name] = str(path)
-            steps.append({"name": name, "entry_file": entry, "script": content})
+        # Split into per-phase stages, retrying until each stage validates. The
+        # split is what resolves the monolithic deck into single-integrator
+        # phases; the transient monolithic deck is never run, so validity is
+        # based on the STAGES here, not on that deck. Retrying with the concrete
+        # validation errors makes a categorical slip (e.g. a stage that kept two
+        # integrators) self-correct engine-neutrally instead of being papered
+        # over per engine.
+        stage_scripts: Dict[str, str] = {}
+        steps: List[Dict[str, str]] = []
+        stage_errors: List[str] = []
+        prompt = base_prompt
+        for _ in range(3):
+            try:
+                stages = self._generate_json(prompt)
+            except Exception:
+                stages = {"production": full_script}
+            stage_scripts, steps, stage_errors = {}, [], []
+            for name, content in stages.items():
+                if not isinstance(content, str):
+                    continue
+                content = self._clean_and_fix(content, plan)
+                entry = self._entry_name(name)
+                path = self.working_dir / entry
+                path.write_text(content, encoding="utf-8")
+                v = self._validate(str(path), info, plan)
+                if not v.get("valid", True) and v.get("errors"):
+                    stage_errors.extend(f"[{name}] {e}" for e in v["errors"])
+                stage_scripts[name] = str(path)
+                steps.append({"name": name, "entry_file": entry, "script": content})
+            if not stage_errors:
+                break
+            prompt = base_prompt + (
+                "\n\nThe previous split still failed validation — fix these, "
+                "keeping ONE integrator per stage:\n- " + "\n- ".join(stage_errors)
+            )
 
         shared = self._campaign_shared_files(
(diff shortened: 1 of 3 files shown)


ziatdinovmax (changes requested):
It looks like the root-cause story in the description/comments does not match the code. Can you please correct it before merge and also add the test?

## Requested changes

1. **Root cause as written is wrong — fix the description and the two code comments.** Nothing consumed the monolithic deck's `validation`: `simulation_pipeline._generate_classical_md_inputs` does `setdefault("status","success")` and never reads `result["validation"]`; the pre-run `InputValidator` sees `input_files` = stage 0; `_collect_stages` runs the split stages. So the monolithic deck never "gated" anything. What actually happened: a bad split (a stage keeping both integrators, or `${visc}`) went straight to the engine and the refinement loop could not repair it → `refinement_failed`. The PR's real contribution is *adding* a per-stage structural check + re-split retry that did not exist — and, after three misses, a *new* hard fail at generation. Say that; drop "superseding the transient monolithic deck's checks" / "its validation gated the whole campaign".

2. **Add the test.** "Verified with mocked splits" is in the description but `changedFiles=2` — no test. A `MDSimulationAgent.__new__` + `patch.object(_generate_json/_get_skill_context/generate_simulation)` test against the real `lammps.validate_script` covers all three branches (clean / self-corrects / persistent → status error) in ~40 lines; I have one written and can push it to the branch.

## Suggestions (non-blocking)

- **Last-resort repair before `status="error"`.** The new hard fail is a behaviour change: previously invalid stages still reached the dry-run gate + LLM fixer in `run_campaign`. Since `_attempt_fix(script, errors, plan)` is already engine-neutral and used on the monolithic deck, run it per failing stage after the third split miss, re-validate, and only then set `status="error"`. Cheap, and matches the description's "split retried/validated → runs" table.
- **Stale stage files.** When a retry renames stages (`minimize` → `equilibration`), the earlier attempt's `run_minimize.lammps` stays in `working_dir` (verified). Harmless for the campaign — `stages` is authoritative — but confusing in the session dir; unlink the previous attempt's files at the top of each retry.
- **Wasted call.** `generate_simulation` still runs `_validate` + one `_attempt_fix` on the monolithic deck, whose verdict is then discarded. Fine to leave; worth a one-line comment so nobody "fixes" it back into a gate.
- **Skill nit.** The `fix ave/time … v_pxy v_pxz v_pyz` snippet references `v_pxy` without the `variable pxy equal pxy` definitions; add them or the deck errors on `v_pxy`.

sarah-allec:
You're right on all counts — thanks for the careful read. Fixed in b4ff74f5:

**1. Root-cause corrected (blocking).** Confirmed nothing consumes the monolithic deck's `validation` — `_generate_classical_md_inputs` setdefaults `status="success"` and the pre-run `InputValidator` sees stage 0. Rewrote the description and both code comments: the actual failure was a *bad split* reaching the engine where the loop couldn't repair it, and this PR's contribution is *adding* the per-stage check + re-split retry (and a new hard-fail after the retries). Dropped the "gated"/"superseding" language.

**2. Test added (blocking).** `tests/test_md_staged_split.py` — `MDSimulationAgent.__new__` + patched `_generate_json`/`_get_skill_context`/`generate_simulation` against the real `lammps.validate_script`, covering clean / self-corrects-on-resplit / persistent→`_attempt_fix`-repairs / unrepairable→`status="error"`. (Went ahead and wrote it rather than take yours, but happy to swap if you prefer yours.)

**Suggestions, all taken:**
- **Last-resort repair before `status="error"`** — after the split retries, run `_attempt_fix` per still-invalid stage and re-validate before failing. Restores the fixer path; agreed the earlier hard-fail was a behavior change that removed a working recovery route.
- **Stale stage files** — unlink the prior attempt's files at the top of each retry.
- **Wasted monolithic `_validate`** — left as-is with a comment noting the staged path discards its verdict, so nobody turns it back into a gate.
- **Skill nit** — `fix ave/time` snippet now defines `variable pxy equal pxy` (etc.).

Ready for another look.

ziatdinovmax (approved):

PR #517, opened 2026-08-31
Scale-neutral simulation recommendations (#45)

### LINKED ISSUE #45: Add MD recommendation option
Now that we have an MD agent, we need to make sure that experimental analysis workflows can produce relevant recommendations of structures and research goals for MD based on the provided image/spectra analysis results and novelty assessment. 

To achieve this, we'll need to 
 1. Extend current [theory recommendation agent](https://github.com/ziatdinovmax/SciLink/blob/main/scilink/agents/exp_agents/recommendation_agent.py) to include MD. Currently it is limited to DFT only. Ideally, it should also be capable of figuring out which type of simulations (DFT or classical MD) is more relevant to the experimental results.
2. Expand and rename [DFT recommendation workflow](https://github.com/ziatdinovmax/SciLink/blob/main/scilink/workflows/dft_recommendation_workflow.py) to have both DFT and MD options. For MD, it should output a recommended structure and a research goal for the MD workflow from text-based analysis and novelty assessment results. 
3. Update the corresponding argument name in experimental analysis workflows  (e.g. DFT_recommendation -> theory_recommendation) as well as print statements throughout the workflow if needed.

ziatdinovmax:
## Summary

After analyzing experimental data, SciLink could recommend theory follow-ups — but only DFT ones: the recommendation agent's prompt, schema, and tool were all DFT-specific, so an analysis whose natural next step is molecular dynamics (coarsening, diffusion, melting) or molecular quantum chemistry still came back with DFT suggestions. Recommendations are now *simulation studies*: a structure to model, a research goal saying what to compute and which experimental question it answers, and an optional scale suggestion. The suggestion is deliberately non-binding — the analysis side recommends the **what**, and the simulation router keeps ownership of the **how** (final scale and engine), the same contract philosophy as `DeployedPotential`.

Closes #45. (Per the re-scoping note on the issue: items 2–3 of the original plan targeted a `workflows/` layer that no longer exists; this delivers the re-scoped item 1 + tool surface.)

---

**Technical details**

- `RecommendationAgent.generate_simulation_recommendations_from_text` with new prompt (`TEXT_ONLY_SIMULATION_RECOMMENDATION_INSTRUCTIONS`); per-recommendation schema adds `research_goal` (required in the prompt, optional in validation) and `suggested_scale`. The scale vocabulary is exactly the simulate router's: `periodic_dft / molecular_qc / molecular_dynamics` — notably `machine_learning_potentials` is NOT offered (an MLIP is a potential *source* chosen downstream of routing, #429); an invalid hint is dropped rather than passed on. `VALID_SCALES` is a mirrored literal (exp_agents must import without the optional sim stack) with a drift-guard test pinning it to `simulation_router.DEFAULT_SCALE_DESCRIPTIONS`.
- Tool surface: registered as `recommend_simulations` (analysis system-prompt guidance updated: MD-scale recommendations are goals to hand to simulate mode; DFT-scale ones still feed `run_dft_workflow` by `recommendation_index`). Back-compat throughout: `recommend_dft_structures` still dispatches via `functions_map` (not in the LLM schema), `generate_dft_recommendations_from_text` aliases the new method, recommendations persist under both `simulation_recommendations` and the legacy `dft_recommendations` record keys, and `run_dft_workflow` reads either.
- Sidecar moves to `simulation_recommendations/simulation_recommendations.json` under the analysis record's directory.

**Verification**

- Offline: 6 new tests (`tests/test_simulation_recommendations.py`) — validation incl. the MLIP-hint drop, back-compat method alias, schema/functions_map split, dual-key storage + sidecar, system-prompt consistency, and the router drift guard; the refine-interpretation consumer invariant (`_effective_full_result` routing) updated to the new name and passing.
- Live (Bedrock opus-4-8), 14/14: a DFT-shaped STEM dopant analysis → orchestrator picked `recommend_simulations` unprompted, 8 studies with substantive goals, 7/8 hinted `periodic_dft`; a dynamics-shaped in-situ coarsening analysis → 6/8 hinted `molecular_dynamics` with MD-shaped goals (heating-ramp premelting onset, coalescence dynamics) and sensible structure downscaling; and the router contract both ways — with engines declared for all three scales, the real simulate router routed the MD-scale recommendation to `molecular_dynamics`/lammps and the DFT-scale one to `periodic_dft`/vasp. Environmental note: on an engine-less machine the router correctly reports "no engines available"; the test declares fixture engines via `SCILINK_AVAILABLE_SOFTWARE`.


sarah-allec (changes requested):
Nice generalization — the scale-neutral framing and the "analysis recommends the *what*, the router owns the *how*" contract are the right call, and the back-compat work (dual record keys, the `functions_map` alias, the `VALID_SCALES` mirror + drift-guard test) is careful. One issue I'd like addressed before merge; the rest are non-blocking.

## Requesting changes

**`run_dft_workflow` doesn't guard against a non-DFT recommendation.**

The recommendations list is now mixed-scale, but `run_dft_workflow(analysis_id=..., recommendation_index=...)` only range-checks the index (`recs = simulation_recommendations or dft_recommendations`, then `0 <= recommendation_index < len(recs)`) and builds. If the index points at a `molecular_dynamics` recommendation — e.g. *"10×10×10 anatase TiO₂, 2% O vacancies for finite-temperature dynamics"* — it will silently generate VASP inputs for a thousands-of-atom supercell that's intractable in DFT. The system prompt tells the model to feed only DFT-scale recs to this tool, but nothing enforces it, so a mis-pick (or a caller selecting by index) fails silently — producing a bad artifact rather than an error.

Suggestion: when the selected recommendation's `suggested_scale == "molecular_dynamics"`, refuse with a message pointing the MD-scale goal to simulate mode (or, at minimum, warn). `molecular_qc` and a missing/None scale can still pass through. Small guard, but it closes the one place where a wrong routing yields a bad output silently instead of a clear error.

## Non-blocking

1. **Sidecar path moved but isn't dual-written.** The record keys are dual (`dft_recommendations` + `simulation_recommendations`), but the on-disk sidecar moved to `simulation_recommendations/simulation_recommendations.json`. External scripts reading the old `dft_recommendations/dft_recommendations.json` path break even though the record API didn't — intentional? A one-line confirmation either way.
2. **`research_goal` is required in the prompt but optional in validation.** A recommendation that comes back without one still passes, then has nothing to hand to simulate mode; consider falling back to `description`.
3. Nit: `recommendation_agent.py` and `instruct.py` still end without a trailing newline.

Otherwise this LGTM — happy to re-approve once the `run_dft_workflow` guard is in.

ziatdinovmax:
Thanks for the careful review — all addressed in eba3c2a.

**Blocking — MD-scale guard:** `run_dft_workflow` now refuses a recommendation with `suggested_scale == "molecular_dynamics"`, returning an error that carries the recommendation's description/goal/scale and points to simulate mode; `molecular_qc` and unscored recommendations still pass through, and the tool description states the contract. Covered by two new tests (refusal + pass-through with a stubbed pipeline).

**Sidecar path:** intentional — the record keys (`simulation_recommendations` + `dft_recommendations`) are the back-compat surface; the sidecar is a session artifact named for what it now holds, and dual-writing it would leave a stale-looking `dft_recommendations/` copy to drift.

**`research_goal`:** now falls back to `description` during validation, so the simulate-mode handoff never carries an empty goal (test updated).

**Trailing newlines:** fixed.

sarah-allec (approved):
Verified `eba3c2a`: the MD-scale guard in `run_dft_workflow` is exactly right — it refuses `molecular_dynamics` with an error that carries the description/goal/scale and points to simulate mode, while `molecular_qc` and unscored recommendations still pass through. The `research_goal` fallback closes the empty-goal handoff, and the sidecar single-write rationale makes sense (dual-writing would just leave a stale `dft_recommendations/` copy to drift). Thanks for the quick turnaround!

PR #522, opened 2026-08-31
Per-tool checkpointing, informed restore, and resumable fan-outs

ziatdinovmax:
Crash-and-resume support across all four chat orchestrators, extensively live-validated on Bedrock (8 kill/restore scenarios, 62 live checks).

**What changes for users:**

1. **A crashed or killed session loses at most one tool call of work.** Previously the session saved only at the end of a turn — and one turn can contain an entire analysis, delegation, or parallel fan-out, so a mid-turn crash lost all of it. Every orchestrator now saves after each completed tool step (quietly, throttled so it costs nothing noticeable).

2. **A restored session knows where it stopped.** On restore after an interruption, the model is told which step completed last and continues from the recorded results instead of redoing work. Live: a session killed right after a curve fit finished was restored and reported the fit (R² ≈ 0.998, both peaks) without re-running anything.

3. **Interrupted parallel analyses are resumable.** If a session dies mid-fan-out, restoring it now shows a notice, and asking to "finish the interrupted analysis" resumes only the unfinished branches — in their original folders, with completed branches untouched — then fusion proceeds as usual. This works even if the resume itself is interrupted, and for harmonized runs (the resumed branch still replays the donor's script verbatim).

4. **A runaway tool result can no longer break a session.** Found during live testing: one tool echoed a 7.7 MB file into the conversation, making every later model call fail. Oversized results are now truncated with a recovery note (the tool's saved files and stored state are unaffected); normal results are untouched byte-for-byte.

---

**Technical details**

Implements `phase3_checkpoint_resume_remap_plan.md` (the remapped Phase 3 of the fan-out plan): resume as shared orchestrator-level infrastructure, with nothing added below the orchestrators (foundation-agent pipelines keep resume-by-re-derivation via the script bank / locked replay).

- **Per-tool checkpointing (`f7f1d655`).** `_checkpoint_after_tool()` after each tool iteration in both chat loops of all four orchestrators (copy-shaped per the no-base-class rule), mirroring the MCP server's `_checkpoint_after_tool`. Quiet (`quiet=`/`verbose=False` on `_auto_checkpoint`; simulate gained `force=` past its N-message gate) and wall-clock-throttled (15 s; an expensive call always lands one). Invariant: persist points follow tool results, so mid-turn saves are pair-complete — no dangling `tool_use` reaches disk from this path.

- **Informed continuation (`385166db`).** `close_interrupted_turn`'s stub (already at all eight turn-start seams) now names the last completed tool call(s) and instructs continuation from recorded results; healthy histories byte-identical, malformed `tool_calls` fall back to the generic note.

- **`resume_fanout` (`62baea87`).** New meta tool + restore-time notice. Reopens ledger entries left `running`/`interrupted` (the turn-start sweep converts the former to the latter; budget-degraded `timed_out` verdicts stay final), rebuilds each branch child in its original `fanout/<NN>_<slug>/` dir with `restore_checkpoint=True`, and re-runs the recorded pre-meshed task verbatim (+ resume note) through `_run_one_branch` — budget/cancellation shared via the extracted `_cancel_overdue_branches`. Three persistence gaps found by the live drivers are fixed here: provisional entries checkpoint at fan-out launch (previously a crash before the first branch completed left no on-disk trace), harmonize-mode follower tasks + `harmonized_with` stamps checkpoint before followers run (previously a crash mid-replay resumed them silently unharmonized), and resume checkpoints its reopened entries so it converges under repeated interruption.

- **Tool-result bounds (`a4096c86`).** `clip_tool_result` (200 k chars, recovery marker) at the message-build chokepoints of all four orchestrators — `build_tool_message` clips the text part only, images unaffected — plus `load_metadata`'s echo capped at 20 k (`_bounded_metadata_echo`; full metadata still stored). Root cause of the one live failure observed during the edge sweep (5.3 M-token prompts, durable via per-tool checkpoints).

**Live validation (Bedrock, all via SIGKILL harnesses):** Phase A kill/restore 8/8; Phase B post-fit kill 6/6; Phase C fan-out crash/resume 10/10; edge sweep — multi-turn mid-kill 8/8, simulate-orchestrator copy 7/7, partial fan-out resumed through the meta LLM's own tool choice 7/7, double kill (resume killed mid-resume) 9/9, interrupted harmonized replay 7/7.

**Offline:** 34 new tests (`test_per_tool_checkpoint`, `test_resume_fanout`, `test_tool_result_clip`, plus interrupted-note cases in `test_stop_turn_alternation`); several assert the on-disk checkpoint state from *inside* a running branch. Full 149-test sweep over the touched suites green.

### DIFF
--- a/scilink/agents/exp_agents/analysis_orchestrator.py
+++ b/scilink/agents/exp_agents/analysis_orchestrator.py
@@ -13,6 +13,7 @@ import inspect
 import json
 import logging
 import os
+import time
 from pathlib import Path
 from typing import List, Dict, Any, Optional
 from datetime import datetime
@@ -1613,7 +1614,7 @@ class AnalysisOrchestratorAgent:
             result["error"] = error_msg
         return result
 
-    def _auto_checkpoint(self):
+    def _auto_checkpoint(self, quiet: bool = False):
         """Internal auto-checkpoint without LLM interaction."""
         try:
             checkpoint_data = {
@@ -1634,10 +1635,37 @@ class AnalysisOrchestratorAgent:
             with open(self.checkpoint_path, 'w', encoding="utf-8") as f:
                 json.dump(checkpoint_data, f, indent=2)
 
-            print(f"    ✅ Auto-checkpoint saved")
-            
+            if not quiet:
+                print(f"    ✅ Auto-checkpoint saved")
+
         except Exception as e:
             logging.warning(f"Auto-checkpoint failed: {e}")
+
+    # Wall-clock throttle for the per-tool-call checkpoint: a burst of cheap
+    # tool calls must not thrash the disk, while an expensive call (a whole
+    # run_analysis) always lands one because it outlasts the window.
+    TOOL_CHECKPOINT_MIN_INTERVAL_S = 15.0
+
+    def _checkpoint_after_tool(self) -> None:
+        """Persist history + state after a completed tool iteration.
+
+        The chat loop used to persist only at turn end, so a crash mid-turn
+        lost every completed tool call since the last turn boundary — and one
+        turn can contain a whole run_analysis. Mirrors the MCP server's
+        checkpoint-after-tool. Quiet and throttled; a mid-turn history ends
+        in a dangling tool_use pair at worst, which the loop's
+        repair_dangling_tool_calls heals on the next load.
+        """
+        now = time.monotonic()
+        if (now - getattr(self, "_last_tool_checkpoint_ts", 0.0)
+                < self.TOOL_CHECKPOINT_MIN_INTERVAL_S):
+            return
+        self._last_tool_checkpoint_ts = now
+        try:
+            self._save_history()
+            self._auto_checkpoint(quiet=True)
+        except Exception as e:  # noqa: BLE001 - persistence must not kill the turn
+            logging.warning(f"Per-tool checkpoint failed: {e}")
 
     def _tool_message(self, tool_call_id: str, result: str) -> dict:
         """Build the tool-result message. When the active provider renders
@@ -1748,7 +1776,9 @@ class AnalysisOrchestratorAgent:
                 result = self.tools.execute_tool(func_name, **args)
                 
                 self.messages.append(self._tool_message(tool_call.id, result))
-        
+
+            self._checkpoint_after_tool()
+
         self._last_chat_hit_iter_cap = True
         return "⚠️ Maximum tool iterations reached. Please simplify your request."
 
@@ -1913,7 +1943,9 @@ class AnalysisOrchestratorAgent:
                 result = self.tools.execute_tool(func_name, **args)
                 
                 self.messages.append(self._tool_message(tool_call.id, result))
-        
+
+            self._checkpoint_after_tool()
+
         self._last_chat_hit_iter_cap = True
         return "⚠️ Maximum tool iterations reached. Please simplify your request."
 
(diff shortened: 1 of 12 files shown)


ziatdinovmax:
Follow-up commit e2c0e4f1: persist the provisional ledger entry when a single delegation opens (the fan-out launch checkpoint's missing twin). A crash mid-delegation previously left no on-disk trace of the delegation — the restored ledger showed it as never having existed instead of interrupted. Live-validated on Bedrock (meta SIGKILLed mid-delegation → restored ledger shows the entry → sweep surfaces the interruption → chat-driven recovery completes through the persistent child, 7/7 checks). Related follow-up filed as #523 (artifact-level salvage for killed pipeline runs).

sarah-allec (approved):
Approving — the crash-safety core is solid, and I checked the parts that would be scary to get wrong:

- **No dangling `tool_use` reaches disk.** `_checkpoint_after_tool()` lands *after* the result-append loop in all four orchestrators (both handlers), so saves are pair-complete; `repair_dangling_tool_calls` before the first API call is a clean backstop, and a mid-loop exception never checkpoints.
- **Ledger transitions are correct.** `resume_fanout` reopens only `running`/`interrupted` fan-out entries, leaves `success`/`error`/`timed_out` final (no re-run, no double work), and converges under repeated interruption (reopened entries persisted before branches start).
- **The `e2c0e4f` single-delegation provisional entry** is correct and consistent with the fan-out launch checkpoint — and correctly *not* auto-resumed (it isn't a fan-out).
- **`clip_tool_result`** is applied at all four chokepoints, clips the text part only (images intact), and has no mid-JSON parse hazard (clipped text is only read by the model, with a recovery marker).
- Cross-orchestrator consistency is good; the `quiet=` / `verbose=` / `force=` differences are intentional per each class's `_auto_checkpoint` signature.

One follow-up I'd like handled (not blocking this approval — fine as a follow-up issue):

**Harmonize resume can silently drop harmonization when the crash lands during the donor phase.** Followers are only re-stamped with the verbatim-replay task + `harmonized_with` *after* the donor completes (the post-donor `_auto_checkpoint` in `run_fanout` — good, and its comment names exactly this risk). But if the coordinator dies after the launch checkpoint and before the donor finishes, the followers' persisted task is still the pre-harmonize independent one. On restore, `resume_fanout` flat-reopens every stale entry and runs them concurrently with their recorded tasks, so the followers run **unharmonized**, with no `harmonized_with` stamp, and there's no donor-first pipeline replay to reconstruct it — the user asked for harmonized and silently gets independent results. Narrow window and degraded-not-corrupt (the branches still produce valid independent analyses), but silent. Suggestion: when `resume_fanout` reopens a harmonize fan-out whose donor entry didn't complete, either reconstruct the donor→stamp→followers pipeline, or surface a notice that harmonization was lost (or force a fresh re-run of that fan-out) — anything that makes it non-silent.

Minor:

- The "loses at most one tool call" / "last tool persists" framing isn't quite true for a *cheap* final tool: the 15 s throttle can skip its checkpoint and it's lost on crash. It self-heals (re-run on reload), so it's a coverage limitation, not corruption — worth softening the claim.
- Pre-existing, low: the child-orchestrator checkpoints write via `open()` + `json.dump` (no temp-file + rename), so the per-tool cadence widens the torn-write window. Loads are try/except-guarded, so it degrades gracefully — a temp+rename would close it.

Nice work — the pair-complete invariant and the resume idempotency are the hard parts, and they're right.

PR #607, opened 2026-09-10
fix: re-probe unavailable software on cache load so new installs are seen

jenna1701:
# fix: re-probe unavailable software on cache load so new installs are seen

## Problem

The simulation agent reports an MLIP backend unavailable even after the user installs it. Live case: `pip install fairchem-core` succeeds, `import fairchem.core` works, yet the agent keeps saying UMA is unavailable.

## Root cause

`AvailableSoftware.auto()` loads the cached `~/.scilink/available_software.yaml` and returns it unchanged whenever the file exists — it never re-probes. A cache built before the install recorded `uma: available: false`, so the later install is invisible until a manual refresh. The live probe was correct (`fairchem.core` importable → `True`); only the stale cache was wrong. (The agent's "`pip install fairchem`" suggestion is a separate hallucination — the stored guidance text already says the correct `fairchem-core`.)

## Fix

On load, re-probe the entries that are **not yet available** and persist any that flipped to available (`_reprobe_unavailable`). This is cheap:

- A still-missing module's `importlib.util.find_spec` returns `None` without importing, so only a newly-installed package pays an import — once — after which it caches `available: true` and is skipped.
- Entries already `available: true` or `user_confirmed: true` are left untouched: trusted, and re-importing heavy MLIP packages on every startup is exactly the cost the cache exists to avoid. A stale *uninstall* surfaces as a clear import error at use time — cheaper than re-probing every installed backend on every startup.
- The re-probe is wrapped so a probe failure never breaks the load hot path.
- Also catches a newly-added skill bundle absent from an older cache, not just a newly-installed package.

## Verification

The stale cache self-healed on the next `auto()`: `uma` false → re-probed → `true` → persisted (YAML now records the `fairchem/core` origin). No manual `refresh`/delete needed.

## Tests

`tests/test_available_software_reprobe.py` — heal+persist, skipping trusted/user_confirmed entries, and still-missing staying false.

### DIFF
--- a/scilink/agents/sim_agents/mlip_agent.py
+++ b/scilink/agents/sim_agents/mlip_agent.py
@@ -368,6 +368,7 @@ Return JSON:
             elements=elements,
             working_dir=self.working_dir,
             device=sim.get("device", "cpu"),
+            task_name=sim.get("task_name"),
         )
 
         # Delegate run generation to MDSimulationAgent. MLIPAgent's job
--- a/scilink/skills/_shared/mlip_tools.py
+++ b/scilink/skills/_shared/mlip_tools.py
@@ -207,23 +207,29 @@ def deploy(
     elements: List[str],
     working_dir: str,
     device: str = "cpu",
+    task_name: Optional[str] = None,
 ) -> DeployedPotential:
     """
     Deploy an MLIP model as an engine-neutral DeployedPotential.
 
     ``model`` is a model *identifier* — each backend's deploy function
     decides whether it's a foundation-model keyword (``"mace-mp-0"``,
-    ``"chgnet"``) or an on-disk path to a trained/fine-tuned model
-    (the output of ``train()`` / ``fine_tune()``). This single entry
-    point covers both cases so adding a backend touches exactly one
-    deploy function — no separate pretrained/trained code paths.
+    ``"chgnet"``, ``"uma-s-1p2"``) or an on-disk path to a
+    trained/fine-tuned model (the output of ``train()`` /
+    ``fine_tune()``). This single entry point covers both cases so
+    adding a backend touches exactly one deploy function — no separate
+    pretrained/trained code paths.
 
     Args:
-        backend:     "mace" | "chgnet" | "nequip" | "deepmd"
+        backend:     "mace" | "chgnet" | "uma" | "nequip" | "deepmd"
         model:       Foundation-model keyword or path to a trained model
         elements:    Chemical elements in the system
         working_dir: Output directory
         device:      "cpu" or "cuda"
+        task_name:   Prediction-head selector for multi-task foundation
+                     models (UMA: "omat", "omol", "oc20", ...). Ignored
+                     by single-task backends. Defaults per backend when
+                     None.
 
     Returns:
         A DeployedPotential descriptor — the engine-neutral contract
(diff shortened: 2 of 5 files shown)


jenna1701:
This branch now carries two related fixes for the fairchem/UMA path. First, AvailableSoftware.auto() re-probes not-yet-available engines when it loads a cached YAML, so a package installed after the cache was built (the fairchem-core / UMA case) is detected and persisted instead of staying invisible until a manual refresh; available and user-confirmed entries are left untouched. Second, mlip_tools.deploy() gains a real UMA branch: the backend previously had an availability probe but no deploy dispatch, so any attempt to deploy UMA fell through to Unknown backend: uma even though check_backends() reported it available. The new _uma_deploy() builds the calculator through the correct shipping fairchem API  (pretrained_mlip.get_predict_unit(name, device=...) then a positional FAIRChemCalculator), adds an optional task_name head selector (default "omat", "omol" for molecules), and the stale checkpoint_path= snippet in uma.md is corrected to match. Together they fix both halves of the reported problem — the agent now sees UMA as installed and can actually deploy it.

ziatdinovmax:
Conceptually this is the simulation-side counterpart to #581 (image exec-env probing): both make the agent's capability claims match the real environment rather than assumption. Notably #581 sidesteps the stale-install problem by caching per-interpreter in-process, whereas this PR retrofits that safety into the persistent `available_software.yaml` cache.

sarah-allec (approved):
## PR #607 Review

**TLDR:** Two fixes for the same user-visible bug: "I installed fairchem-core but the agent still says UMA is unavailable." The cache-heal logic is well-scoped and safe. The UMA deploy function correctly fills a real gap — `check_backends()` already reported UMA available but `deploy()` had no branch for it. One minor nit on the generated `construct_expr`. Approve.

---

### Commit 1: `_reprobe_unavailable` in `AvailableSoftware.auto()`

On load, re-probes entries marked `available: false` and persists any that flipped. Leaves `available: true` and `user_confirmed: true` untouched.

The asymmetry is the right call — re-probing unavailable entries is cheap (`find_spec` without importing), while re-probing available entries would import heavy MLIP packages on every startup. A stale uninstall surfaces as a clear import error at use time.

The `except Exception` wrapping is appropriate — breaking the hot path over a probe failure would be worse than the stale cache it's trying to fix.

Tests cover the three key behaviors (heal-and-persist, skip trusted, leave still-missing). Minor gap: no test for the "newly-added skill bundle absent from cache" path mentioned in the docstring.

### Commit 2: `_uma_deploy` in `mlip_tools.py`

`check_backends()` on main already reports UMA available via `_probe_uma()`, but `deploy()` had no `"uma"` branch — every deploy attempt fell through to `raise ValueError("Unknown backend: uma")`. This commit fills that gap.

The fairchem API usage is correct: two-step `pretrained_mlip.get_predict_unit()` → `FAIRChemCalculator(predictor, task_name=...)`. The `uma.md` skill example is fixed to match. The `task_name` plumbing through `mlip_agent.py` is the right thread point.

**One nit — the `construct_expr` uses `__import__` inline** to get `pretrained_mlip`, while the `import_line` only carries `FAIRChemCalculator`. Cleaner to add `pretrained_mlip` to the `import_line` and reference it directly in `construct_expr`, matching how MACE does it (`import_line` carries the callable, `construct_expr` references it). Not a correctness bug — just inconsistent with the existing pattern.

**MPS:** `_uma_deploy` strips MPS (maps to CPU) while `_chgnet_deploy` keeps it. Probably correct for fairchem, but worth a comment if intentional.

---

**Verdict:** Approve. The `__import__` cleanup is a nit, not a blocker.

PR #642, opened 2026-09-15
Documents: append the workflow diagram only when the body has none (#639)

### LINKED ISSUE #639: Plan/document: workflow figure appears twice in a deliverable (inline embed + unconditional append)
## Summary

A generated `write_technical_document` deliverable can contain the **same workflow figure twice** — one embedded inline in the authored body and one appended by the workflow-diagram step, with neither checking the other. Observed in `meta_session_20260914_101841`, delegation 02 (`mg_selective_recovery_proposal.md`).

## Evidence

Two `` embeds of the same screening workflow:

- **Line 69** (in body): `Process and screening workflow…` — `mg_recovery_workflow.png` is a **byte-identical copy of `campaign_workflow.png`** (same md5 `10ea6376349a5673bac9a4a6525f08a1`; `deliverables.json` title: "Copied from campaign_workflow.png").
- **Line 251** (end): `Campaign workflow diagram` — a different PNG (md5 `7001046bd5cf96c5df501d31b3e6dd53`).

The two PNGs are not byte-identical and their `.mmd` sources differ in node labels, but they render the **same workflow** (same stages/decision gate/two-arm structure/colloid caveat), so a reader sees the figure twice — and under two different captions ("Process and screening workflow" vs "Campaign workflow diagram").

## Cause

Two diagram paths both fire, neither aware of the other:

1. **Inline:** the document author embeds a workflow figure into the body (referencing `mg_recovery_workflow.png`, created as a copy of the campaign-level `campaign_workflow.png`).
2. **Appended:** the workflow-diagram step (`_append_workflow_diagram`, `scilink/agents/planning_agents/orchestrator_tools.py:1311` — "Append a compact workflow diagram section to a document") generates a proposal-specific diagram and unconditionally appends it.

The append does not check whether the document body already embeds a workflow figure, so the deliverable ends up with a redundant one.

## Fix direction

Make the workflow-diagram append idempotent: before appending, detect an existing workflow-figure embed in the document (an `` / `*_workflow.png` reference, or a prior appended workflow section) and **skip or replace** rather than add a second. Alternatively, pick one owner of the workflow figure — either the authored body OR the append step, not both.

## Scope note

This is deliverable-content quality (the `.md`/`.html` a user reads), independent of the separate figure-rendering issues (#636 currency-as-math) and the artifact-sweep embedding (#633).

ziatdinovmax:
## Summary

Closes #639.

A technical document could show the same workflow figure twice: once where the author placed it in the body, and once more at the end where the workflow-diagram step appended its own rendering. The append step now checks first and leaves the document alone when it already carries a workflow figure. Documents without one still get the diagram exactly as before.

## Technical description

`_maybe_embed_workflow_diagram` (used by `write_technical_document` on new documents and by the white paper) gains an `_existing_workflow_figure` check that runs before any rendering, so a skipped append also saves the diagram-agent call. Three signals, any of which counts:

- a prior appended `## Campaign Workflow` section (a second pass over the same text);
- an embedded image whose alt text or filename contains "workflow" (the live case: `Process and screening workflow…`);
- an embedded image byte-identical to a `campaign_workflow*.png` already in the output directory (a copy under an unrelated name).

Unrelated figures (a micrograph, a cost chart) do not block the append. The skip prints one line naming the figure it found. The revision path already skipped the append; the memo path never appends.

Tests: `tests/test_workflow_figure_not_duplicated.py` drives the helper with a stubbed diagram agent: inline workflow figure suppresses the append and the render; a byte-identical copy under another name counts; a document without one gets the section once and a second pass adds nothing; unrelated figures still get the diagram. The document, image-embedding and diagram-agent suites pass (39 tests).

No live run: the change is confined to the helper the tests drive directly, and whether the author embeds a figure inline is the model's choice, so a live document cannot be made to reproduce the duplicate on demand.

Co-Authored-By: Claude <noreply@anthropic.com>


ziatdinovmax:
Live verification (Bedrock Opus 4.8, real Mermaid renderer), one planning session:

- `generate_initial_plan` → `generate_workflow_diagram` rendered `campaign_workflow.png`.
- `write_technical_document` asked to embed that figure inline: the author did (`Process and screening workflow`), the append step printed "Workflow figure already in the document (…) — not appending a second one", and the saved proposal carries exactly one workflow figure and no appended "Campaign Workflow" section.
- Control: a roadmap authored without figures got exactly one appended `## Campaign Workflow` with `staging_roadmap_workflow.png`.

Both the reproduced live shape and the unchanged default path behave as intended.

PR #645, opened 2026-09-15
read_file search: page through matches, say where they are, never claim exhaustiveness (#644)

### LINKED ISSUE #644: read_file search silently caps at first 40 matches (front matter on large docs); tool descriptions wrongly say 'returns every matching line'
## Summary

`read_file` in search mode silently returns only the **first 40 matching lines** of a file (`SEARCH_HIT_CAP = 40`, `scilink/utils/file_io.py:49`), taken by line position (`hits[:SEARCH_HIT_CAP]`). On a large document this surfaces the **top of the file** (front matter) and hides the rest, while three of the four orchestrators' tool descriptions claim it "**Returns every matching line**" — so the model believes a search was exhaustive when it saw ~4% of its own matches. Observed in `meta_session_20260914_101841`.

## Evidence

The planning agent searched the uploaded 267-page DOE Critical Materials Assessment PDF:

- `read_file(doe-...-assessment.pdf, search="magnesium|critical material|near-critical|copper|aluminum")`
- result: `"matches": 1033`, `"total_lines": 22629`, `content` = `"1033 matching line(s) in 22629 total; showing the first 40 ..."` — the 40 shown were the document's front matter (title, foreword, intro at lines 1–~130).
- The agent did not narrow or paginate; it took the 40 front-matter lines and then asserted in the deliverable that "Magnesium — CRITICAL (red), **confirmed verbatim in the DOE PDF**" — a claim the retrieved text does not support (the criticality designation lives deeper in the PDF; the real confirmation was the separately-viewed `criticality_matrix.jpg`).

So 993 of 1033 matches, and essentially the whole 22,629-line document, never entered context, yet the source was cited as if consulted.

## Root cause

`scilink/utils/file_io.py:128-139` (`window_lines`, search branch):
```python
hits  = [i for i, ln in enumerate(lines) if rx.search(ln)]
shown = hits[:SEARCH_HIT_CAP]            # first 40 by line position
...
note = f"{len(hits)} matching line(s) in {total} total" + (f"; showing the first {SEARCH_HIT_CAP}" if len(hits) > SEARCH_HIT_CAP else "")
```
Two problems:
1. **First-40-by-position** means a broad pattern over a big file returns only the head of the file, not a representative or dense-region sample, and there is **no way to page to matches 41+** (the `offset` param windows raw lines, not match ranks).
2. **The truncation note is weak and easy to glide past**, and it is **contradicted by the tool descriptions**: `scilink/agents/planning_agents/orchestrator_tools.py:6525`, `scilink/agents/exp_agents/analysis_orchestrator_tools.py:6283`, and `scilink/agents/sim_agents/simulation_orchestrator_tools.py:2007` all say search "**Returns every matching line**". (The meta description at `meta_orchestrator_tools.py:2032` omits "every".)

## Fix direction

- **Fix the descriptions first** (cheap, high-impact): "Returns every matching line" → "Returns up to the first 40 matching lines (of N total); narrow the pattern or page to see the rest." Align all four orchestrators.
- **Make truncation actionable**: add a way to page through matches (a `match_offset` / next-cursor), and/or return the cap as a structured `truncated: true` + `next` field, not just prose the model can ignore.
- **Optionally distribute the shown matches** across the file (not pure first-40) so a broad pattern over a large doc does not degenerate to front matter; narrowing the pattern remains the better user-side move, so pair any change with guidance that says so.

## Related

Compounds the "no auto-embedding in planning" behavior: with no dense index over an uploaded PDF, `read_file` search is the primary way to consult it — so a silent 40-match cap means large authoritative documents are effectively read only at the top unless the agent iterates. Also an over-attribution risk (citing a source consulted only shallowly).

ziatdinovmax:
## Summary

Closes #644.

A `read_file` search that finds more than 40 matching lines used to hand back the first 40 by position with a soft aside, while the tool description promised "every matching line". On a 267-page PDF that meant the agent saw the front matter, believed it had read the document, and cited it. Now a capped search says plainly that it is truncated and partial, shows where in the file the remaining matches fall, and can be paged with `match_offset`; the descriptions in all four modes say the same thing and tell the model to narrow or page before citing.

### Engine (`scilink/utils/file_io.py`)

- `window_lines` / `read_file_content` take `match_offset` (1-based match rank). The search result carries structured `truncated`, `next_match_offset`, `shown`, `match_offset`, `match_lines` and `match_density`.
- Notice text on a capped search: `1033 matching line(s) in 22209 total; showing matches 1-40 of 1033 (lines 1-856). TRUNCATED: 993 more match(es), up to line 22026, are NOT shown — this is a partial result, not the document. Narrow the pattern to what you actually need, or continue with match_offset=41`, followed by `Where the matches are (equal line bands): lines 1-2221: 98 · lines 2222-4442: 132 · …`.
- Paging past the last match is explained (`match_offset=999 is past the last match (#200, line 996)`). A search under the cap is unchanged apart from the new fields; the head/offset/tail modes are untouched.

### Tools

All four `read_file` wrappers (planning, analysis, simulation, meta) take `match_offset`, forward it, and carry the same `search` and `match_offset` descriptions: up to 40 per call, total count and location, `truncated` / `next_match_offset`, and "a truncated search has NOT read the document — narrow the pattern, or page with match_offset, before citing it". "Returns every matching line" is gone from all of them.

### Tests

`tests/test_read_file_search_paging.py`: first page truncated with next page and a 10-band density map; paging walks every match exactly once and the last page is untruncated; paging past the end is explained; an under-cap search has no density noise; `match_offset` rides through `read_file_content`; every orchestrator describes the cap and forwards the parameter. The two older "showing the first 40" assertions were updated. `tests/test_read_file_offset_and_index.py` pointed at another user's checkout and a session that does not exist here; it now reconstructs its fixture, gives the exec-lifted wrapper its package identity for the relative import, and no longer calls `sys.exit` at import (which aborted pytest collection). 38 read_file / file_io tests pass.

### Live verification

Engine on the real documents: the DOE Critical Materials Assessment (267 pages, 22,209 extracted lines) with the issue's own pattern gives 1033 matches, `truncated: true`, `next_match_offset: 41`, a density map with hits in every tenth of the file, page 2 continuing at line 877, and the narrowed pattern `magnesium.*(critical|criticality)` landing on line 969 ("nickel, platinum, magnesium, SiC, and praseodymium become critical"). A USGS Mineral Commodity Summaries PDF gives 34 matches, untruncated.

Agent on Bedrock Opus 4.8 (planning orchestrator, both PDFs in the data folder, asked to start with the broad search): the model ran the broad search (1033 matches, truncated), then read around the last hit shown (`offset=857`) instead of stopping, and answered with a verbatim quote that really is at lines 855-857 ("The 2019 CMS report found seven materials to be critical in the medium term … and magnesium (Mg)") plus the correct USGS production figure with its line. It did not page or narrow further, so it reported the 2019 finding rather than the 2023 assessment's own statement at line 969; the truncation was visible to it and the citation is real, which is the failure the issue describes no longer happening.

Co-Authored-By: Claude <noreply@anthropic.com>


ziatdinovmax:
Full offline suite on this branch: 2757 passed, 23 failed, 26 errors — every failure and error is in the main baseline (environment-specific HPC / Materials Project tests and the import-time logging.disable ordering noise); no branch-only failures. This branch also un-breaks collection of `tests/test_read_file_offset_and_index.py`, which now runs under pytest (previously a checkout-specific path).

PR #656, opened 2026-09-19
Fast analysis profiles and a live measurement loop for 1D spectra

ziatdinovmax:
## Summary

SciLink's analysis agents are thorough, which makes them slow, and they could not run next to an instrument. This PR adds two things.

**Analysis depth you can choose.** `run_analysis`, the CLI (`scilink analyze --profile`) and Mission Control can now ask for a `quick` or `extract` analysis instead of the default thorough one, name the quantities they actually need (`targets`), and set a time budget. A curve analysis that took about two minutes takes about one under `extract`, and about two seconds when the script bank already holds a recipe for that kind of data. The thorough path is unchanged and remains the default.

**A live measurement loop for 1D spectra and, in a first version, for datacubes (spectrum images)** (`scilink.live`, and a new **Live** tab in the web UI). One reference measurement (or the first few frames, or a curve analysis you already did in the session) is analysed in full and its verified script is locked as the recipe. Every later frame is answered by that recipe in about a second, with no model call. The loop watches fit quality, flags frames the recipe no longer describes, and rebuilds the recipe in the background while frames keep being answered. Separately from the fit, it measures how much of each frame is unlike the frames seen so far, with no model. A lasting change is always announced, with where on the axis it happened (new, missing, shifted or broadened intensity), before anything is rebuilt or accepted. From there you can send that frame to Chat for a thorough analysis in one click, and where the experiment can wait, the run can pause at the change until you resume, change the parameters, or stop. An independent second analysis can audit the locked recipe in the background, on a schedule or when the data changes, because a fit can look perfect and still be the wrong model. The quantities you name are reported under those names whatever recipe is in use. Optionally a recommender proposes the next acquisition parameters (Bayesian optimization, a rule table, or a language model that takes a goal in words and is guided by a per-technique acquisition skill). SciLink recommends and never actuates: parameters change only through the `apply` policy you choose or the Accept button.

You can try it without an instrument. Four simulated experiments are included (beamline XRD through a phase transition, in-situ Raman through an anneal, AFM force curves crossing onto a stiff inclusion, STM dI/dV with a tip change), and a folder of recorded spectra can be replayed as a stream. Moving to a real instrument does not need SciLink-specific code: put a small MCP server in front of the instrument controller (any language, one tool that takes acquisition parameters and returns a measurement), connect it in the MCP tab and pick it in the Live tab. The parameters and their safe limits are read from the tool's own schema. A small Python class is the in-process alternative (`examples/live_loop_demo.py`).

**What is and is not validated.** Everything was tested live on Bedrock Opus 4.8 against the four simulators and against one real dataset (a line scan from the experimental low-loss EELS cube in `examples/`). The real data broke assumptions the simulators never touched (see below), which is the main reason this note exists: **more real data series are needed** (XRD, Raman, AFM, ideally each with a known change somewhere in the run) before the gate tolerance and the behaviour on real noise can be called settled. Live mode covers 1D curves and, in a first version, datacubes (section 6): the reference cube is analysed once, every later cube is answered by replaying its per-pixel script with no model call (about 1 s for a small simulated cube, 3 to 4 s for a 20 × 18 pixel tile of the real EELS cube with 1496 channels), the tracked quantities are the means of the maps it computes, and a change is reported with where on the spectral axis and in which part of the field it is. The only real-data test for cubes is one EELS field cut into tiles; real cube *series* are needed. Images are future work.

## Technical description

35 commits, no change to default behaviour of existing analyses. Design notes and the measurements behind each decision are in the commit messages.

### 1. Instrumentation and script bank (foundation)

- `_stage_timing.py`: `StageTimer` and `stage_timings` in `analysis_results.json` for all three analysis agents. `tracing.py`: always-on LLM call counters, with an `off_path` context so calls made by a background recommender are not charged to a frame.
- `_script_bank.py` hardening: failures recorded, independent evidence keyed on data (separate `evidence` and `verbatim_evidence` ledgers), hash verified on load, fingerprint version, skill mismatch as a hard filter for curves, reentrant domain lock for concurrent writers, aging and archiving (`scilink memory bank-stats | bank-sweep | bank-archived | bank-restore`; Memory panel shows archived counts), and an assist log of how the bank was used per item.
- `curve_fitting_tools.validate_bound_pinning`: a value far above a positivity floor is not "pinned" (false positive that cost a model call per frame).

### 2. Depth as a property of the request

- `_qc_profile.py`: presets `thorough | quick | extract | realtime` with downstream fields (plan validation, literature, adaptive refit, trend, synthesis level, tier 2, verification mode, time budget). `resolve_profile` accepts `{"base": ..., **fields}`.
- Purpose-scoped verification: with `targets`, the verifier judges what the consumer needs. Targets also scope the plan (`planning_addendum`).
- Soft run budget: deferrable stages are skipped, the result is returned flagged `unverified` with `stopped_by`. A capped verification loop stops at its cap.
- Bank first under `quick` / `extract`: an audition that clears the gate and a data-similarity bar is the recipe (zero model calls).
- Typed channel: `delegate_to_analysis(profile, targets, time_budget_s)` → `run_task` → `run_analysis` → agent. Fan-out branches carry their own depth.

### 3. `scilink.live`

- `measurement_loop.py` `MeasurementLoop`: `setup()` (reference file, list of first frames, or `anchor=` a prior curve-fit run), `step()` (strict zero-LLM replay through the same `reuse_locked_script` path a series uses after its anchor), `amend()`, `escalate()`, `status()`, `resume()`. `loop_log.jsonl` is the contract, `loop_state.json` the restart point.
- Two clocks. Fast: `step()` never calls a model (`strict_replay` fails the frame instead of repairing it). Slow: reference analysis, re-anchors (`python -m scilink.live._reanchor`, own interpreter, spec on stdin), pinning, LLM recommender (daemon thread, `SlowSlot`).
- Flags: `fit_failed`, `gate_poor`, `drift_suspected`, `out_of_reference_range`, `deadline_missed`, `llm_used`. Breach patience, auto-escalation, range gate with warm-up and adoption, watching the pinned outputs when they are declared.
- `pinning.py`: named outputs. The model writes only a code block, spliced mechanically before the results print, gated deterministically (names present and finite, nothing else changed) and then by an independent value review (see "final sweep" below). Re-done by the re-anchor worker so a new recipe reports the same names.
- `gates.py`: the fit gate is calibrated on the reference run in units of its own noise (residual scatter over MAD of second differences). Calibration only relaxes the constant bar, so clean data is judged as before. (The fingerprint-similarity calibration of the first version is gone with the fingerprint, see section 5.) `noise_gate_tolerance=1.25` is an exposed knob whose docstring states its evidence.
- `portability.py`: at setup the recipe is replayed on the reference scaled ×3 and ×0.35 (verdict) and with the x axis shifted 3 % (information only: follows features or fixes positions). Zero model calls.
- Planning for a stream: `analyze(stream_reference=True)` adds one planning principle. `setup(reference=[files])` plans from the first N frames as a series and locks on the last through a derived single-spectrum anchor. Re-anchors are planned from a window of recent frames (`reanchor_frames=5`), asking the bank about the newest frame first (`analyze(bank_only=True)`). (The `drift_rebased` step of the first version is replaced by the change monitor's own state memory, section 5.)
- `recommend.py`: `InstrumentSchema`, `finalize()` (validation by the loop: refuse, never clamp; protocols never executed; `requires_approval`; `hold` is a valid answer), `RuleTableRecommender`, `GPRecommender` (reuses `bo_tools.SingleObjectiveOptimizer`, no `BOAgent` change), `LLMRecommender` (experiment context from the loop's `system_info`, its own earlier decisions shown back to it, one retry on a non-JSON reply).
- `instruments.py`: `Instrument`, `Frame`, `run_experiment(apply="never|approved|valid", operator=...)`, `ReplayInstrument`, `EndOfData`. `simulators.py`: four simulators with per-frame truth.
- `mcp_instrument.py`: `MCPInstrument` (an instrument whose `acquire` is one tool of an MCP server, via the existing `MCPConnection`). Schema from the tool's `inputSchema`; a number with no declared limits is held at its default and never steered; replies as x/y arrays, a data table or a file path; optional `describe_instrument` tool. `mcp_demo_server.py` is a reference server. In the Live tab it uses a server already connected in the session, so nothing is imported or executed by SciLink.
- `acquisition_skills.py` + `skills/acquisition/{raman,powder_xrd,afm_force_curve,sts_didv}`: knowledge-only domain with its own vocabulary (`overview · tradeoffs · limits · quality · strategy`, declared in `loader._DOMAIN_VOCABULARIES`), selected through the shared selector (exclusive), per technique and never per instrument, excluded from the `run_analysis` menu. `CLAUDE.md` records the convention: live loops reuse the analysis skills and do not get copies of them.

### 4. Web UI

- `server/live_api.py` + `/api/v1/.../live` routes: a live run is a session-level background thread, one per session, under `<session>/live/run_NNN/`. The page polls a snapshot built by an incremental log reader (offset, running counters, last 600 frames). Custom instrument classes and folder replay are local-server only, and a named attribute is checked to be an `Instrument` subclass before it is called.
- `LivePanel.tsx`, `LiveChart.tsx`: minimal copy with info buttons (every flag explained, the frame deadline set under Options, unordered x drawn as points), open-ended runs (frame count optional), reference from first frame(s) or a past analysis in the session, latest frame with the fitted model over it, one small chart per tracked quantity, operator parameter changes, recommendation card with Accept, collapsed event log. Hand-rolled SVG, no new dependency, two series colours validated for CVD separation and contrast on both theme surfaces.

### 5. Added after the first review: change signal, audits, novelty, pause

Four commits on top of the reviewed state (`134d52d1`, `33be5a7d`, `8015233d`, and the `recommend_now` fix).

- **Three separate judgements per stream.** (1) Does the recipe fit this frame: the noise-calibrated fit gate. (2) Has the data changed: `live/drift.py` `DriftMonitor`, a graded, model-free signal that replaces the thresholded peak-counting fingerprint the review found saturating on rich patterns. It keeps an SVD subspace of accepted frames (plus a constant and a slope) and reports `fraction` (noise-debiased share of a frame outside the subspace, 0 to 1) and `score` (relative to the stream's typical residual-to-noise ratio), with a coherent test over runs of 3 and 6 frames for changes too weak per frame. A single glitch is judged alone. A partial measurement window is compared on the overlap. Because it reads only the data it cannot depend on which recipe was locked. (3) Is a recipe that fits also right: no gate can say, so `audit()` runs the re-anchor worker with a different adoption rule and compares named outputs (`audit_every`, `audit_profile="quick"`, `audit_tolerance=0.05`). Two analyses are never asked to agree better than the output's own frame-to-frame scatter, or three combined uncertainties.
- **A lasting change is announced, located, and never quietly absorbed.** After the usual patience the loop emits one `novelty` event per change: since which step, `fraction`, `score`, distance from the reference, and `where` from `DriftMonitor.locate()` (robust Huber residuals, kinds `new | missing | shifted | broad`), whether the recipe still fits, and the frame. The recommender is told and asked at once (`Recommender.notify`, `urgent`). What happens next is the `on_change` policy when the recipe still fits: `report` (default: accept the state for tracking once the changed frames agree, no model call, `state_accepted`), `audit`, or `rebuild`. A recipe that fails is always rebuilt. Accepted states are remembered so the same kind of region is not asked about twice.
- **A change can be a decision point.** `run_experiment(pause_on=("novelty", "breach"), on_pause=...)`: the frame that made a change lasting carries `record["novelty"]`, the driver asks the instrument to hold (`Instrument.pause()` / `resume()`, mapped by `MCPInstrument` onto optional `pause` / `resume` tools; `can_pause` says whether the experiment is really held or only the acquisition), waits for `on_pause(event, record)` and resumes with its answer: resume, stop, or schema-checked parameters. `on_pause` is required (a pause nobody can end is a hang), pausing is off by default, and it is recorded as `paused` / `resumed` through a public `MeasurementLoop.record_event`. SciLink still never actuates outside this driver.
- **Two layers of novelty.** The loop's novelty is statistical. Whether it is scientifically new is the question of the first SciLink paper's pipeline (arXiv:2508.06569: observation → falsifiable claims → novelty scored against the literature → follow-up), which `assess_novelty` still asks in analyze mode. The Live tab's "Analyse in Chat" hand-off sends the frame (absolute path) with where the change is and asks for exactly that chain. It is one click for a person and never automatic; running it automatically during a pause is not built.
- **Instrument identity.** `Instrument.describe()` (id, name, technique, kind, `can_pause`; an MCP server supplies `id` through `describe_instrument`), stamped on the setup record and in `loop_state.json`. A test guards that `scilink.live` imports nothing from the server, chat or orchestrators. `CLAUDE.md` records the direction: the live layer is written so it can later be organised per instrument instead of per chat session.
- **User notes.** A free-text field in the Live tab travels as `system_info["notes_from_the_user"]` to every model-driven stage. It informs and does not constrain.
- **Web.** Snapshot gains `novelties`, `last_audit`, `paused`, instrument `id` / `can_pause`; `POST /live/resume` carries the pause decision (`pause_timeout_s` resumes unchanged). Live tab: a Change chart, novelty cards with "Analyse in Chat", an audit-disagreement banner, `on_change`, audit interval, Pause setting, a paused banner with Resume / Resume with changes / Stop. Strict-replay frames write no HTML report, and the chat artifact sweep ignores `live/` (50 per-frame reports had been listed as chat artifacts).
- New offline tests: `test_drift_monitor`, plus additions to `test_measurement_loop`, `test_instruments_and_simulators`, `test_mcp_instrument`, `test_web_measurement_loop`, `test_qc_profile_quick`.

Live results for section 5 (Bedrock Opus 4.8):

| run | result |
|---|---|
| Real EELS line | Leaving the nanocrystal announced and located (intensity gone from 0.62 to 0.73 eV), accepted for tracking with zero model calls. |
| Real EELS line + injected feature | Located at 0.795 to 0.935 eV, peak 0.867 (truth 0.86), marked as breaking the recipe, rebuilt. |
| STM dI/dV, tip change | Novelty from the event frame, located at −2.52 mV (truth −2.6). Recommender asked at once and reasoned from the location. |
| AFM, stiff inclusion | Novelty with the measured window at 52 % of before, accepted with no rebuild, values within 0.3 %. Recommender raised the trigger force. |
| Beamline XRD, phase transition | Followed with no rebuild and no model call, 88 of 90 frames clean. |
| In-situ Raman, periodic audits | Audit and recipe agreed to four digits. |

**Pause, live** (AFM behind the reference MCP server, which has `pause` / `resume` tools): paused at frame 27 for a change since frame 25, the instrument held for 113 s, a quick analysis of the new frame during the pause claimed a stiff, near-linear contact (3.24 nN/nm) consistent with an inclusion, the recommender raised the trigger force 20 → 30 nN under the stated 45 nN limit, and the run resumed with that through MCP. 37 of 40 frames clean, no model call on any frame, `paused` / `resumed` on the log with the decision. The first live attempt resumed unchanged and exposed a gap: a slow recommender's answer is collected by the next `step()`, and a pause has none. `MeasurementLoop.recommend_now()` fetches and validates it at the decision point, and the Live backend calls it when a run pauses. The Live tab's paused state was then checked in a browser against a live model: the banner, the recommendation arriving during the pause, an edited parameter turning Resume into Resume with changes (the run finished with it), and Stop from the banner. Looking at it led to a tidier paused state (one banner instead of three boxes saying the same thing, pauses marked on the charts with labels that avoid flagged frames, a Goal placeholder that is not about Raman). Two more fixes came out of these runs. A change that also shortens the measured window (the AFM curve reaching its trigger early) had no location, because such frames were never held; they are now held, compared and located where they overlap the stream's window, and the lost part of the axis is reported as a `window` entry (live: new intensity from 27.9 to 33.7 nm against a fitted contact point of 27.77 nm, window lost from 33.8 to 64.8 nm). And the frame that announces a change no longer also accepts the new state, so the log reads novelty → paused → resumed → state_accepted; without a pause this costs one more flagged frame per change. Known limits of this section: a gradual transition is followed silently and shows only in the distance from the reference; the location label is sometimes imprecise ("shifted" where "new" was correct); audits lag on fast line scans.

### 6. Datacubes: a second modality of the same loop

- `live/modality.py` is the seam. The loop (two clocks, flags, patience, change signal, novelty, audits, pausing, recommender, log) is modality-neutral; an adapter says which agent locks and replays the recipe, how a result becomes flat features, what "still fits" means, which curves the change signal reads, and which curve-only mechanisms apply (pinning, snippet edits, the portability check, a multi-frame reference and window re-anchors are capability flags, not branches in the loop). An instrument declares `modality`; the loop follows. `CurveModality` is the loop as it was.
- `HyperspectralModality`, built from the series machinery: the fast path is the locked replay made strict (`HyperspectralAnalysisAgent.analyze(strict_replay=True)`: a whole run with zero model calls, no skill selection, no execution repair, no salvage or not-measurable judge, no synthesis, no per-frame HTML report; a script that raises fails the frame); the verdict is `_replay_map_gate` against the reference's map statistics (now persisted as `feature_records`, so an adopted run arms a loop with no model call); a rebuild or an audit is a `locked_targets` run in the worker, so tracked names hold by construction and nothing is pinned; features are per-map means and scalars.
- `drift.DriftBank`: one `DriftMonitor` per curve, the frame as changed as its most changed curve. A cube is read as the mean spectrum of the whole field and of each quarter, on its physical axis. On the simulated series a new mode over one corner was 3 % of the whole-field mean and went unseen; by region it is announced, located, and the region named. One curve behaves exactly like one monitor, and old saved state loads.
- Instruments: `Frame.cube` (`.npy` + sidecar), `ReplayInstrument` over a folder of `.npy` / HDF5 cubes, and `spectrum_image_series` (simulated in-situ heating: a red-shifting plasmon on a continuous field, a second mode appearing over one corner). Sidecars may carry small nested descriptions of scalars (a cube's `energy_range`), never arrays or bulk.
- Found in the hyperspectral agent on the way: a replay under `synthesis="none"` ended "Pipeline returned no results" (both interpretation stages skipped; it now gets a deterministic result); a reduced profile silently raised a replay's forced single attempt to 2; the replay gate's range rule is right for sibling datasets and wrong for a stream, where the tracked quantity is expected to move (live, it withheld the plasmon map on 4 of 11 real tiles for a plasmon 35 to 60 meV below the first tile's), so a live loop asks the gate for method health only (`values_may_move`) and judges plausibility with its own range gate, the series rule unchanged; a reference whose required map never passed verification now fails setup with what happened.
- Live, Bedrock Opus 4.8. Simulated series: armed in 164 s (`quick`, 9 calls), 20 cubes at 1.2 s median with no model call, plasmon energy within about 1 meV of truth on every frame, the new mode announced on its first frame and located at 0.951 eV (truth 0.95) in the lower right, audit with locked targets agreeing to 0.4 %. Real data (the EELS demo cube as 12 tiles): under `quick` the reference's per-pixel map was rejected three times; with a thorough reference (334 s, 14 calls), then adopted with no model call (5 s): 11 of 11 tiles clean at 3.4 s median, no model call, plasmon energy 0.407 to 0.462 eV across the field, audit agreeing to 0.04 %. A cube rebuild takes minutes while frames take seconds, so a short stream ends before it lands.
- Not yet for cubes: window re-anchors, a multi-cube reference, recipe edits, the portability check, maps in the Live tab (it shows the mean spectrum), finer or adaptive regions, a gradual onset (absorbed silently, as for curves).
- New offline tests: `test_hyperspectral_measurement_loop` (the real agent on the fast path with a model that raises if called), additions to `test_hs_locked_replay`, `test_drift_monitor`, `test_web_measurement_loop`.

Also since the last review: **a first fit that produces nothing gets one re-plan** (the reviewer's last open item). Every later QC stage returns the best attempt so far; a first fit whose every script pass failed has none, adaptive refit skips single spectra, and reduced depth has no best-of-N. The engine now offers the host one recovery (`qc_recover_initial_failure`): the curve host re-plans once, told which plan could not be fitted and how it failed, and re-enters the normal flow under the same profile. Off for realtime and strict replays, inside best-of-N candidates, after a timeout, and once the run budget is spent. Live with the first fit planted to fail under `extract`: `success`, R² 0.969, where it was `error`.

### Verification

- Offline: full suite with `--ignore-glob='*live*'` (the `*_live.py` files need real API keys), 23 failed / 3181 passed / 29 errors on this branch (last run on the final commit). Diffed against a `main` worktree run with the same flags: **no branch-only failures** (the failures are pre-existing environment and ordering noise, including 6 QC goldens that already fail on `main`).
- New offline tests: `test_stage_timing_and_bank_assist`, `test_script_bank_hardening`, `test_qc_profile_quick`, `test_run_budget`, `test_depth_channel`, `test_measurement_loop`, `test_recommendation_contract`, `test_pinned_outputs`, `test_instruments_and_simulators`, `test_noise_calibrated_gates`, `test_recipe_portability`, `test_acquisition_skills`, `test_web_measurement_loop`. New test files deliberately avoid "live" in their names so the glob does not drop them.
- Live, Bedrock Opus 4.8, final sweep on the final code:

| run | reference | result |
|---|---|---|
| Beamline XRD, phase transition | 1 frame | 2θ within 0.01 %, FWHM 0.7 %. Change flagged 2 frames before the scripted onset. **One** rebuild planned from 5 frames (2 regimes seen), drift reference re-based 2 frames after adoption. |
| In-situ Raman, anneal + cosmic rays | 3 frames | G position 0.01 %, D/G 0.5 %. Recommender (raman acquisition skill) shortened the frame once precision was met. 58 of 60 frames clean. |
| In-situ Raman | 1 frame | G position 0.01 %, D/G 0.6 %. |
| AFM force curves, stiff inclusion | 1 frame | Contact point 0.2 %, stiffness 1.2 to 1.4 %. Flagged on the frame, rebuild from 5 frames, 27 of 28 clean after. Recommender raised the trigger force 20 to 30 nN, under the stated 45 nN limit. |
| STM dI/dV, tip change | 3 frames | Gap 0.14 % then 0.7 %. Flagged on the frame, rebuild from 5 frames (2 regimes seen), 37 of 37 clean after. Recommender named the tip change as something no setting repairs. |
| Real EELS line, 74 pixels | 8 frames | 63 of 67 frames clean, peak energy within 8 meV of a model-free estimate (FWHM about 200 meV), height r = 0.999, recipe portable, no needless rebuild. |
| Real EELS line + injected feature from pixel 50 | 8 frames | All 25 injected frames flagged, the first one on arrival, rebuild started 2 frames later. |

No frame called a model in any run. Every rebuilt recipe reported the same pinned names, and the pinned-value review passed on the first attempt in all runs.

The 10 s frame-deadline check **failed in the simulator runs of this final sweep** (1 to 47 frames per run flagged slow, worst frame 62 s). The machine had a load average of 8 to 12 at the time: several live runs, the offline suite, UI builds and a manual Live-tab session were running together. The loop flagged those frames as designed and no measurement was affected. In the earlier sweeps of the same experiments on a quieter machine the median was 1 to 2.5 s per frame, the slowest frame under 7 s, and every check passed (8/8, 5/5, 9/9, 9/9). Frame latency on a dedicated machine next to an instrument has not been measured.

### What real data and the final sweep exposed (kept here, not in prompts)

- Both replay gates were calibrated on clean data. On the real EELS line R² tracked signal strength (the approved reference sat at 0.91) and the peak-counting fingerprint wobbled on noise: 54 of 74 frames flagged, two pointless re-anchors. A zero-LLM sweep of one locked recipe over 225 real spectra: the constant R² bar called 59 % of untouched spectra poor, the constant fingerprint bar 83 %. Calibrated, the only alarms (4.4 %) were pixels off the nanocrystal array. An injected feature 60 / 30 / 15 % of the peak height was caught 100 / 47 / 2 % of the time at tolerance 1.25. The fingerprint drift signal separated nothing at that noise level. **One instrument and one recipe: the tolerance is provisional.**
- A locked recipe can carry a constant read off the reference despite the prompt rule (`max=600` on an amplitude from a weak reference pixel). The portability check catches it at setup. Repair is not built.
- Pinned outputs could be wrong while every check passed: the recipe reported the right D/G ratio while the added lines re-derived it with the baseline subtracted twice. Fixed with an independent value review and a prompt principle (prefer what the script already reports).
- Model choice is a systematic error the loop cannot see. Raman references at R² 0.99 differed by 8 % in D/G between line shapes, and one that subtracted a soft ALS baseline lost half of the broad D band (D/G 0.59 against 1.19). The Raman analysis skill now says a data-driven baseline is the wrong tool for broad bands and that results must survive a tenfold stiffer baseline.
- On weak EELS frames the fitted plasmon width doubled with a small reported error (peak trading against the background tail). Fit-quality gates cannot see a parameter correlation. Not fixed.

### Response to the independent live test (first review comment)

Fixed: a background re-anchor could outlive its parent (`MeasurementLoop.close()`, context manager, worker terminated at exit, Live backend closes in `finally`); per-file sidecar metadata now reaches the reference, every frame and every re-anchor (the caller's `system_info` wins, a simulator's truth never travels); `targets` against a mandatory skill rule now has a stated winner (the skill's rule about fit range or model stands, the scope decides what is verified and reported); the reduced-depth log no longer claims plan validation is skipped when the skill-rule check still runs; the realtime banner no longer prints per frame.

Documented: a stream shorter than a rebuild never sees its result (pace a replay); reduced depth can end flagged or in error where thorough converges; output names must be identifiers and definitions should stay true through the run; the reference should look like the run (use several reference frames when signal strength varies). On the bank: the zero-call fast path needs cross-dataset warm-up. A recipe banked under a reduced-depth profile is provisional until it has proven itself on a second, different dataset, and the bank is off unless persistent memory or `SCILINK_SCRIPT_BANK=1` is set.

Addressed since (section 5): the drift fingerprint that saturated on rich multi-peak spectra is replaced by a graded, model-free change signal, and the case no gate can see (a model that fits well and cannot represent the physics) now has independent audits. Audits detect a disagreement between two analyses; they do not tell which is right. The instrument-constant case where a recovery ends in a defaulted constant is an analysis-skill question and is not changed here. Per-frame constants inside a locked script (a value that differs from file to file) are not handled: the sidecar reaches the model-driven stages, but a locked script only sees the data.

### Not in this PR

- Live mode for images (needs a deterministic per-item gate and an acquisition-quality skill first). For datacubes see section 6 and its "not yet" list.
- Analysis skills for STS and AFM force curves in `curve_fitting/` (the STM reference only converged once `system_info` named the Dynes model).
- Automatic repair of a non-portable recipe, multi-objective BO recommender, a force-rebuild control, running the claims and literature-novelty chain automatically on a pause, a per-instrument home for recipes and history, restoring a finished run after a server restart, SSE instead of polling, a Live tab control for `reanchor_frames`.
- Bank matching quality (fingerprint noise sensitivity, threshold re-calibration) and the edit-adapt whitespace fragility.
- `docs/proposals/fast-and-live-analysis.md` (the working proposal with all measurements) is intentionally kept local and is not part of this PR.


ziatdinovmax:
# PR 656 — independent live test (review notes)

Ran the fast-analysis profiles and the `scilink.live` measurement loop end to end
on **Bedrock Opus 4.8**, against the four bundled simulators and against
independent **real** 1D spectra spanning several techniques (a multi-peak
diffraction series of a few dozen frames through a genuine structural change; a
noisy real line-scan series; several variable-temperature resonance-spectroscopy
series over both narrow and wide temperature ranges; and single spectra including
a band-crowded case, a core-loss edge, and a resonance technique that needs an
instrument constant).
Frames were run with the frame-deadline disabled — the test machine was under
heavy concurrent load, so absolute per-frame latencies are inflated and
slow-frame flags would be spurious; latency numbers below are for orientation
only, not a clean-machine measurement.

Overall: **the core design holds up and the PR's own "what real data exposed"
caveats are corroborated on independent data.** No blocking correctness issue
found. Notes below are calibration/robustness/UX/documentation, roughly in
priority order.

## What is validated

- **Lock-once / replay-per-frame:** zero model calls per frame across every run;
  pinned outputs reported under the requested names; change detection fires on the
  correct frame (a simulated inclusion crossing and a tip change were both caught
  the instant ground truth flipped — the tip change via the fit gate, not drift).
  Per-technique accuracy vs. simulator truth was excellent (primary quantities
  within ~0.01–0.2%, secondary within ~1–2%). On the noisy real line scan the
  tracked peak position stayed stable to ~2% across every frame even where the
  fit gate was unhappy — the quantity of interest was robust. The loop also held
  up across additional real series and techniques: a coherent narrow-range
  variable-temperature series tracked mostly clean (peak position + linewidth),
  and a broad-lineshape wide-range series ran the full range with zero flags —
  gate behaviour tracked the technique rather than a fixed threshold.
- **Rebuild → adopt → recover, on real data:** on a realistically paced real
  stream the background re-anchor completed (~150 s, 7 model calls, charged
  off-path — not to any frame), was adopted between frames, and the drift
  resolved: the clean-frame fraction jumped from ~1/3 before adoption to ~9/10
  after. The re-anchor re-pinned the same output names on the first attempt.
- **Pinned-output value review works on real data:** at lock time it *rejected* a
  first pinning attempt that computed a peak height as amplitude + (negative
  baseline offset) instead of the amplitude alone, and forced a correct
  re-pin — exactly the class of silent error the PR added this review to catch.
- **Script-bank fast path (verbatim reuse, zero model calls):** with the bank
  enabled and a *proven* recipe already present for that kind of data, repeated
  reduced-depth analyses short-circuited to verbatim reuse with **0 model calls**
  (locking at R²≈1.0, match score rising as cross-session evidence accumulated).
  Note the gate that makes this conservative: a recipe banked under a
  reduced-depth profile is **not** reused verbatim until it has proven itself on a
  *second, different* dataset — so two back-to-back reduced-depth runs on the same
  data do *not* short-circuit (the first bank entry is provisional). The claim is
  real; the fast path just needs cross-dataset warm-up, which the PR text could
  make explicit. (Bank is off by default unless persistent memory or the env
  override is set.)
- **Profile plumbing:** `extract`/`quick` demonstrably skip the stages they claim
  (conformance, literature, adaptive refit, trend, synthesis) and cap
  verification, with deterministic gates unchanged.
- **Domain-skill behaviour on techniques needing an instrument constant:** across
  more than one resonance technique the skill correctly *refused to guess* a
  missing instrument constant and aborted; on retry it recovered by broadening its
  metadata probe and, in one case, falling back to the near-universal default for
  that instrument class. The plan-conformance check separately caught a generated
  script silently hardcoding a fallback constant. Sound guards overall, but two
  notes: (a) the recovery can end in a *defaulted* constant — arguably the guess
  the first attempt refused, so consider making it a required input; and (b) the
  constant was present in each file's sidecar, yet the live-loop reference setup
  reads only `system_info` (not per-file sidecars), so I had to inject it by hand
  — a real user of these techniques would hit the same abort. Worth wiring the
  reference file's sidecar through, or documenting it.

## Findings

1. **Drift fingerprint saturates on rich multi-peak real spectra.** With a clean,
   high-SNR reference the calibration cannot relax the constant bar (it only
   lowers it, and only for a noisy reference), so a genuinely evolving pattern
   pins at a single sub-threshold similarity and flags most frames with no
   gradation. On simple spectra the same gate localizes changes cleanly. Not a
   false positive (the data really did change) but the signal carries no magnitude
   information once past threshold. Corroborates the PR's "tolerance is provisional
   / one instrument" note — now on a second real technique. Worth treating the
   fingerprint-bar calibration as an open item and considering a graded/continuous
   drift signal.

2. **A rebuild is only *adopted* when the frame interval ≳ the re-anchor time.**
   At full replay speed the re-anchor starts correctly but the stream ends before
   it lands, so `reanchors` stays 0 and the recipe never recovers. Only realistic
   pacing produced adoption. Inherent to burst-replay, but worth documenting (and
   the replay/demo helpers might default to a pacing that lets a rebuild land) so a
   reviewer replaying recorded data at full speed doesn't conclude the loop "never
   re-anchors."

3. **Reduced-depth profiles trade robustness — a hard fit can exit `error`.**
   `extract` on a hard single spectrum (a core-loss edge whose concave pre-edge a
   locked linear background can't follow) exhausted its reduced iteration budget
   (2 passes, no refit) and returned `status: error`. The same data under
   `thorough` would very likely converge. Fine as a documented trade-off, but the
   depth knob should carry a clear expectation that reduced depth can fail to
   converge on difficult fits where the default succeeds.

4. **`extract` + `targets` collides with mandatory domain-skill rules, and the
   resolution is nondeterministic.** On a band-crowded spectrum, one run honored
   the target scope and cropped to the target band, then hit the reduced iteration
   cap with the verifier still rejecting and returned a *flagged* best-effort;
   another run fit the full range per the skill's mandatory rule. The
   purpose-scoping and the "fit the full range" mandate pull against each other;
   worth deciding which authority wins when a target scope and a mandatory skill
   rule conflict.

5. **Live-loop robustness depends on a representative reference.** On the noisy
   real line scan, an unrepresentative reference (locked on an atypically weak
   stretch) made the fit gate flag most later frames on amplitude/signal-strength
   alone — which also masked a deliberately injected feature. The tracked peak
   position stayed correct throughout, so the physics survived, but this reinforces
   the PR's "fit quality tracks signal strength on real spectra" caveat and argues
   for guidance (or a check) that the reference span the run's dynamic range.

6. **A background re-anchor can outlive its parent on a clean exit.** After a run
   whose auto-escalation had started a background re-anchor finished its frames and
   exited, the re-anchor subprocess was left running orphaned and had to be killed
   by hand. The loop should terminate (or join) its background re-anchor on
   shutdown so a headless caller doesn't leak a model-driven process.

7. **Fit-quality gates cannot see a model-adequacy problem — confirmed on real
   data.** On a broad-lineshape wide-range series the run was flag-free end to end,
   yet the reported linewidth trend did not show the physical change the series was
   designed to reveal: the single locked lineshape can't represent it, and no gate
   catches that. Reinforces the PR's own "gates can't see a parameter correlation"
   caveat — the pinned numbers can be gate-clean and still model-limited.

8. **Minor / cosmetic:**
   - Pinned output names must be valid Python identifiers; a name led by a digit is
     rejected at setup with a clear error (fail-fast — good, just undocumented).
   - The `extract` summary logs "skipped: plan validation," but a skill-rule
     plan-validation model call still runs (only the skill-free sanity pass is
     skipped). The message reads as more aggressive than the behavior.
   - The realtime-profile banner prints on every frame — log noise on a long run.
   - Pinned-output definitions that name "the strongest peak" become ambiguous
     when peaks trade dominance through a transition; the loop faithfully reports
     whatever the recipe computes, so this is a usage lesson, but a one-line
     caution in the docs would help.

## Not covered

- Recommender paths (rule-table / GP / LLM) were not exercised — no recommender
  was attached in these runs.
- Web UI Live tab not tested (backend/CLI only).
- Clean-machine frame-latency numbers (the load confound above).

ziatdinovmax:
### Re-tested after the review-response commit — three fixes verified

Re-ran on **Bedrock Opus 4.8** against the commit that addresses shutdown, sidecar
metadata, and targets-vs-skill. Each fix confirmed **both** by the new offline
tests and by a live re-run; the two calibration/robustness items I'd flagged as
harder are still open, as expected.

| item | offline | live re-run |
|---|---|---|
| **Sidecar metadata** (a technique needing an instrument constant aborted at setup) | ✅ `test_sidecar_metadata_reaches_the_analysis_and_truth_never_does` | ✅ a resonance series run with **no constant in `system_info`** now completes setup — the value is read from the file's sidecar; the identical config aborted before |
| **Background re-anchor outliving the caller** | ✅ `..._reanchor_is_stopped_not_orphaned`, `test_the_worker_process_is_really_terminated`, `test_a_context_manager_closes` | ✅ no orphaned worker process left after runs exit |
| **Targets vs. a mandatory skill rule** | ✅ `test_targets_do_not_override_a_skills_mandate` | ✅ both reduced-depth runs on the band-crowded case now deterministically fit the full range per the skill rule and succeed (no cropped/flagged best-effort) |

All 6 new tests pass.

**Still open (not targeted by this commit):**
- **Drift-fingerprint calibration.** The re-run underlined it: the *same* real
  series that flagged most frames in the first pass flagged **none** this time,
  under a different (nondeterministically-locked) reference recipe. The drift
  behaviour is strongly reference-dependent — the main calibration item.
- **Reduced-depth robustness.** A hard single fit can still exit `error` under the
  shallow iteration budget where the default profile would converge (a documented
  trade-off).

Nothing blocking here — the fixes do what they claim.

ziatdinovmax:
### Re-tested the drift-signal rework — the main open item is fixed

Re-ran on **Bedrock Opus 4.8** after the commits that replace the thresholded
peak-counting fingerprint with a graded, model-free change signal + located
novelty. This was my main open finding (the fingerprint saturated on rich
patterns and gave opposite verdicts on the same series under two recipes).
Verified three ways:

1. **Offline** — all 18 `test_drift_monitor.py` tests pass (graded-not-just-flagged,
   glitch-is-one-frame-not-three, deterministic-across-restart, and located
   new / missing / shifted / broad).

2. **Recipe-independence (the core fix).** The same real series that previously
   gave wildly different drift counts under two locked recipes now gives an
   **identical verdict across two independent recipes** — the signal reads the
   data, not the fit, so it no longer depends on which recipe locked. (A gradual
   evolution is correctly absorbed as the stream evolving rather than flagged as
   a discrete change.)

3. **Sensitivity, grading and location.** On a noisy real line with a discrete
   feature injected partway, the loop raised a `novelty` at the **exact** frame
   the feature appears, **graded** it far above a subtler real change (score ~37
   vs ~4.5), and **located** it to the right position (peak within ~0.001 of
   where it was injected), typed as a new feature. It also decomposed a real
   change elsewhere into located new/missing regions.

The three-judgements separation is doing its job: the fit gate still flags the
frames where signal strength varies, but the **change signal is independent and
clean** — the reference-representativeness confound that muddied the earlier run
no longer contaminates the change verdict. No orphaned worker processes; the
default `on_change="report"` accepted the changed states without a needless
rebuild.

Still open (untouched, as before, and non-blocking): reduced-depth profile
robustness — a hard single fit can still exit `error` under the shallow iteration
budget where the default profile converges (a documented trade-off).

ziatdinovmax:
### Re-tested the datacube + image live loops, the recovery fix, and conventional analysis

Ran on **Bedrock Opus 4.8** after the commits that add the datacube and image
modalities, a "first fit produced nothing" recovery, and touch shared
(non-live-loop) code. Verified the new work live and confirmed nothing regressed
outside the loop.

**All three live-loop modalities now validated live**

- **Curve** — the graded, model-free change signal is recipe-independent (the same
  real series that once flagged most vs. no frames under two recipes now agrees),
  and a discrete injected feature is caught at the exact frame, graded, and
  located.
- **Datacube** — strict cube replay, zero model calls, sub-second/frame. A gradual
  drift of the whole field is correctly absorbed; a discrete second mode is caught
  at its onset frame and located to **both** the spectral position **and** the
  field region (the correct corner on a 4×4 grid), via the per-region pyramid.
- **Image** — strict image replay, zero model calls. Because an image analysis has
  no R², the replay verdict is method-health only and *cannot* see a
  wrong-but-running segmentation: on a coarsening series the locked recipe kept
  counting only the large particles (its gate said "good") and the **change signal
  was what caught the new population**, located by length scale and field region —
  exactly the design's claim. On real atomic-resolution data (a tiled mosaic with
  a defocus injected partway) the change signal graded genuine tile-to-tile
  variation apart from the defocus, typing the blur as *missing fine detail across
  the whole field* — the physically correct signature.

Offline for the new modalities: image loop + strict-replay golden, datacube loop,
and drift-monitor suites all pass.

**Recovery fix (first fit produced nothing → one re-plan)**

- Offline recovery tests pass. Note it targets the *produced-nothing* path; a
  distinct failure mode remains where a reduced-depth profile's verifier rejects a
  produced fit to its iteration cap and returns `error` (a hard fit the default
  profile would converge). That looks like the intended depth trade-off rather
  than a bug, but worth documenting as such.

**No regression outside the live loop** (diffed against current `main`)

- Conventional-analysis goldens (curve / image / hyperspectral): the only failures
  are **identical on `main`** (pre-existing environment noise) — **zero
  branch-only regressions**, and the branch adds new passing strict-replay
  goldens.
- Non-live suites for the touched modules (profiles, depth channel, recovery /
  budget, model cache, script bank, and conventional curve / image /
  hyperspectral / orchestrator) pass; the handful of errors reproduce identically
  on `main` (a stale test for a method removed earlier).
- Live confirmation: each agent's *conventional* reference analysis ran live on
  this HEAD and succeeded — the image mosaic's reference is a full multi-pass image
  series analysis, and the cube's reference is a full hyperspectral series
  (scout → regime plan → anchor decomposition).

**One practical note:** the reference *setup* for a complex atomic-resolution
image is expensive — a full multi-pass analysis plus segmentation and the
portability check took ~30 min on a loaded machine. It is a one-time slow-clock
cost; per-frame replay after it is sub-second to a few seconds. Worth a word in
the docs so a first setup on rich image data is not mistaken for a hang.

tkavinchakravarthy-del (approved):

